PR #10, opened 2021-02-08
Fix broken editUrl in docs caused by master -> main base branch change

brettgoss:
When viewing [the docs](https://www.remotion.dev/docs/), I noticed that the 'Edit this page' link uses the master branch, but the base branch for the repo is `main`, so the link 404's. This should fix that issue.


JonnyBurger:
You are right! I forgot this when renaming the main branch.
Thanks a lot for sending a PR 💙

PR #14, opened 2021-02-08
fix for windows cp -> copy

nico-martin:
When using NPM on Windows I'm getting the a `spawn cp ENOENT`-Error.
```
Error: Command failed with ENOENT: cp {path}\node_modules\@remotion\bundler\web\index.html
{tempPath}\Temp\react-motion-graphicsXYlAK9
```
Therefore I changed `cp` to `copy` and now it works. Do you think this will have negative impact on Apple or Linux users?


JonnyBurger:
Hey Nico!

Thanks a lot for the PR!
Yes, unfortunately I don't have Windows at all. I'm planning on installing it to test Remotion on it too!

`copy` doesn't exist on my machine, so I'm simply adding a ternary statement for it 👍

PR #42, opened 2021-02-10
Test out JPEG rendering

JonnyBurger:
Turns out to be much faster

- [x] Test quality setting
- [x] Allow PNG still for PNG sequences
- [x] Benchmark

---

**Update:** Switching to JPEG rendering dramatically speeds up rendering! Took a benchmark and the rendering of the frames went from 14sec to 6.5sec. Making this the new default, except when you use `--png` flag. Also added a new flag `--quality`. For SSR usage, we keep it at PNG be default but make it configurable. plus new templates will have jpeg as default in the config.


tomByrer:
I like speed optimizations!
Faster because of disk I/O?  How much space would this need to render?  I have 32Gb of RAM, so perhaps I could make a RAMDrive 0:)
Would WebP or [AVIF](https://reachlightspeed.com/blog/using-the-new-high-performance-avif-image-format-on-the-web-today/) would be worth the trade offs?

JonnyBurger:
I think encoding a JPEG is just more efficient than PNG, that's the whole optimization! Doing more experimenting with this now.

PR #49, opened 2021-02-10
Use `npm init video` instead of `npx create-video`

knpwrs:
This PR updates the docs to use `npm init`, the `npm` counterpart of `yarn create`.


JonnyBurger:
Yes, sounds a bit better! Thanks for the PR 💙

PR #58, opened 2021-02-11
Remove term "open source"

### LINKED ISSUE #47: Incorrect wording in license document
The [license document](https://github.com/JonnyBurger/remotion/blob/main/LICENSE.md) says:

>  This two-tier system was designed to ensure funding for this project while still allowing it to be open source.

However, this is incorrect; the license conditions violate points 3 and 5 of the [Open-Source Definition](https://opensource.org/osd-annotated), and therefore it cannot be called open-source. Something being open-source is about more than just being able to look at the source code, and so several additional requirements around usage rights and non-discrimination apply.

The common term for the sort of license used for this project is a "source-available license", though "while still allowing the source code to be publicly readable" would probably be easier to read and understand :)

Relatedly, the following [from the website](https://www.remotion.dev/docs/license) is also not entirely right:

> Unlike many open source projects, it is not allowed to clone or fork Remotions code for the purpose of creating a competitor.

You can just leave out "many" here - people being allowed to fork and/or create a 'competitor' is inherently allowed by an open-source license. If that is not allowed, then it wouldn't qualify as open-source.

Those seem to be the only two places I've found; everything else I've looked at seems fine to me, wording-wise.

JonnyBurger:
Although I thought of Remotion as an open source project, technically it is not open source because the license tells you cannot redistribute Remotion on your own.

Now this doesn't mean that Remotion is closed source or anything, in many ways it works just like all OSS projects, except that there is a clause that big companies cannot use it for free, and therefore, I also have to close the loophole of being able to fork and redistribute.

Nonetheless I don't wanna throw around the word 'open source' when it technically doesn't meet the definition.

I took the feedback from #47 and changed the wording. I didn't wanna say "unlike open source projects", because it sounds like Remotion is closed source. Therefore I said "unlike most projects on Github".

Fixes #47

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,7 +6,7 @@ For bigger projects, please coordinate with [Jonny Burger](https://jonny.io) to 
 
 Please note that since I charge for Remotion when companies are using it, this is a **commercial project** by me (Jonny Burger). By sending pull requests, you agree that I can use your code changes in a commercial context.
 
-Furthermore, also notice that unlike most open source project, you **cannot redistribute** this project. Please see [LICENSE.md](LICENSE.md) for what's allowed and what's not.
+Furthermore, also note that you **cannot redistribute** this project. Please see [LICENSE.md](LICENSE.md) for what's allowed and what's not.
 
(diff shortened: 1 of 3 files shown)


joepie91:
Changes look good to me :+1:

PR #91, opened 2021-02-14
Leading zero padding on output filenames

### LINKED ISSUE #89: PNG output - zero-pad output indices
## Issue:

Because the render `--png` mode outputs images without leading zero-padding, `ls` outputs them in the wrong order (using alphabetical rather than numerically aware order). This means that using ImgMagick to convert them into a gif fails because the frames are out of order.

## Fix:

The indices should be padded with leading zeroes so that ls -a of the output folder is in frame order.

zackdotcomputer:
Resolves #89 by adding leading zero padding the index number on saved filenames so that all index numbers in a batch are the same string length. This fixes `ls` ordering and `*` glob ordering of the collected files.

Also switches ffmpeg stitcher to using `*` glob selection rather than `%d` selection to remain compatible with that new numbering.


packages/tsconfig.settings.json
```diff
@@ -6,7 +6,7 @@
     "declarationMap": true,
     "sourceMap": true,
     "composite": true,
-    "lib": ["es2015", "dom"],
+    "lib": ["es2018", "dom"],
```
zackdotcomputer, line 9: Note: this lib change was required to add `padStart` to the lib. Since the compilation target was already es2018, this should not affect compatibility, just what TS flags as available.

Iamshankhadeep: I see now. I was thinking why this change is necessary.

JonnyBurger: Yes, let's not worry about old Node versions in this brand new project. Will check it out now 👍

JonnyBurger:
Works wonderfully! Thanks a lot for the PR 💙 

This will be in the next release, likely coming out tomorrow, or you can upgrade to the prerelease right now, hope the little roadbump didn't slow you down too much!

```
 - @remotion/bundler@1.3.0-alpha.ca2e3e45
 - @remotion/cli@1.3.0-alpha.ca2e3e45
 - remotion@1.3.0-alpha.ca2e3e45
 - create-video@1.3.0-alpha.ca2e3e45
 - @remotion/eslint-config@1.3.0-alpha.ca2e3e45
 - @remotion/eslint-plugin@1.3.0-alpha.ca2e3e45
 - @remotion/renderer@1.3.0-alpha.ca2e3e45
```

zackdotcomputer:
👍 not a problem at all. I really like the tool and am happy to help make it better.

PR #92, opened 2021-02-14
Spelling

jsoref:
Normally I offer a fancy explanation of what I did, but sadly, it appears that none of these items were flagged at the end.

`GitHub` and `JavaScript` are brands and have special casing rules (sadly, they both allow lowercase which confused my poor tool at the end).

`vide` is apparently a word that means `see!`, but I'm pretty sure you meant `video` :-)


JonnyBurger:
Thanks a lot for the spell check @jsoref! 💙 Very nice bot you developed!

PR #93, opened 2021-02-14
Fixed remotion.config.ts typo

Iamshankhadeep:
Fixing typo on remotion.config.js which was causing `render` to fail


JonnyBurger:
Thank you! 💙 Oh I see the typo now, I might wanna add this file to the test suite as well 🤦

PR #100, opened 2021-02-15
Added support for H265, VP8, VP9, customized CRF

Iamshankhadeep:
Support for webm-v8 and webm-v9


packages/example/remotion.config.ts
```diff
@@ -1,6 +1,6 @@
 import {Config} from 'remotion';
 
-Config.Rendering.setConcurrency(16);
+Config.Rendering.setConcurrency(4);
```
JonnyBurger: We can set it to

```
import os from 'os';
Config.Rendering.setConcurrency(os.cpus().length);
```

So it will work across all our computers

packages/cli/src/get-filename.ts
```diff
@@ -6,5 +6,5 @@ export const getOutputFilename = () => {
 		console.log('Pass an extra argument <output-filename>.');
 		process.exit(1);
 	}
-	return args._[3];
+	return args._[3].split('.')[0];
```
JonnyBurger: Let's make it more strict so that if a user has an weird configuration like MP4 file but with WebM extension, it doesn't magically replace but instead throws an error.

Let's enforce:

- If user has MP4 format selected, the output extension must be .mp4
- If user has WebM format selected, the output extension must be .webm
- If user has PNG sequence selected, there can not be an extension (if this is hard to do, let's do that it cannot be .mp4 or .webm)

If these conditions are not met, let's `console.log` a message and `process.exit(1)`.

packages/cli/src/parse-command-line.ts
```diff
@@ -24,6 +25,9 @@ export const parseCommandLine = () => {
 	if (parsedCli.png) {
 		Config.Output.setOutputFormat('png-sequence');
```
JonnyBurger: Let's deprecate the PNG flag and undocument it, because it's redundant now.

If user passes CLI flag `--png`, let's log `--png has been deprecated, use --format=png from now on` (but let's let the user continue for now)

packages/cli/src/render.tsx
```diff
@@ -28,20 +28,33 @@ export const render = async () => {
...
 	const absoluteOutputFile = path.resolve(process.cwd(), outputFile);
+	const absoluteOutputFileWithExtension = (): string => {
+		if (renderMode === 'mp4') {
+			return absoluteOutputFile + '.mp4';
+		}
+		if (renderMode === 'webm-v8' || renderMode === 'webm-v9') {
+			return absoluteOutputFile + '.webm';
+		}
+		return absoluteOutputFile;
+	};
```
JonnyBurger: The console message 'Your video is ready! -> {output location}' doesn't seem to contain the .webm extension.

JonnyBurger:
In `image-formats.ts`, we need to handle the new formats as well. I am getting an error if I just set

`Config.Output.setOutputFormat('webm-v9');` in remotion.config.ts.

JonnyBurger:
It's not as easy as I thought.. with webm-v8 the output quality of npm run render in example is very noticeabley bad.
And with webm-v9, it takes forever to encode.

We need to investigate further if we are missing some command line parameters or if there is something else we can do to make the experience better.

packages/example/remotion.config.ts
Iamshankhadeep: okay. that sounds awesome.

packages/cli/src/get-filename.ts
Iamshankhadeep: I was thinking let user provide only folder name. We will added the necessary extension automatically. so we can use folder name for `png` and will be adding necessary extension when they use `--format` flag and if they don't provide anything we will add `.mp4` automatically. How's that sounds?

packages/cli/src/parse-command-line.ts
Iamshankhadeep: yeah that is needed I guess. Let's keep --png flag for 2 more version. then we can just drop the support for png flag.

JonnyBurger: sounds good. or we can release like version 2.0 with audio support, and drop all the deprecated stuff 👍

packages/cli/src/get-filename.ts
JonnyBurger: Ok, sounds good, if you omit the extension, we will add it automatically.
Let's still allow to explicitly define the extension and validate it, it would be nice that if you specify the exact file it is guaranteed end up being that file. In case you are using it inside a script or something

Iamshankhadeep: Yeah that sounds like a nice functionality improvement.

packages/cli/src/parse-command-line.ts
Iamshankhadeep: yeah I was thinking the same.

packages/cli/src/render.tsx
Iamshankhadeep: Just fixed this.

packages/docs/docs/cli.md
```diff
@@ -15,12 +15,21 @@ Besides choosing a video and output location with the command line arguments, th
 
 - `--props`: React Props to pass to the root component of your video. Must be a serialized JSON string.
 - `--concurrency`: [How many CPU threads to use.](config#setconcurrency) Minimum 1. The maximum is the amount of threads you have.
-- `--png`: [Render a PNG sequence with transparency instead.](config#setoutputformat) If you use this option, the output location you pass must be a folder name.
+- `--png`: [Render a PNG sequence with transparency instead.](config#setoutputformat) If you use this option, the output location you pass must be a folder name. It will override `--format` flag.
+
+:::warning
+`--png` will be deprecated in thee upcoming major version release. Use `--format=png` instead of `--png` flag.
+:::
```
JonnyBurger: Let's remove any mention of --png and add the information about PNG to the `--format` parameter. Ideally the user doesn't even know this parameter exists if it's deprecated

Iamshankhadeep: okay done.

packages/renderer/src/stitcher.ts
```diff
@@ -38,10 +39,16 @@ export const stitchFramesToVideo = async (options: {
 			`${options.width}x${options.height}`,
 			'-i',
 			`element-%0${numberLength}d.${format}`,
-			'-vcodec',
-			'libx264',
+			options.outputFormat === 'mp4' ? '-vcodec' : '-c:v',
```
yuta0801: If I understand correctly, `-vcodec` and `-c:v` is exactly the same :)

JonnyBurger: Oh, good find @yuta0801! Can we remove this ternary statement @Iamshankhadeep? Doesn't matter which one we use

Iamshankhadeep: Yeah my bad I thought I removed it. But thanks again.

JonnyBurger:
TODO:

- [ ] Test manually all codecs
- [ ] Add some automated tests
- [ ] Try on Windows
- [ ] Go through docs and check if everything looks correct
- [ ] Check for backwards compatibility, what if using old options, everything clear?

PR #102, opened 2021-02-15
fix (maybe) typo in docs

LetsMelon:
I'm not sure if this is a typo. But if it is, here is a PR to fix it!
On the [docs-page](https://www.remotion.dev/docs/javascript/) for JavaScript is written `You may import import ...` so I 'fixed' it and deleted the second `import`.


retarsis:
that's definitely a typo

JonnyBurger:
100%! Thanks a lot for reporting and fixing it! 💙

PR #116, opened 2021-02-17
fixing wrong variable in the Error message

StasinskiPL:
As above, there was a name, but it should be id.


JonnyBurger:
Good catch! During the development, name was renamed to id, looks like it's some sort of global variable in JS. Thanks for the PR 💙

PR #117, opened 2021-02-17
Added support for node version checking

### LINKED ISSUE #108: Enforce minimum node version
Users need at least Node v12.10.0 for Remotion to work. Let's enforce that or at the very minimum show a warning

Iamshankhadeep:
fixes #108.


JonnyBurger:
Thanks! This works!
I would still like to perfect it a little more:

- Let's make the check into a function and call it as the first thing in `packages/cli/src/index.tsx`. Benefit: We don't need to load ts-node twice which is quite slow. Also the check will run before you start the CLI and not after install, this is a little bit better because the node version can change inbetween.

- By adding to the `engines` field in package.json, it will hard-fail when your node version is too low. So people will probably not even see the warning we put because NPM is failing. Let's inline the requirement so it's just a warning and not an error (for now)

JonnyBurger (approved):

JonnyBurger:
Thanks a lot! 💙

PR #120, opened 2021-02-18
fix(render): use queue instead of stack for resources and waiters

jeetiss:
I work on gifs and I figured out that frames render from end to start, because i use stack in pool
that affect render video perf and maybe something else 

fix it


JonnyBurger:
Interesting! Yes I can imagine <Video> will be slower this way. Thanks for the PR 💙

PR #124, opened 2021-02-19
fix: mark onProgressUpdate as optional in bundle

arthurdenner:
Hi! 👋🏽 

This fixes a current small issue where TypeScript doesn't compile `server.tsx` when running `npm run server` because the types don't match - `bundle` requires two arguments currently.

## Steps to reproduce

- Run `yarn create video`;
- Run `yarn server`;


JonnyBurger:
Good catch! That's a regression 1.3. Thanks for spotting and fixing it! 💙

PR #125, opened 2021-02-19
feat: improve keyboard navigation

arthurdenner:
Hi! 👋🏽 

The current implementation, using `div`s, prevents better keyboard navigation - we can't tab through the list of compositions or interact with the buttons without using the arrow keys.

This PR improves keyboard navigation by using `button` tags were applicable and adding the `href` attribute to the `a` tags of compositions.

While implementing these changes, I wondered if it'd interesting to try and add `react-router-dom` or something similar to implement actual navigation in React.

### DIFF
--- a/packages/cli/src/editor/components/CheckboardToggle.tsx
+++ b/packages/cli/src/editor/components/CheckboardToggle.tsx
@@ -3,6 +3,7 @@ import {
 	CheckerboardContext,
 	persistCheckerboardOption,
 } from '../state/checkerboard';
+import {ControlButton} from './ControlButton';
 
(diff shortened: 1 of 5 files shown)


Iamshankhadeep:
I must say this is a nice feature improvement. Let @JonnyBurger review it.

JonnyBurger:
Wow this works really well! Thanks a lot for this quality PR! 💙 
I'll make sure to shout you out in the next blog post.
And also I'll probably use it as a reference when I struggle in the future with adding keyboard controls 😁

arthurdenner:
Thank you for this awesome project and kind words!

PR #127, opened 2021-02-21
added delete webpack bundle in /tmp dir support

### LINKED ISSUE #126: Delete bundle in /tmp dir
After every run,the bundle stays in a tmpdir. Not crucial since it's temporary but would be better if Remotion cleans it

Iamshankhadeep:
fixes #126

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4356,29 +4356,6 @@
       "integrity": "sha512-YZo3K82SD7Riyi0E1EQPojLz7kpepnSQI9IyPbHHg1XXXevb5dJI7tpyN2ADxGcQbHG7vcyRHk0cbwqcQriUtg==",
       "dev": true
     },
-    "semver": {
-      "version": "7.3.4",
-      "resolved": "https://registry.npmjs.org/semver/-/semver-7.3.4.tgz",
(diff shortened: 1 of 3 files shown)


packages/cli/src/render.tsx
```diff
@@ -130,6 +130,9 @@ export const render = async () => {
 		await fs.promises.rmdir(outputDir, {
 			recursive: true,
 		});
+		await fs.promises.rmdir(bundled, {
```
JonnyBurger: Looks like we can speed it up using Promise.all!

package-lock.json
```diff
@@ -4356,29 +4356,6 @@
       "integrity": "sha512-YZo3K82SD7Riyi0E1EQPojLz7kpepnSQI9IyPbHHg1XXXevb5dJI7tpyN2ADxGcQbHG7vcyRHk0cbwqcQriUtg==",
       "dev": true
     },
-    "semver": {
```
JonnyBurger: Shouldn't change the package-lock file, please revert

JonnyBurger (changes requested):

JonnyBurger (approved):

PR #130, opened 2021-02-22
Throw error if useVideoConfig is not used in Composition

### LINKED ISSUE #123: useVIdeoConfig sometimes returns null
Looking at the type of useVIdeoConfig, it is a type that return () => VideoConfig, but sometimes it return null

code
スクリーンショット 2021-02-19 17 48 29

log
スクリーンショット 2021-02-19 17 49 11

yuta0801:
fixes #123


JonnyBurger:
Thanks a lot! That's perfect. Added a hopefully helpful error message and merged it, will come out in Remotion 1.4 later this week.

PR #132, opened 2021-02-22
Support for custom Puppeteer / Apple M1 support

### LINKED ISSUE #39: Apple Silicon support
Hey! 👋  Having issues running `yarn create video` on M1 powered device w/out Rosetta 2.

image

Node v15.6.0 (native ARM), npm 7.4.0, yarn 1.x

It seems like the issue is related to https://github.com/puppeteer/puppeteer/issues/6622, as Puppeteer does not find the proper installed binary during npm (re)build nor during launching browser

### LINKED ISSUE #109: When timing out, Chromiums are left opened on Windows
image

As reported on Discord

### LINKED ISSUE #131: Ubunutu on WSL2 - npm run build gets stuck
Default project gets stuck unless X server is running

Iamshankhadeep:
fixes #39 
Fixes #109
Fixes #131 

To do
- [x]  test on windows (Just tested on windows on both bash and cmd.exe, seems to be working)
- [x]  test on mac (intel) and mac (M1)
- [x] to write custom logic to locate chrome or chromium executable


Iamshankhadeep:
Currently it works on the default chromium revision `848005`.

packages/renderer/src/get-compositions.ts
```diff
@@ -1,10 +1,16 @@
...
 
 export const getCompositions = async (
 	webpackBundle: string
 ): Promise<TCompMetadata[]> => {
-	const browser = await puppeteer.launch();
+	const executablePath = await getLocalChromiumExecutable();
+	console.log(executablePath);
+	const browser = await puppeteer.launch({
+		executablePath,
+		args: ['--no-sandbox', '--disable-setuid-sandbox'],
```
JonnyBurger: Why do we need no-sandbox now?

packages/renderer/src/get-local-chromium-executable.ts
```diff
@@ -0,0 +1,25 @@
...
+import puppeteer_node from 'puppeteer-core/lib/cjs/puppeteer/node';
+import {downloadBrowser} from 'puppeteer-core/lib/cjs/puppeteer/node/install';
+
+const getLocalRevision = (): puppeteer.BrowserFetcherRevisionInfo => {
+	const productName = 'chrome';
+	const browserFetcher = puppeteer_node.createBrowserFetcher({
+		product: productName,
+		host: 'https://storage.googleapis.com',
+	});
+	const revisionInfo = browserFetcher.revisionInfo('848005');
```
JonnyBurger: Pretty impressive!

So this will only work if someone has exactly Chrome 84.8005 installed though? 🤔  Might be a bit too strict, maybe better to check for path (e.g. /Applications/Google Chrome.app)

JonnyBurger:
Let's also have a setting in config file `Config.Rendering.setChromiumExecutable()` to let people define it!

packages/renderer/src/get-compositions.ts
Iamshankhadeep: We don't need it now. But the puppeteer instance was not launching and I went to puppeteer troubleshooting docs for linux and it said to provide `no-sandbox` arg. That's why I included it.

packages/renderer/src/get-local-chromium-executable.ts
Iamshankhadeep: actually 848005 corresponds to [Chromium 90.0.4403.0](https://github.com/puppeteer/puppeteer/commit/890d5c2e57cdee7d73915a878bda86b72e26b608).

To check for executable path for chrome or chromium for each os is a bit complicated than it seems. Figuring out how to write the script for chrome or chromium executable for each os.

Currently how it works (in linux) is that if you don't have the chormium executable in `puppeteer-core` it downloads the default `Chromium 90.0.4403.0` in `node-modules/puppeteer-core/.local-chromium` and from the next time it automatically detects the executable path in `.local-chromium` and use that chromium executable.

Iamshankhadeep:
> Let's also have a setting in config file `Config.Rendering.setChromiumExecutable()` to let people define it!

will work on it.

monophthongal:
Would this also make it possible to use H.264/MP4 videos with the `<Video>` component?

Iamshankhadeep:
> Would this also make it possible to use H.264/MP4 videos with the `<Video>` component?

No it won't. It will make `yarn create video` smaller in size and when you try to render the video it will download latest chromium and use it for rendering.

if you go to [`<Video/>` api doc](https://www.remotion.dev/docs/video) you can see this.

`
Please note that while H.264 video (the codec of MP4 videos) will show up during Preview, it will not work during render. This is because Puppeteer with Chromium is used during rendering, which does not ship with H.264 codecs. Convert your videos to a Chromium compatible format such as WebM instead.
`

I hope it helps. @monophthongal

joshlatimer:
> > Would this also make it possible to use H.264/MP4 videos with the `<Video>` component?
> 
> No it won't. It will make `yarn create video` smaller in size and when you try to render the video it will download latest chromium and use it for rendering.
> 
> if you go to [`<Video/>` api doc](https://www.remotion.dev/docs/video) you can see this.
> 
> `Please note that while H.264 video (the codec of MP4 videos) will show up during Preview, it will not work during render. This is because Puppeteer with Chromium is used during rendering, which does not ship with H.264 codecs. Convert your videos to a Chromium compatible format such as WebM instead.`
> 
> I hope it helps. @monophthongal

According to https://stackoverflow.com/questions/47976790/play-mp4-in-chromium-with-puppeteer-windows it might, have you tested this? would be amazing as I don't really want to convert every mp4 to a webm since that's super slow

Iamshankhadeep:
> > > Would this also make it possible to use H.264/MP4 videos with the `<Video>` component?
> > 
> > 
> > No it won't. It will make `yarn create video` smaller in size and when you try to render the video it will download latest chromium and use it for rendering.
> > if you go to [`<Video/>` api doc](https://www.remotion.dev/docs/video) you can see this.
> > `Please note that while H.264 video (the codec of MP4 videos) will show up during Preview, it will not work during render. This is because Puppeteer with Chromium is used during rendering, which does not ship with H.264 codecs. Convert your videos to a Chromium compatible format such as WebM instead.`
> > I hope it helps. @monophthongal
> 
> According to https://stackoverflow.com/questions/47976790/play-mp4-in-chromium-with-puppeteer-windows it might, have you tested this? would be amazing as I don't really want to convert every mp4 to a webm since that's super slow

Thanks for the pointer. Now I think it will be possible @monophthongal  will try to solve it asap. thank you @joshlatimer.

JonnyBurger:
Looking and working well!

I did some things as well:

- Upgraded Puppeteer to 8.0.0
- Added some macOS paths to search for Chrome
- Ensure browser is there as the first thing when you render
- Made it generic and add support for Firefox but it does not yet work! FF has some bugs. So I put it behind a feature flag but I think we don't yet have to throw away the flag.
- disable `no-mp4-import` ESLint rule
- `Config.Renderring.setChromiumExecutable()` -> `Config.Puppeteer.setBrowserExecutable()` - more future proof in case we support Firefox and or support more puppeteer customization options

There are a few things still left to do:

- Add Windows and Ubuntu Chromium paths to search for (I will do this)
- Document setBrowserExecutable config
- Allow to configure via CLI: `--browser-executable` and document it
- Update docs and rewrite the MP4 import warning and explain it: It only happens when you have no Chrome installed. Also mention that the ESLint rule can be enabled

@Iamshankhadeep do you want to finish the PR off?

Thanks for this great PR! Works even better than I expected it could 😊 🙌

Iamshankhadeep:
> @Iamshankhadeep do you want to finish the PR off?
Yeah I will finish this PR and you can tweak later if needed.

JonnyBurger (approved):

PR #148, opened 2021-02-26
Add support to use requestAnimationFrame for render frames in preview

yuta0801:
This PR adds option to use `requestAnimationFrame` for play frames instead of `setTimeout`.

It was just an idea but it improves FPS greatly, especially effective for BetaText in examples.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -10,7 +10,7 @@
 	"scripts": {
 		"test": "eslint src --ext ts,tsx",
 		"build": "tsc -d && node add-bin.js",
-		"dev": "ts-node src/preview.ts x ../example/index.tsx"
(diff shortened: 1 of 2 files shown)


JonnyBurger:
Brilliant! 💡 

It indeed seems to be faster. I'm not convinced its because of RAF, but I believe it's because of the useRef technique you used.

2 things:

- I only add to lastFrames if actually triggering a rerender - don't wanna show 60FPS if composition is defined as only 10fps
- There was a one-off bug in FPS calculation - I fixed it

Result: BetaText example plays in perfect 30.0 FPS!! 🥳 
Before it showed like 27.8FPS and in actuality it was even slower because of the bug. You can see the improvement with your own eyes! Thanks a lot 💙

PR #151, opened 2021-02-26
Fix bugs in getTimelineClipName

yuta0801:
Fixed potential bugs in getTimelineClipName, one of them I found is `<Sequence><p>some text</p></Sequence>` will crash


JonnyBurger:
Looks good! Thanks a lor for the PR! 💙 😊

PR #157, opened 2021-03-01
Fix bugs in ffmpeg feature check

yuta0801:
<!--- 
  Please make sure to read CONTRIBUTING.md before sending a pull request.
  Thank you!
>

### DIFF
--- a/packages/renderer/src/ffmpeg-flags.ts
+++ b/packages/renderer/src/ffmpeg-flags.ts
@@ -1,5 +1,5 @@
 import execa from 'execa';
-import {binaryExists} from './validate-ffmpeg';
+import {validateFfmpeg} from './validate-ffmpeg';
 
 let buildConfig: string | null = null;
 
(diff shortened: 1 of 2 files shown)


packages/renderer/src/ffmpeg-flags.ts
```diff
@@ -1,12 +1,13 @@
...
-import {binaryExists} from './validate-ffmpeg';
+import {validateFfmpeg} from './validate-ffmpeg';
 
 let buildConfig: string | null = null;
 
 const getFfmpegBuildInfo = async () => {
 	if (buildConfig !== null) {
 		return buildConfig;
 	}
+	await validateFfmpeg();
```
JonnyBurger: Unfortunately this function calls `process.exit(1)`.
Functions in `renderer` should never call this, only from the CLI. Otherwise we are at risk crashing servers to whom Remotion is just one of many dependencies. Reverting this and merging, hope you understand!

JonnyBurger:
Thanks a lot for spotting this and making a PR! 💙 
Just one comment, see above. Still merging, just without this change.

PR #161, opened 2021-03-02
Allow user to select frame range

### LINKED ISSUE #104: Allow to select frames you want to render
It would be nice in some cases to only render the first n frames, or get a still frame at a certain position. Thinking about an API like `--range=1-10` or `--range=50`.

Iamshankhadeep:
fixes #104 
- [x] to write doc for `--range` flag
- [x] Writter test for `--range` flag


packages/renderer/src/render.ts
```diff
@@ -30,6 +32,8 @@ export const renderFrames = async ({
 	imageFormat?: ImageFormat;
 	quality?: number;
 	browser?: Browser;
+	frameRange: FrameRange;
+	frames: number;
```
JonnyBurger: kind of sucks but we might have to make it optional or release it as part of 2.0 breaking change. otherwise I think when you upgrade your project the types break

JonnyBurger:
Looking good to me, thanks a lot! 💙 
Will do some testing to see if everything is working as intended.
You could add docs and an integration test in `tests/rendering.test.ts` to improve the PR still.

JonnyBurger (approved):

PR #168, opened 2021-03-05
Improve development experience

yuta0801:
This PR do

- adds watch build way, just run `npm run watch` in each package
  - places hashbang to cli entrypoint directly since watch build and add-bin is incompatible (Node.js supports that from long time ago)
- ~allows to change bundling/rendering output path and rendering/stitching from it using environment~
  - it would be nice to support this feature officially, at the moment this is only for remotion developing


JonnyBurger:
Thanks a lot! Looking very good, trying it out now.

JonnyBurger:
Really awesome PR!

I have also added watch script to `core` package and a command to watch all packages at the same time.

Running `npm run watch` from the root is recompiling `core`, `cli`, `bundler` and `renderer` at the same time with one command, together with the TS project references setup it's really awesome, almost like it's not a monorepo!

Can you move the environment variable part into it's own PR?
I would like to merge the monorepo stuff but env variable stuff I think is not quite ready. Would be better IMO to make a separate PR and iterate on it there. I totally see the need for it though 👍 

Otherwise no objections.

packages/cli/src/render.tsx
```diff
@@ -188,7 +188,6 @@ export const render = async () => {
 		imageFormat,
 		quality,
 		browser,
-		frameRange: frameRange ?? null,
```
JonnyBurger: This shouldn't be removed

packages/cli/src/render.tsx
```diff
@@ -207,7 +206,7 @@ export const render = async () => {
 			},
 			cliProgress.Presets.shades_grey
 		);
-		stitchingProgress.start(rendered.frameCount, 0);
+		stitchingProgress.start(rendered?.frameCount ?? config.durationInFrames, 0);
```
JonnyBurger: This can be reverted now

JonnyBurger (changes requested):
Looking all good except these two things. I can also do them myself when I reach my computer

JonnyBurger (approved):

PR #169, opened 2021-03-06
add number array support to `interpolate`

### LINKED ISSUE #145: FEATURE: adding `none` as `extrapolate` config for `interpolate`
It would be good if we can chain several `interpolate` functions like:
```
const position= 
  interpolate(frame, [0, 10], [-200, 140], { extrapolateRight: "none" }) || 
  interpolate(frame, [90, 100], [140, -200], { extrapolateLeft: "none" }) || 100
```
For this to work, `interpolate` needs to return `undefined` for frames outside range specified.
This will allow us to make start/end transitions with much less complexity and code.

### LINKED ISSUE #164: Interpolate with arrays of any size
As I'm used to RN, I wanted to do something like:

```js
	const opacity = interpolate(
		frame,
		[0, intro , intro + main, intro + main + outro],
		[0, 1, 1, 0],
	);
```

Unfortunately, only arrays of 2 are supported  for now.

This is not very convenient as we  need to create multiple smaller interpolations to obtain the final result you want.

I'm using something like `const opacity = Math.min(opacityIntro, opacityOutro);`

Iamshankhadeep:
fixes #164 
fixes #145

### DIFF
--- a/packages/core/src/interpolate.ts
+++ b/packages/core/src/interpolate.ts
@@ -89,7 +89,11 @@ function checkValidInputRange(arr: number[]) {
 	}
 	for (let i = 1; i < arr.length; ++i) {
 		if (!(arr[i] >= arr[i - 1])) {
-			throw new Error('inputRange must be monotonically non-decreasing ' + arr);
+			throw new Error(
+				`inputRange must be monotonically non-decreasing but got [${arr.join(
+					','
+				)}]`
+			);
 		}
 	}
 }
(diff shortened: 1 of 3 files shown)


JonnyBurger:
Thanks! 😃

Looking good, but we need some tests!
See the unit tests for frame range that were added to core package yesterday.
I recommend using `npx jest --watch` to quickly iterate.

Iamshankhadeep:
> Thanks! 
> 
> Looking good, but we need some tests!
> See the unit tests for frame range that were added to core package yesterday.
> I recommend using `npx jest --watch` to quickly iterate.

okay will do

JonnyBurger (approved):

slorber:
Hi @JonnyBurger 

Is this published in 1.5.3? 

Because it seems not https://unpkg.com/remotion@1.5.3/dist/interpolate.js

JonnyBurger:
Hey Sebastien,

Sorry it’s not out yet! Working on a bigger release so it’s taking a bit longer between releases. Please just copy the function into your project for the moment and use it for the moment.

PR #172, opened 2021-03-06
fix: properly restart video in development mode

arthurdenner:
Hi! 👋🏽 

There's currently a bug in the `Video` component in development:

When playing in loop, the `video` doesn't restart correctly after its first run. It's reproducible in the `features` example:

https://user-images.githubusercontent.com/13774309/110224449-fcd9c580-7edb-11eb-9b4c-2e47ae7b412f.mp4

That's because, after setting the `currentTime` to `0`, we need to play the video again.

This PR tries to fix this behaviour:

https://user-images.githubusercontent.com/13774309/110224518-afaa2380-7edc-11eb-8bb7-18fbaf31493b.mp4

Let me know if you have any suggestions or find any bugs.

### DIFF
--- a/packages/core/src/video/VideoForDevelopment.tsx
+++ b/packages/core/src/video/VideoForDevelopment.tsx
@@ -22,11 +22,19 @@ export const VideoForDevelopment: React.FC<RemotionVideoProps> = (props) => {
 		if (!videoRef.current) {
 			throw new Error('No video ref found');
 		}
+
 		if (!videoConfig) {
 			throw new Error('No video config found');
 		}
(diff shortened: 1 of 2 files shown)


JonnyBurger:
Thanks a lot for this PR! 💙 

I can confirm this is a bug and that your PR solves the issue.
It made me think if we can write it even more generically, and I pushed another commit which I think achieves it - in case we ever want to skip frames when the rendering is too slow it also works if frame 0 is not played.

In any case, thanks a lot for figuring this out and we'll ship it in the next release! 😊

JonnyBurger (approved):

PR #173, opened 2021-03-07
Allow to skip bundling and rendering

yuta0801:
This PR is separated from #168

### DIFF
--- a/packages/cli/src/render.tsx
+++ b/packages/cli/src/render.tsx
@@ -5,6 +5,7 @@ import {
 	getActualConcurrency,
 	getCompositions,
 	renderFrames,
+	RenderFramesOutput,
 	stitchFramesToVideo,
 	validateFfmpeg,
 } from '@remotion/renderer';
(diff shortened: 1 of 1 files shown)


JonnyBurger:
Thanks for kicking this off! I want to do this but we need a better way to control it than with environment variables. By default we delete the bundle afterwards so we need to allow to change the output directory where it doesn't get deleted, but it's not obvious to the user they have to do this. Plus we don't yet handle any edge cases such as skipping bundling when there is no bundle.

Nonetheless this is very valuable to kick this off, I will myself also contribute to the PR and hopefully we can ship this as a nice built-in functionality.

JonnyBurger:
I am closing this PR because I think with the recent developments it's not as useful anymore - bundling is now a lot faster with ESBuild and Webpack caching, rendering has become faster as well.

The feature I think was meant to be for faster development speed anyway, we now have a lot of short tester compositions and the render pipeline has become more solid. For complex audio filter stuff, we have unit tests, so I have personally never felt the need to skip bundling or rendering stage recently.

Sorry and still, thanks a lot for sending the PR! 💙

PR #174, opened 2021-03-07
Fix bugs sequence which was causing 1 extra frame

### LINKED ISSUE #154: Sequencing bug
Reported by yuta on discord:

 this reproduce it (it's a bit strange but working)
```ts
import music from './music.mp3';

const AudioTesting: React.FC = () => {
    const {durationInFrames} = useVideoConfig();
    return (
        <div>
            <Sequence from={0 - 40} durationInFrames={durationInFrames + 40}>
                <Video src={music} />
            </Sequence>
        </div>
    );
};
```

### LINKED ISSUE #156: Sequence is 1 frame too long
A sequence with `durationInFrames = 100` is 101 frames long.
We may need to collect these breaking changes and make a major release.

Iamshankhadeep:
fixes #156 
fixes #154


JonnyBurger:
Thanks a lot!
This is a breaking change, so I will merge it for Remotion 2.0 and announce it.

You said it could potentially also help with #154? Is this the case?

Iamshankhadeep:
> You said it could potentially also help with #154? Is this the case?

I am not sure but I plan to bug fix #154 on this PR too. Currently researching about it

JonnyBurger:
@Iamshankhadeep Do we understand why `absoluteFrame === 0` is solving the issue? Or is it rather random? 🤔

JonnyBurger:
Wow, I found more bugs!

- Video `currentTime` was calculated completely the wrong way! Instead of `currentFrame / (1000 / videoConfig.fps)` it should just be `currentFrame / videoConfig.fps`. This led to huge timeshifts even in my own projects! Very embarassing but never too late to address it.
- Bug mentioned by @yuta0801 was not fully fixed, we need to check `videoRef.current.ended` to make sure we don't restart the video too early
- Composition could be 1 frame too long in the preview, although timeline slider would not allow it the jump frame arrows would allow it!
- Timeline elements were also off by one frame
- Sequence was indeed 1 frame too long, thanks for fixing that! I put it behind a feature flag that can be turned off in case we need to put out 1.x releases in the future or before 2.x releases.
- Potentially had the same bug in Audio, applied the same fixes there.

I will do some more testing especially for Audio, and then merge it within a few days.

JonnyBurger (approved):

PR #176, opened 2021-03-08
Use multiple browser instances for concurrency

monophthongal:
This PR modifies the renderer to use a separate puppeteer Browser for each renderer page. In my tests this improves screenshot capture time by 2x or more. 

Overall render time improves significantly, although the time required to bundle the project and stitch frames together is still a factor. Here are some benchmarks using the Spotify Wrapped example (run on my development machine):

_Before change:_

    Benchmark #1: npm run build -- --overwrite --concurrency=16
      Time (mean ± σ):     54.226 s ±  1.831 s    [User: 148.745 s, System: 14.465 s]
      Range (min … max):   52.386 s … 56.792 s    5 runs
      
    Render performance (w/ --concurrency=16):
      activate-target => 3.194736842105263 (n = 570)
      capture => 439.1824561403509 (n = 570)
      save => 1.712280701754386 (n = 570)
     
    Benchmark #2: npm run build -- --overwrite --concurrency=8
      Time (mean ± σ):     49.707 s ±  4.794 s    [User: 136.615 s, System: 12.910 s]
      Range (min … max):   45.979 s … 55.199 s    5 runs
     
    Render performance (w/ --concurrency=8):
      activate-target => 1.4228070175438596 (n = 570)
      capture => 196.3421052631579 (n = 570)
      save => 1.6245614035087719 (n = 570)

_After change:_

    Benchmark #1: npm run build -- --overwrite --concurrency=16
      Time (mean ± σ):     38.599 s ±  1.481 s    [User: 207.078 s, System: 26.778 s]
      Range (min … max):   36.112 s … 40.034 s    5 runs

    Render performance (concurrency=16):
      activate-target => 2.7982456140350878 (n = 570)
      capture => 161.7719298245614 (n = 570)
      save => 13.478947368421053 (n = 570)

    Benchmark #2: npm run build -- --overwrite --concurrency=8
      Time (mean ± σ):     35.942 s ±  4.464 s    [User: 159.159 s, System: 18.091 s]
      Range (min … max):   32.843 s … 43.773 s    5 runs

    Render performance (concurrency=8):
      activate-target => 1.0017543859649123 (n = 570)
      capture => 89.09298245614035 (n = 570)
      save => 2.3228070175438598 (n = 570)

Notes:

* The default, auto-detected setting for `concurrency` ends up being too high with this change, but I haven't come up with a good way to calculate the optimal number yet. On my machine, remotion tries to use 24 renderer instances by default, but performance tops out at a much lower number of instances. Spinning up more browsers makes it slower (probably because of memory?). Also, puppeteer emits a warning after 11 browser instances, but you can disable that.
* Sending `Target.activateTarget` before taking screenshots (the activate-target benchmark above) may not be needed here since there is only 1 page per browser instance.

Let me know if you have any questions or suggestions for improvements.


JonnyBurger:
Hey @monophthongal!

Thanks a lot for this awesome PR, I really love it!
I did some testing myself and I can confirm that if you set the concurrency to `require('os').cpus().length`, it acts weird and crashes. I added a `<Video/>` element, and if I load the video in 16 browsers, it will not load everywhere and time out!

So while I want to merge this, I need some more time to explore the implications of this. Maybe the ideal will be 4x4? (4 browsers at 4 pages). Let's experiment a bit and see what works well.
Right now I think we cannot ship it in the current state because of some issues it surfaces (I am actively debugging these to see if we can make it more concurrency safe). Nonetheless I will over the next week or two run some more benchmarks and hopefully we can ship this great performance improvement! 🙏

tomByrer:
@monophthongal I like your idea!  Any progress on the suggested changes please?

JonnyBurger:
I guess it's up to me to approve it, I would like to evaluate it again and make it an opt-in feature, but I have some reservations still because it seems that if you go too crazy with it, it becomes really unstable. So I would like to prevent configuration which don't seem sane.

Hope to make a decision on this soon!

iamkhalidbashir:
Waiting for this...

JonnyBurger:
I have opened #527, a PR that aims to solve a similar problem. My hypothesis that the slowdown is because tabs are being marked as inactive if they go into the background and this PR improves this by opening more browser instances. However this increases RAM usage by a lot.

By instead cycling through the tabs, we decrease RAM usage and hopefully achieve a similar performance improvement.

This is just a hypothesis though, now we need to run some benchmarks!

JonnyBurger:
I did some measuring though and this is what I found (calculating only the frame rendering time, doing 3 runs each, using the Spotify Wrapped example):

image

First thing to note is that since this PR has been opened, how Remotion is rendering stuff has changed a lot, on all benchmarks, it's much faster than before. Rather than 400ms per image, I am now seeing 50ms per image and 90ms if a lot of cores are being used. It also means that instantiating multiple browsers does not yield a performance benefit anymore on my test machines. I used a 128GB RAM Intel i9 Hackintosh and a 16GB Macbook M1 to test and there were no significant improvements to even slight performance decreases, also considering that it probably takes more RAM.

I am going to close it in favor of #527, which also doesn't seem to improve the performance in this case like I expected, but does so in WebGL contexts and other scenarios where Chrome might bottleneck when a tab goes into the background.

Nonetheless, a big thank you for opening this PR and investigating this issue. We've learned a lot from it and I think at the time, it was a PR with a lot of merit. Apologies again for taking so long to deal with it and hope you'll continue to contribute!

PR #180, opened 2021-03-12
fix(preview): timeline slider tweaks

jeetiss:
use `transform translate` instead of `left` because it is better for perf and doesn't create weird alignment glitches like this

https://user-images.githubusercontent.com/6726016/110927701-05e0f180-8337-11eb-956f-97d97fd658b6.mov

and align timeline cursor to center


JonnyBurger (approved):

JonnyBurger:
I noticed that a few days ago too, great improvement! Thanks a lot! 💙

PR #181, opened 2021-03-12
fix(preview): isolate FpsCounter

jeetiss:
- isolate FpsCounter now it uses only `Internals`
- fix bug with NaN
- show and hide fps immediately after play/pause


JonnyBurger (approved):

JonnyBurger:
Nice! Love that now the FPS disappears immediately when you pause. Thanks again!!

PR #183, opened 2021-03-12
chore(examples): remove ts-node, run via remotion

jeetiss:
`ts-node` is unnecessary because everything works fine with new `watch` command 

send kudos and love ❤️ to @yuta0801


JonnyBurger (approved):

JonnyBurger:
That's right! Thanks for improving 💙 and also once again thanks to @yuta0801 💙  this was awesome

PR #190, opened 2021-03-14
Improve testing (core)

j0nezz:
This PR was done together with @calvin-f and @dydent with the goal of improving the quality and coverage of the core package tests.

A few notes on things we changed outside the actual tests:

- The `interpolate()` function was accepting all monotonic non-decreasing input ranges. However, this can lead to ambiguous interpolations e.g. `interpolate(0.75, [0, 1, 1], [0, 2, 0])`. We have therefore narrowed down the conditions for input ranges to only accept strictly monotonic non-decreasing arrays.
- The check if `typeof value !== 'number'` in `frame-range.ts` would always return true since the values were casted to a number before. Changed it to check if the value is `NaN` instead (e.g. if a string is passed)
- Added additional check at the beginning for valid ImageFormat in `image-format.ts`

We ran some coverage stats (just with --coverage flag of jest) and based on them, we improved branch coverage for the core by more than 20%.

So we hope this PR helps, if you have some comments / change requests let us know :)

### DIFF
--- a/packages/core/src/test/codec.test.ts
+++ b/packages/core/src/test/codec.test.ts
@@ -1,4 +1,5 @@
 import {
+	CodecOrUndefined,
 	getFinalOutputCodec,
 	getOutputCodecOrUndefined,
 	setCodec,
@@ -8,18 +9,12 @@ import {
 // getFinalOutputCodec
 
 test('Codec tests valid codec input', () => {
-	expect(
-		getFinalOutputCodec({codec: 'h264', emitWarning: true, fileExtension: ''})
-	).toEqual('h264');
-	expect(
-		getFinalOutputCodec({codec: 'h265', emitWarning: true, fileExtension: ''})
-	).toEqual('h265');
-	expect(
-		getFinalOutputCodec({codec: 'vp8', emitWarning: true, fileExtension: ''})
-	).toEqual('vp8');
-	expect(
-		getFinalOutputCodec({codec: 'vp9', emitWarning: true, fileExtension: ''})
-	).toEqual('vp9');
+	const values: CodecOrUndefined[] = ['h264', 'h265', 'vp8', 'vp9'];
+	values.forEach((entry) =>
+		expect(
+			getFinalOutputCodec({codec: entry, emitWarning: true, fileExtension: ''})
+		).toEqual(entry)
+	);
 });
 
 test('Codec tests undefined codec input with known extension', () => {
@@ -69,15 +64,9 @@ test('Codec tests setOutputFormat', () => {
 // setCodec
 
 test('Codec tests setOutputFormat', () => {
-	expect(getOutputCodecOrUndefined()).toEqual(undefined);
-	setCodec('h264');
-	expect(getOutputCodecOrUndefined()).toEqual('h264');
-	setCodec('h265');
-	expect(getOutputCodecOrUndefined()).toEqual('h265');
-	setCodec('vp8');
-	expect(getOutputCodecOrUndefined()).toEqual('vp8');
-	setCodec('vp9');
-	expect(getOutputCodecOrUndefined()).toEqual('vp9');
-	setCodec(undefined);
-	expect(getOutputCodecOrUndefined()).toEqual(undefined);
+	const values: CodecOrUndefined[] = ['h264', 'h265', 'vp8', 'vp9', undefined];
+	values.forEach((entry) => {
+		setCodec(entry);
+		expect(getOutputCodecOrUndefined()).toEqual(entry);
+	});
 });
--- a/packages/core/src/test/codec.test.ts
+++ b/packages/core/src/test/codec.test.ts
@@ -11,7 +11,7 @@ test('Codec tests valid codec input', () => {
 	const values: CodecOrUndefined[] = ['h264', 'h265', 'vp8', 'vp9'];
 	values.forEach((entry) =>
 		expect(
-			getFinalOutputCodec({codec: entry, emitWarning: true, fileExtension: ''})
+			getFinalOutputCodec({codec: entry, emitWarning: false, fileExtension: ''})
 		).toEqual(entry)
 	);
 });
@@ -20,14 +20,14 @@ test('Codec tests undefined codec input with known extension', () => {
 	expect(
 		getFinalOutputCodec({
 			codec: undefined,
-			emitWarning: true,
+			emitWarning: false,
 			fileExtension: 'webm',
 		})
 	).toEqual('vp8');
 	expect(
 		getFinalOutputCodec({
 			codec: undefined,
-			emitWarning: true,
+			emitWarning: false,
 			fileExtension: 'hevc',
 		})
 	).toEqual('h265');
--- a/packages/core/src/test/codec.test.ts
+++ b/packages/core/src/test/codec.test.ts
@@ -1,0 +1,83 @@
+import {
+	getFinalOutputCodec,
+	getOutputCodecOrUndefined,
+	setCodec,
+	setOutputFormat,
+} from '../config/codec';
+
+// getFinalOutputCodec
+
+test('Codec tests valid codec input', () => {
+	expect(
+		getFinalOutputCodec({codec: 'h264', emitWarning: true, fileExtension: ''})
+	).toEqual('h264');
+	expect(
+		getFinalOutputCodec({codec: 'h265', emitWarning: true, fileExtension: ''})
+	).toEqual('h265');
+	expect(
+		getFinalOutputCodec({codec: 'vp8', emitWarning: true, fileExtension: ''})
+	).toEqual('vp8');
+	expect(
+		getFinalOutputCodec({codec: 'vp9', emitWarning: true, fileExtension: ''})
+	).toEqual('vp9');
+});
+
+test('Codec tests undefined codec input with known extension', () => {
+	expect(
+		getFinalOutputCodec({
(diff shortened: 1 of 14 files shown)


packages/core/src/test/crf.test.ts
```diff
@@ -0,0 +1,148 @@
...
+test('crf tests getDefaultCrfForCodec', () => {
+	// input codec, output
+	const valuesA = [
+		['h264', 18],
+		['h265', 23],
+		['vp8', 9],
+		['vp9', 28],
+	];
+	valuesA.forEach((entry) =>
+		// @ts-expect-error
```
JonnyBurger: Unnecessary Typescript error suppression, you can write `valuesA = [...] as [Codec, number][]`

packages/core/src/test/crf.test.ts
```diff
@@ -0,0 +1,148 @@
...
+test('crf tests getValidCrfRanges', () => {
+	// input crf, input codec, valid range
+	const valuesA = [
+		['h264', [0, 51]],
+		['h265', [0, 51]],
+		['vp8', [4, 63]],
+		['vp9', [0, 63]],
+	];
+	valuesA.forEach((entry) =>
+		// @ts-expect-error
```
JonnyBurger: here too, you can use `as [Codec, [number, number]]` and the TS error will not appear

packages/core/src/test/crf.test.ts
```diff
@@ -0,0 +1,148 @@
...
+		[20, 'vp8'],
+		[4, 'vp8'],
+		[63, 'vp8'],
+		[20, 'vp9'],
+		[0, 'vp9'],
+		[63, 'vp9'],
+	];
+	valuesA.forEach((entry) =>
+		expect(
+			// @ts-expect-error
```
JonnyBurger: ts-expect-error can be avoided

packages/core/src/test/crf.test.ts
```diff
@@ -0,0 +1,148 @@
...
+		[80, 'vp8', [4, 63]],
+		[3, 'vp8', [4, 63]],
+		[64, 'vp8', [4, 63]],
+		[80, 'vp9', [0, 63]],
+		[-1, 'vp9', [0, 63]],
+		[64, 'vp9', [0, 63]],
+	];
+	valuesB.forEach((entry) =>
+		expectToThrow(
+			// @ts-expect-error
```
JonnyBurger: TS error can be avoided (2x)

packages/core/src/test/crf.test.ts
```diff
@@ -0,0 +1,148 @@
...
+		[undefined, 'h264', 18],
+		[20, 'h265', 20],
+		[undefined, 'h265', 23],
+		[20, 'vp8', 20],
+		[undefined, 'vp8', 9],
+		[20, 'vp9', 20],
+		[undefined, 'vp9', 28],
+	];
+	valuesA.forEach((entry) => {
+		// @ts-expect-error
```
JonnyBurger: more TS errors which can be avoided, also below

packages/core/src/test/easing.test.ts
```diff
@@ -0,0 +1,52 @@
+import {Easing} from '../easing';
+
+const numbersToTest = [0, 0.2, 0.5, 0.7, 1];
+
+test('Easing linear', () => {
+	const easingIn = Easing.in(Easing.linear);
+	numbersToTest.forEach((n) => expect(easingIn(n)).toBe(n));
+
+	const easingOut = Easing.out(Easing.linear);
+	numbersToTest.forEach((n) => expect(easingOut(n)).toBeCloseTo(n));
```
JonnyBurger: should be able to use `toBe`, not `toBeCloseTo` I think.

out() is defined as `1 - easing(1 - t)` and linear/easing is just defined as `t => t`.

If we shorten that we get `1 - (1 - t)` --> `1 - 1 + 1` --> `t`.
So `toBe` should work, otherwise we discovered a bug.

packages/core/src/test/codec.test.ts
```diff
@@ -0,0 +1,72 @@
...
+	setOutputFormat,
+} from '../config/codec';
+
+// getFinalOutputCodec
+
+test('Codec tests valid codec input', () => {
+	const values: CodecOrUndefined[] = ['h264', 'h265', 'vp8', 'vp9'];
+	values.forEach((entry) =>
+		expect(
+			getFinalOutputCodec({codec: entry, emitWarning: true, fileExtension: ''})
```
JonnyBurger: Please set `emitWarning` to `false` everywhere, otherwise it will log noise during tests.

packages/core/src/test/codec.test.ts
```diff
@@ -0,0 +1,72 @@
...
+	).toEqual('h264');
+});
+
+// setOutputFormat
+
+test('Codec tests setOutputFormat', () => {
+	expect(getOutputCodecOrUndefined()).toEqual(undefined);
+	setOutputFormat('mp4');
+	expect(getOutputCodecOrUndefined()).toEqual('h264');
+	setOutputFormat('png-sequence');
```
JonnyBurger: I would prefer to remove the tests for `setOutputFormat` as it's deprecated and been undocumented, and now it logs warnings during tests. If you really need it for test coverage, you can add a new `(newLegacyFormat: LegacyFormat, emitWarnings = true) =>` argument similarly to how it works above

packages/core/src/test/image-format.test.ts
```diff
@@ -0,0 +1,49 @@
...
+	expect(() =>
+		validateSelectedPixelFormatAndImageFormatCombination('yuva420p', 'png')
+	).not.toThrow();
+});
+test('It should not throw if pixel and image format combination is valid', () => {
+	expect(() =>
+		validateSelectedPixelFormatAndImageFormatCombination('yuv420p', 'png')
+	).not.toThrow();
+});
+test('It should not throw if pixel and image format combination is valid', () => {
```
JonnyBurger: Can we have unique names for tests?

packages/core/src/test/interpolate.test.ts
```diff
@@ -15,42 +17,63 @@ test('Must be the same length', () => {
...
-	}, /inputRange must be monotonically non-decreasing/);
+	}, /inputRange must be strictly monotonically non-decreasing/);
+	expectToThrow(() => {
+		interpolate(0.75, [0, 1, 1], [0, 2, 0]);
+	}, /inputRange must be strictly monotonically non-decreasing/);
 });
 
-test('Output range can be monotonically decreasing', () => {
+test('Output range strictly monotonically decreasing', () => {
 	expect(interpolate(0.75, [0, 0.5, 1], [0, 2, 0])).toEqual(1);
```
JonnyBurger, line 36: The test name is (and was) kind of bad here, what we are testing here is that output range doesn't need to have any monotonic increase/decrease properties at all, it works if it is increasing then decreasing again.

packages/core/src/test/sequence-validation.test.tsx
```diff
@@ -21,10 +21,12 @@ test('It should allow null as children', () => {
...
-		/You passed to durationInFrames an argument of type undefined, but it must be a number./
-	);
+test('It should allow undefined as children', () => {
+	expect(() =>
+		render(
+			<Sequence durationInFrames={100} from={0}>
+				{undefined}
+			</Sequence>
+		)
+	).not.toThrow();
```
JonnyBurger, line 31: Can you add the previous test back? It tested for when you were missing the `durationInFrames` property. I know the test name was bad, maybe we can change it to "Should enforce durationInFrames props"

JonnyBurger (changes requested):
Thanks a loooot! 💙  This is so valuable and reinforces the core of Remotion so much! The PR looks very carefully coded overall, I appreciate it a lot! I have a few comments still, they shouldn't take too long, if you can fix these, I am merging. In any case I will also give you a shoutout in the next blog post!

packages/core/src/test/sequence-validation.test.tsx
calvin-f: Previously this third test was an exact duplicate of the first test. Now, it's testing something different. 

The name of the first test should be explanation enough, I think: 'It should throw if Sequence has missing duration'

packages/core/src/test/easing.test.ts
j0nezz: You're right, if we assume `t=1` then it works. But if we use values like `0.3` or `0.2` it gives some odd results due to floating point calculations. You can reproduce it by just typing `1-1*(1-0.3)` in your JS console. I think a quick workaround would be to swap the test of `0.3` with `0.4` as this does not suffer from floating point precision issues

packages/core/src/test/sequence-validation.test.tsx
JonnyBurger: You are right! Thanks for correcting that.

packages/core/src/test/easing.test.ts
JonnyBurger: You are right! Thanks for the example, now it makes sense to me!

j0nezz:
All done @JonnyBurger, if you have time you can quickly have a look! :)

JonnyBurger:
Amazing! I'll have a look this evening 🙂

JonnyBurger (approved):

JonnyBurger:
Thanks a lot guys! Much appreciated! 💙

PR #195, opened 2021-03-16
added startAt and endAt props for video

### LINKED ISSUE #150: Some way to let the video start at a certain time
Discussed with @jonpacker on Discord, it would be cool to have a more obvious API for starting a video at 2 seconds:

way to video

Iamshankhadeep:
fixes #150

### DIFF
--- a/packages/core/src/video/Video.tsx
+++ b/packages/core/src/video/Video.tsx
@@ -13,10 +13,7 @@ export const Video: React.FC<RemotionVideoProps & RemotionMainVideoProps> = (
 		const endAtFrameNO = endAt ? endAt : Infinity;
 		console.log(0 - startAtFrameNo, endAtFrameNO - startAtFrameNo, 'wow');
 		return (
-			<Sequence
-				from={0 - startAtFrameNo}
-				durationInFrames={endAtFrameNO - startAtFrameNo}
-			>
+			<Sequence from={0 - startAtFrameNo} durationInFrames={endAtFrameNO}>
 				<Video {...otherProps} />
 			</Sequence>
 		);
--- a/packages/core/src/video/Video.tsx
+++ b/packages/core/src/video/Video.tsx
@@ -1,11 +1,32 @@
 import React from 'react';
-import {RemotionVideoProps} from './props';
+import {Sequence} from '../sequencing';
+import {RemotionMainVideoProps, RemotionVideoProps} from './props';
 import {VideoForDevelopment} from './VideoForDevelopment';
 import {VideoForRendering} from './VideoForRendering';
 
-export const Video: React.FC<RemotionVideoProps> = (props) => {
+export const Video: React.FC<RemotionVideoProps & RemotionMainVideoProps> = ({
+	startAt,
+	endTo,
+	...props
+}) => {
+	const startAtFrameNo = startAt ? startAt : 0;
+	const endToFrameNO = endTo ? endTo : Infinity;
 	if (process.env.NODE_ENV === 'development') {
-		return <VideoForDevelopment {...props} />;
+		return (
+			<Sequence
+				from={0 - startAtFrameNo}
+				durationInFrames={endToFrameNO - startAtFrameNo}
+			>
+				<VideoForDevelopment {...props} />;
+			</Sequence>
+		);
 	}
-	return <VideoForRendering {...props} />;
+	return (
+		<Sequence
+			from={0 - startAtFrameNo}
+			durationInFrames={endToFrameNO - startAtFrameNo}
+		>
+			<VideoForRendering {...props} />;
+		</Sequence>
+	);
 };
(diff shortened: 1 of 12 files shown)


packages/core/src/video/VideoForDevelopment.tsx
```diff
@@ -40,6 +32,7 @@ export const VideoForDevelopment: React.FC<RemotionVideoProps> = (props) => {
 
 		if (!playing || absoluteFrame === 0) {
 			videoRef.current.currentTime = currentTime;
+			videoRef.current.pause();
 		}
 
 		if (videoRef.current.paused && !videoRef.current.ended && playing) {
```
JonnyBurger: I think the changes in this file are not necessary anymore right?

packages/core/src/video/Video.tsx
```diff
@@ -1,11 +1,32 @@
...
-	return <VideoForRendering {...props} />;
+	return (
+		<Sequence
+			from={0 - startAtFrameNo}
+			durationInFrames={endAtFrameNO - startAtFrameNo}
+		>
+			<VideoForRendering {...props} />;
+		</Sequence>
+	);
 };
```
JonnyBurger: How about recursion? It would remove redudancy and be easier to reason about right?

```tsx
const {startAt, endAt, ...otherProps} = props
if (startAt || endAt) {
  return (
    <Sequence
      from={0 - startAtFrameNo}
	  durationInFrames={endAtFrameNO - startAtFrameNo}
    >
      <Video {...otherProps}>
    </Sequence>
  )
}
// Logic we had until now
```

packages/core/src/video/VideoForRendering.tsx
```diff
@@ -20,7 +21,9 @@ export const VideoForRendering: React.FC<RemotionVideoProps> = ({
 		if (!videoRef.current) {
 			return;
 		}
-		const frameInSeconds = frame / videoConfig.fps;
+		const frameInSeconds = FEATURE_FLAG_V2_BREAKING_CHANGES
+			? (frame + 1) / videoConfig.fps
+			: frame / videoConfig.fps;
```
JonnyBurger: Let's make this change in the audio branch instead, then we won't have conflicts

packages/docs/docs/video.md
```diff
@@ -11,7 +11,7 @@ Videos with audio are not supported - yet! Audio will be muted in the final rend
 
 ## API / Example
 
-Use an import or require to load an audio file and pass it as the `src` prop. All the props that the native `<video>` element accepts (except `autoplay` and `controls`) will be forwarded (but of course not all are useful for Remotion). This means you can use all CSS to style the video.
+Use an import or require to load an video file and pass it as the `src` prop. All the props that the native `<video>` element accepts, `startAt` (optional) and `endAt` (optional) props (except `autoplay` and `controls`) will be forwarded (but of course not all are useful for Remotion). This means you can use all CSS to style the video.
```
JonnyBurger: Nice! Makes it sound like `startAt` prop will be forwarded to the video element, which shouldn't happen. Can we leave this sentence and then mention `startAt` in a new paragraph? It will probably also be easier to read and digest for the reader.

JonnyBurger (changes requested):

JonnyBurger:
Thanks a lot! Super awesome stuff. I added some comments. When we are happy, we could also add it to the audio element!

packages/core/src/video/VideoForDevelopment.tsx
Iamshankhadeep: Yup changes in this file is not necessary.

packages/core/src/video/VideoForRendering.tsx
Iamshankhadeep: okay I'll revert this changes then.

packages/core/src/video/Video.tsx
Iamshankhadeep: will surely do,

packages/docs/docs/video.md
Iamshankhadeep: done

JonnyBurger (changes requested):
Looking good! Now we can refine.

Let's do the following to further increase stability and user experience:

- Remove the console.log
- Add the same behavior to <Audio/> component
- Add documentation for the props in audio component
- add some runtime validation if people pass in parameters that are not numbers (for non-typescript users), negative numbers, NaN, or if they pass `Infinity` as start numbers
- Add some tests to validate that errors are being thrown - you can probably copy paste it from `packages/core/src/test/sequence-validation.test.tsx` and adapt it to work for the Video component. Also add it to the Audio component
- Test the edge case: What if `endAt` is 0? It will lead to `durationInFrames={Infinity}`. this should probably be an error
- Also validate that endAt > startAt

I know it's a lot extra, but we can take our time to make it as robust as possible and help users if they are writing bugs!

Iamshankhadeep:
> I know it's a lot extra, but we can take our time to make it as robust as possible and help users if they are writing bugs!

Done with everything. Have a look at it now. @JonnyBurger

JonnyBurger (approved):

PR #197, opened 2021-03-19
Using userProps at root component

FelippeChemello:
As I said in this issue https://github.com/JonnyBurger/remotion/issues/196 I would like to user props at root component, so I pass it in Root and made some corretions at getCompositions to allow passing userProps to validation, so it can allow the dinamically component to be rendered.

### DIFF
--- a/packages/bundler/src/index.ts
+++ b/packages/bundler/src/index.ts
@@ -16,6 +16,7 @@ export const startServer = async (
 	userDefinedComponent: string,
 	options?: {
 		webpackOverride?: WebpackOverrideFn;
+		inputProps?: object;
 	}
 ): Promise<number> => {
 	const app = express();
@@ -30,6 +31,7 @@ export const startServer = async (
 		environment: 'development',
 		webpackOverride:
 			options?.webpackOverride ?? Internals.getWebpackOverrideFn(),
+		inputProps: options?.inputProps ?? null,
 	});
 	const compiler = webpack(config);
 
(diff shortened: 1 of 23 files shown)


JonnyBurger:
I will probably want to make the following changes (I will do them or you can also help if you like to):

- Also pass the target composition name to getCompositions, in case you have multiple of them, it's good to know which one is being rendered before you make any calculation
- Make sure you can also pass --props it in debug mode
- turning the parameters 2 and 3 of getCompositions into an object, almost nobody wants to pass in a browser preference, user props will be the much more popular

I am also keeping my mind open if there is a more elegant way to do it (might have to break some existing conventions for it). I would love it if we can come up with a solution that is more typesafe and intuitive.

I'll return to this PR once audio support is merged, but this one will also make the next release!

FelippeChemello:
Sure, I can work on it. But have some doubts.

How can I activate debug mode? I haven't find it on documentation.

What does "pass in a browser preference" means? Do you means like unifying the secound and thirth element into an object?

JonnyBurger:
Thanks a lot! 🙏🏼
By debug mode I just mean the preview, the player that opens when you type `yarn start`.

And yes, let's just concatenate the second and third parameter into an object.

FelippeChemello:
Ok.

Props in debug mode I will understand as passing props directly in URI, that's right? Or it need to be possible to pass on CLI also?
As soon as I finish it I will contact you.

JonnyBurger:
I would say also pass it on the CLI so one can have the same behavior on both commands.
So it's a bit complicated to pass those, I wouldn't necessarily do it over URL, but inject it into the page using some Webpack plugin. I can take care of this!

I'll work a lot on Remotion next week, hope I can also contribute to getting this merged :)

JonnyBurger:
I've finished the implementation, one change: Instead of retrieving the props in the root, you can get them using the new `getInputProps` API which will be documented on the website.

Also now you can pass --props while still in development mode.

Thanks a lot for bringing this up and sending this PR! 💙

PR #199, opened 2021-03-19
Better --props api

### LINKED ISSUE #119: Better API for --props which is more convienient for windows
Windows removes all `"` from the command line, so `--props` is just unusable on Windows. Need to improve the API!

Iamshankhadeep:
fixes #119


JonnyBurger:
Can we allow the previous behavior too? Like, if --props looks like a file path, use that, if not, parse it as JSON?

Seems harsh to break existing workflows, for simple props structures I actually prefer to inline the JSON.

JonnyBurger (approved):

PR #201, opened 2021-03-20
feat: allow resize of timeline

arthurdenner:
Hi! 👋🏽 

As I'm creating videos, I think the timeline occupies too much space sometimes.

This PR improves the editor by allowing the user to resize the timeline as they see fit.
It also improves the layout a bit by preventing tracks from overflowing and adding a cursor to the slider - small tweaks.

I have tested this with the `HelloWorld` example and it looks like this:
https://user-images.githubusercontent.com/13774309/111884339-70420380-89c1-11eb-8260-6dd14d292f58.mp4

I hope you like it and let me know if you find any issues.


JonnyBurger:
😍😍 absolutely love this idea! Thanks a lot for implementing this! I will review it tomorrow and play with it, can't wait!

packages/cli/src/editor/components/TimelineDragHandler.tsx
```diff
@@ -1,33 +1,27 @@
...
 import {useWindowSize} from '../hooks/use-window-size';
 
 const Container = styled.div`
+	flex: 1;
 	position: relative;
-	padding-left: ${TIMELINE_LEFT_PADDING}px;
-	padding-right: ${TIMELINE_RIGHT_PADDING}px;
-	padding-top: 20px;
-	margin-top: -10px;
+	padding: ${TIMELINE_PADDING}px;
```
JonnyBurger, line 10: We replaced padding-left and padding-right with padding on all 4 sides. Was that intentional?

packages/example/src/Features/index.tsx
```diff
@@ -1,9 +1,22 @@
...
 
-const Features: React.FC = () => {
+const Features: FC = () => {
+	const [canPlay, setCanPlay] = useState<boolean>();
 	const tray = require('./tray.webm');
 	const watermelon = require('./watermelon.webm');
 	const textstickers = require('./textstickers.webm');
+
+	useEffect(() => {
+		const videoEl = document.createElement('video');
```
JonnyBurger: Out of curiosity, which browser do you use that doesn't support playing WebM videos?

packages/example/src/ReactNativeWeb/index.tsx
```diff
@@ -3,7 +3,7 @@ import {Text, View} from 'react-native';
 
 const ReactNativeWeb: React.FC = () => {
 	return (
-		<View>
+		<View style={{backgroundColor: 'white', width: '100%'}}>
```
JonnyBurger: any reason why you removed the white background?

packages/bundler/web/index.html
```diff
@@ -5,6 +5,10 @@
 		<meta name="viewport" content="width=device-width, initial-scale=1.0" />
 		<link rel="preconnect" href="https://fonts.gstatic.com" />
 		<link rel="icon" type="image/png" href="/favicon.png" />
+		<link
+			href="https://unpkg.com/react-reflex@4.0.0/styles.css"
+			rel="stylesheet"
+		/>
```
JonnyBurger: Let's not load it from the internet, it might break Remotion working offline and makes the preview unnecessarily slower. We can probably import the CSS in the Preview using an import statement.

JonnyBurger (changes requested):

JonnyBurger:
Thanks a lot! Looking mostly good, but I added some comments. You are welcome to resolve them or if it doesn't happen, I will do it in the foreseeable time and add it as a new feature of Remotion 2.0, the next release :)

JonnyBurger:
I've tried it out and realized it would also be nice if Remotion would remember the position (by saving into localstorage or something).

packages/example/src/Features/index.tsx
arthurdenner: Safari doesn't play it and throws errors.

packages/cli/src/editor/components/TimelineDragHandler.tsx
arthurdenner: Yes, it was. From what I've tested, there was no need for different padding. It changes the layout a bit but I think it looks good. Do you prefer to revert it?

arthurdenner:
I forgot to add details for the changes in the `example` package - sorry about that - so I removed the commit and it'll open a new PR later proposing them since they don't really belong to this feature proposal. Then I can elaborate on the changes without messing up the discussion here. 😄

JonnyBurger:
Thanks for pushing some changes! Tomorrow is my next free day where I'll take another look 😊

packages/bundler/web/index.html
arthurdenner: When I try to `import 'react-reflex/styles.css';`, I get this error: `Error: Cannot find module 'react-reflex/styles.css'`

image

We need to tweak the Webpack config somewhere. I'll try to have a look later but if you already know the answer, please let me know.

packages/cli/src/editor/state/timeline.ts
```diff
@@ -0,0 +1,15 @@
...
+};
+
+const loadTimelineFlex = (): number => {
+	const item = localStorage.getItem(localStorageKey);
+
+	return item ? parseFloat(item) : 0.2;
+};
+
+export const useTimelineFlex = (): [number, (value?: number) => void] => {
+	return [loadTimelineFlex(), persistTimelineFlex];
```
arthurdenner: Since `ReflexElement` is not a [controlled component](https://reactjs.org/docs/forms.html#controlled-components), there's no need to use `useState` or something similar that would trigger a re-render.

packages/bundler/web/index.html
JonnyBurger: I see, tried it out and definitely tricky.. ended up inlining and adjusting Webpack config.

packages/cli/src/editor/components/TimelineDragHandler.tsx
JonnyBurger: OK, then LGTM!

JonnyBurger:
Thanks a lot for this PR again! 💙 

I have managed to improve on it a bit more:

- Now you can also resize the left sidebar/canvas panels
- Removed react-reflex and replaced it with a custom solution with zero dependencies
- While you drag or hover, the bars turn blue, like they do in the newest VS Code update

You deserve a big shoutout in the Remotion 2.0 blog post! :) :)

JonnyBurger (approved):

arthurdenner:
Thanks, I'm glad you liked the idea. I'd say that the timeline should be even smaller though. Most of the time, I want to focus on the video once the timeline is set.

JonnyBurger:
@arthurdenner Got you! I made it that you can make the timeline up to twice as small in the audio branch.

PR #208, opened 2021-03-25
Adding new package @remotion/player

### LINKED ISSUE #133: @remotion/player package for embedding Remotion videos in websites

Iamshankhadeep:
fixes #133


Iamshankhadeep:
@JonnyBurger can you have a look at it and let me know if it's okay or not?

JonnyBurger:
Thanks for starting this! I really need to finish audio support first before I can look at more issues. but a few things I noticed (and I am sorry that I might have given you wrong information).

`core` should in no way depend on `player`. Those functions who are needed in `core` need to stay in `core`. Rather, `player` should depend on `core`. We need to extract some functions from `cli` into player instead. What comes to mind is the PlayPause logic in CLI, which will loop through all frames.

Here we need to separate the UI and logic - the UI should stay in `cli` but the effects should come into `player` if that makes sense.

Iamshankhadeep:
> Thanks for starting this! I really need to finish audio support first before I can look at more issues. but a few things I noticed (and I am sorry that I might have given you wrong information).
> 
> `core` should in no way depend on `player`. Those functions who are needed in `core` need to stay in `core`. Rather, `player` should depend on `core`. We need to extract some functions from `cli` into player instead. What comes to mind is the PlayPause logic in CLI, which will loop through all frames.
> 
> Here we need to separate the UI and logic - the UI should stay in `cli` but the effects should come into `player` if that makes sense.

Ah now I see. got it will start working on it asap.

packages/player/src/test/index.test.ts
```diff
@@ -0,0 +1,13 @@
...
+test('It should throw an error', () => {
+	expect(() => {
+		const [
+			toggle,
+			frameBack,
+			frameForward,
+			onKeyPress,
+			isLastFrame,
+		] = usePlaybackTime();
+	}).toThrow();
```
JonnyBurger, line 6: Why do we expect it to throw an error?

packages/player/tsconfig.json
```diff
@@ -0,0 +1,16 @@
...
+		"composite": true,
+		"rootDir": "./src",
+		"outDir": "./dist",
+		"skipLibCheck": true,
+		"jsx": "react-jsx"
+	},
+	"references": [
+		{"path": "../core"},
+		{"path": "../bundler"},
+		{"path": "../renderer"}
```
JonnyBurger: May remove bundle and renderer from references, neither should be accessed by renderer.

JonnyBurger:
Looking good!
I need to think about what the next steps are.

Probably it's a good idea to create a new package `player-example` and there try to initialize this example: https://codesandbox.io/s/remotion-car-slideshow-002-p2vzu?file=/src/VideoPreview.tsx

And then from there try to figure out which functions to share and how to create the most elegant API.
Do you have any thoughts on it?

packages/player/src/test/index.test.ts
Iamshankhadeep: because we are using `playing` and `setPlaying`  ```const {playing} = useContext(TimelineContext);``` but there is not any Provider so we are expecting an Error from it.

packages/player/tsconfig.json
Iamshankhadeep: Okay Noted.

packages/player/src/test/index.test.ts
JonnyBurger: Got it, updated the test name to reflect this 👍

JonnyBurger:
This is great! Already way further and better than I expected 😍 ! Great job!

It took some playing with it until I realized what the next steps are.
The general idea is to emulate the basic API of the [HTML5 video element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video)

- Right now the play / pause button is *inside* the player, but the player should be externally controllable. But this is still good, let's allow people to pass a `controls` element, if they don't want to manually build a controls UI, they can just use the default. I have already implemented the `play()` and `pause()` methods, but we need more, like `seekTo()`, and `currentTime`.

- Let's implement a `loop` property, by default true

- We need to build out the default UI, like Timeline bar, better play / pause icons, etc

- We need to allow people to style the container. Right now there is no way of doing it and currently the player is always the same size as video. So for example you can't play an HD video on mobile because it's too big.

- Player should be handled completely separateable, so two players can play/pause different videos completely independendtly. For the most part this is already done, since I added the timeline context to the player now, so each player has it's own context.

- If possible, remove redundancy (we can do this at the end), like 

```
	<Internals.Timeline.TimelineContext.Provider value={timelineContextValue}>
			<Internals.Timeline.SetTimelineContext.Provider
				value={setTimelineContextValue}
			>
				<Internals.CompositionManager.Provider
					value={compositionManagerContext}
				>
```

is being used in multiple places

JonnyBurger:
Added a commit to jump back by 1 second by using Shift + Left/Right.

nandorojo:
Very excited for this PR!

Iamshankhadeep:
@JonnyBurger is it ready for review?

JonnyBurger (approved):

JonnyBurger (approved):

Issue #200, opened 2021-03-20
Fast frame-perfect seeking [Help needed]


JonnyBurger:
Problem: During render, we want to precisely capture each frame of a `<Video>`. Meaning frame 10 in the source video is frame 10 in the rendered video as well. This works, but the browser might not immediately show the correct frame after seeking to that position and emitting the `seeked` event. There is always some delay until the frame is actually displayed. This led to some frame jumping in Remotion and is currently fixed in the [Audio PR](https://github.com/yuta0801/remotion/blob/audio/packages/core/src/video/VideoForRendering.tsx) by just setting a 300ms timeout after the `seeked` event was emitted. This seems to work, but of course slows down render time.

This is my command with which I tested it: `npx remotion render src/index.tsx video-testing video-testing.mp4 && ffmpeg -i video-testing.mp4 -f image2 out%02d.jpg`

The screenshot below shows the desired output: Frames numbered 0 - n in the correct order.
The wrong output is if some frame numbers are duplicated. Sometimes one has to run a command multiple times to get a buggy output.

![Screenshot 2021-03-20 at 12 07 10](https://user-images.githubusercontent.com/1629785/111867519-cd639800-8974-11eb-8308-9750fbafe119.png)
 

What I've tried:

- `videoRef.current.requestVideoFrameCallback`: This is a new Web API (doesn't have TypeScript support yet) which should tell you when the frame is expected to be rendered. Unfortunately it wasn't accurate enough to be perfect. https://web.dev/requestvideoframecallback-rvfc/

- `Web Codec API`: Allows you to decode a video in Javascript. Needs a flag enabled but we can do that in Puppeteer. Unfortunately there is very little information on the internet out yet on how to do decode for example the example framer.mp4 video. I am struggling to pass the correct `description` parameter and don't know how to figure it out:

```js
  const canvas = document.getElementById("id");
    const video = document.getElementById("video");
    var mimeCodec = 'video/mp4; codecs="avc1.42E01E, mp4a.40.2"';

    var assetURL = "./packages/example/src/resources/framer-24fps.mp4";

    const mediaSource = new MediaSource();

    video.src = URL.createObjectURL(mediaSource);
    mediaSource.addEventListener("sourceopen", () => {
      const sourceBuffer = mediaSource.addSourceBuffer(mimeCodec);
      fetch(assetURL).then(async function (res) {
        sourceBuffer.addEventListener("updateend", function (_) {
[...]

PR #213, opened 2021-03-29
Test & improvements for frame-perfect seeking

cnnrhill:
This PR adds an automated test for the issue described in #200. 

* The `Framer` example is updated to include a deterministically random color in each frame. 
* The test renders the `VideoTesting` example - which contains `Framer` as an embedded video - to an image sequence and checks each frame's image for the correct color.
* If the `<Video>` component doesn't seek to and screenshot the correct frames from `Framer`, the colors will be wrong and the test will fail.
* Note: To make a .webm video from `Framer` for embedding, I found that I needed to render to .mp4 first and convert using ffmpeg (`ffmpeg -i framer.mp4 framer.webm`). Rendering directly to .webm with remotion seemed to shift the colors.

I've also updated the `<Video>` component to seek to the middle of each frame instead of the start. This cuts down on rounding errors that cause Chromium to display the wrong frame sometimes. 

@JonnyBurger, does this change to `<Video`> solve the issue with frame-perfect seeking when running on your system? In GitHub Actions, it reduces the number of missed frames from ~67 to zero. However, you mentioned that the issue is random/intermittent which makes me think it could be something platform-specific or dependent on timing? Let me know your thoughts...


cnnrhill:
Thanks for looking at this @JonnyBurger.

The failed test you linked is from a different branch without the change to `<Video>`. 

_Without `<Video>` change (fails with 67 missed frames):_

https://github.com/cnnrhill/remotion/runs/2216508169

_With `<Video>` change (this PR - passes with 0 missed frames):_

https://github.com/cnnrhill/remotion/actions/runs/696990827

I get similar results on my development machine. There may be some difference in how seeking works across platforms because I've had trouble replicating the exact results from your screenshot, even on MacOS - I do get 1 missed frame occasionally, but many are near the end of the video.

Since the issue is only happening on the initial seek on your computer, could it be that the video is emitting the `seeked` event before the video has actually finished buffering? If so, that would explain why the issue goes away on the next seek (the video finishes downloading between frames) and why it's sporadic (sometimes the video downloads in time, sometimes it doesn't).

The `seekable` property would tell you if this were the case:

https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/seekable

JonnyBurger:
Thanks for the great pointer @cnnrhill about the `seekable property`! I was very hopeful but no dice.

I am logging the `seekable` value, the current frame and the current time of the video, and all looks good: 

image

but still the output is faulty ☹️

image

I think also it only happens if you run it at a high concurrency, like 8x, I suspect that's why in the CI it always works. I wonder if you can also reproduce this effect.

Sorry for looking at the whole CI run, I am going to merge the test then! The other changes as mentioned I unfortunately cannot merge.

JonnyBurger (approved):

JonnyBurger:
@cnnrhill Upon further testing, I realized I did you wrong! 

Suddenly the test is failing on the audio branch badly. After applying your new duration logic, it was fixed!

But, it turns out there is a difference between MP4 and WebM videos! 🤯
So for WebM I am now using your logic, and for MP4 I am using my logic, and there are two tests, and both pass.

I am deleting the test from main branch but keeping it in the audio branch.
Also I have been able to verify that the bug only occurs when concurrency is high. So it's good the test passes always in the CI, and I'll debug this further.

Thanks a lot for bringing this to light!!

cnnrhill:
Okay - that makes sense! Too bad about the seekable property. I still haven't been able to replicate the exact issue but I'll keep thinking about it. 

Maybe Chromium's alternative deterministic rendering mode would help?

> Besides Chrome’s default rendering mode, Headless Chrome supports an alternative deterministic rendering mode (BeginFrameControl), in which BeginFrames are injected into the compositor by a DevTools client and all compositor stages are involved in each produced frame.

[Rendering Modes in Headless Chrome](https://docs.google.com/document/d/1PppegrpXhOzKKAuNlP6XOEnviXFGUiX2hop00Cxcv4o/edit?usp=sharing)

[HeadlessExperimental - Chrome DevTools Protocol](https://chromedevtools.github.io/devtools-protocol/tot/HeadlessExperimental/)

Anyway, thanks for merging - remotion is great!

JonnyBurger:
@cnnrhill thanks for bringing these to my attention! I never heard about these, I don't know how you find these :)

I gave it a try and enabled the following flags:

```
'--disable-threaded-animation',
'--disable-threaded-scrolling',
'--disable-checker-imaging',
'--disable-accelerated-video-decode',
```

And ran the test that you submitted, and removed the 100ms timeout. Unfortunately it went from passing to failing, so these flags don't fix the problem.

I think this is because Chrome delegates the video frames to a graphics thread and once it sends it Chrome doesn't have information over when it's rendered. That's why it's handled differently than other layout stuff.

PR #275, opened 2021-04-18
UseEffect splitting

calvin-f:
While writing tests we discovered, that in the following code the event listeners are only set in the second cycle of the useEffect. This relies on the fact, that actualVolume is in the useEffect dependency and will trigger a second cycle. 

```	
useEffect(() => {
  const ref = mediaRef.current;
  if (!ref) {
	  return;
  }
  if (ref.volume !== actualVolume) {
	  setActualVolume(ref.volume);
	  return;
  }
  const onChange = () => {
	  setActualVolume(ref.volume);
  };
  ref.addEventListener('volumechange', onChange);
  return () => ref.removeEventListener('volumechange', onChange);
}, [actualVolume, mediaRef]);
```

Therefore, I propose to split this into two useEffects. The mentioned tests will be submitted in a separate PR.


JonnyBurger (approved):

JonnyBurger:
Very nice 🙂 definitely better, should have followed best practices in the first place!
Thanks for the PR!

PR #278, opened 2021-04-19
update push.yml

### LINKED ISSUE #270: Run tests on lower FFMPEG versions
Would be nice if tests also run on FFMPEG version 4.1 (minimum supported) in addition to latest to give us some more safety when making changes.

Iamshankhadeep:
fixes #270 

Now github action uses ffmpeg version 4.1 and 4.4 given via action input.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -10,7 +10,7 @@ jobs:
     steps:
       - uses: actions/checkout@main
       - uses: actions/setup-node@main
-      - uses: FedericoCarboni/setup-ffmpeg@v1
+      - uses: Iamshankhadeep/setup-ffmpeg@v1-beta
         with:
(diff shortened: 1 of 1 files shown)


JonnyBurger (changes requested):
Thanks, nice fork!

I should have been more specific, the idea was that we test across multiple versions. So 4.4 and 4.1 would be nice.

I tried to just include both your fork and the original action in a matrix, but that [didn't work](https://github.com/JonnyBurger/remotion/actions/runs/767171152). 

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions

Can you adjust your fork so it builds both 4.1 and 4.4 and we can adjust it with a parameter maybe?

JonnyBurger (approved):
Wow impressive!
Really nice work 🙌 didn't expect it to be done so flawlessly!

PR #279, opened 2021-04-19
Test suite improvements (core)

j0nezz:
This PR was done together with @calvin-f and @dydent.

We focused on adding tests where test doubles are required. We also added `@testing-library/react-hooks` as dev dependency so that hooks can be directly tested. Further improvements include the removal of indirect tests and the increase of coverage for several units that were introduced in Remotion 2.0

As usual, we hope this PR helps and if you have some comments / change requests let us know :)


packages/core/src/config/image-format.ts
```diff
@@ -7,7 +7,7 @@ export type ImageFormat = typeof validOptions[number];
 let currentImageFormat: ImageFormat | undefined = undefined;
 
 export const setImageFormat = (format: ImageFormat) => {
-	if (typeof format === undefined) {
+	if (typeof format === 'undefined') {
```
JonnyBurger, line 10: Good catch! 🎣

packages/core/src/test/media-validation.test.tsx
```diff
@@ -1,101 +1,36 @@
...
-			),
-		/You have passed a volume of type string to your <Audio \/> component. Volume must be a number or a function/
+		// @ts-expect-error
+		() => validateMediaProps({volume: 'wrong'}, 'Audio'),
+		/You have passed a volume of type string to your <Audio \/> component./
 	);
 });
 
-test('It should not allow an video element to have a wrong type', () => {
+test('It should not allow a video element to have a wrong type', () => {
```
JonnyBurger, line 26: I think the previous tests were better because they were testing that e.g. the <Audio/> tag was actually calling the validateMediaProps function.
Shouldn't this in theory result in a higher code coverage as well?

Nevertheless, I am happy to leave it as is, definitely don't spend time refactoring it again!

packages/core/src/test/volume-prop.test.ts
```diff
@@ -0,0 +1,45 @@
...
+		expect(evaluateVolume(toEvaluate)).toEqual(smallVolume);
+	});
+	test('Volume is bigger than one', () => {
+		const toEvaluate = {frame: 10, volume: 10};
+		expect(evaluateVolume(toEvaluate)).toBe(1);
+	});
+	test('evaluated volume from frame that is smaller than one', () => {
+		const smallFrame = 0.5;
+		const toEvaluate = {frame: smallFrame, volume: (frame: number) => frame};
+		expect(evaluateVolume(toEvaluate)).toBe(smallFrame);
```
JonnyBurger: Testing a case that should never happen, `frame` will always be an integer.

I am changing it to frame * 0.5 which should be the same

packages/core/src/test/volume-prop.test.ts
```diff
@@ -0,0 +1,45 @@
...
+	test('It should throw if volume prop is neither number nor undefined', () => {
+		const toEvaluate = {frame: 10, volume: 'NaN'};
+		expectToThrow(() => {
+			// @ts-expect-error
+			evaluateVolume(toEvaluate);
+		}, /volume is not a function/);
+	});
+	test('It should throw if frame is invalid type', () => {
+		const invalidFrame = 'NaN';
+		const toEvaluate = {frame: invalidFrame, volume: (frame: any) => frame};
```
JonnyBurger: I refactored this portion a bit.
Two possible cases are imagineable, NaN and string, so I split it up.

Frame will never be 'NaN', but what can be tested is that the user returns NaN even though the frame is a number. I'm changes the test and removed the `any` also

JonnyBurger (commented):
Wow! Very impressed by these tests!
Great quality and so many of them 😊  (now 242 just in the core package)

And I learned some new techniques for testing, thanks for setting up this framework, this will definitely also be helpful for future tests!

I refactored some small things, and added some comments for them. But everything already done, merging! 🚂

JonnyBurger (approved):

Issue #266, opened 2021-04-17
Put `yarn create video` on homepage


JonnyBurger:
Just like many other projects, put the get started command right on the website and not hide it in the docs section

Tobijudah:
Hey Jonny. Wanting to hop on this! Any ideas on how you want it to be on the homepage? Or I should try some and you decide which is best?

JonnyBurger:
Hey @Tobijudah! Thanks a lot for this! 😊 
I am open to it, I think it can be pretty simple! Here are a few ones I personally like:

![image](https://user-images.githubusercontent.com/1629785/115158464-86111a00-a08e-11eb-9a2b-63b95bb6ed0a.png)


![image](https://user-images.githubusercontent.com/1629785/115158469-90cbaf00-a08e-11eb-8fff-fa16115ed6da.png)

![image](https://user-images.githubusercontent.com/1629785/115158527-ddaf8580-a08e-11eb-95dc-5d89c83b7f3d.png)

![image](https://user-images.githubusercontent.com/1629785/115158534-e86a1a80-a08e-11eb-8128-e141662a9f29.png)

Tobijudah:
okay, thanks. On it now

abhijit-hota:
@Tobijudah, I'm sure you'd have thought of something by now but this is a good place to put it according to me. Just suggesting.

![image](https://user-images.githubusercontent.com/8116174/115188832-230d9a80-a103-11eb-836b-869aa1cd312c.png)

Tobijudah:
yeah, I put it there @abhijit-hota. Thanks!

JonnyBurger:
Fixed by #281, thanks again!

PR #281, opened 2021-04-19
Add `yarn create video` on homepage

Tobijudah:
67 - localhost_3000_(Laptop with MDPI screen)

#266

### DIFF
--- a/packages/docs/src/pages/header.module.css
+++ b/packages/docs/src/pages/header.module.css
@@ -12,6 +12,19 @@
   text-align: left;
 }
 
+.code {
+  width: fit-content;
+  background-color: #303846;
+  color: #fff;
+  font-family: monospace;
+  padding: 0.75rem 1rem;
+  border-radius: 0.35rem;
+}
+
+.codeArrow {
+  padding-right: 0.5rem;
+}
+
 @media screen and (max-width: 1200px) {
   .row {
     flex-direction: column;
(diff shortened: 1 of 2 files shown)


Tobijudah:
@JonnyBurger what do you think?

Also, when I followed the instructions on *CONTRIBUTING.md* several package-lock.json files were updated, should I push those changes as well?

Iamshankhadeep:
Lgtm. Let @JonnyBurger review it.

Iamshankhadeep (approved):
Lgtm. Let's wait for @JonnyBurger review.

abhijit-hota:
I think we should use the `npx create-video` as the primary snippet and `yarn create video` as a secondary one.

JonnyBurger:
Thank you @Tobijudah! Looking good, and is also pretty much what I had in mind!
Thanks a lot for sending this contribution! 💙 

@abhijit-hota I think it's pretty equivalent, do you see any benefit? So far we've gone for `yarn create video` in all our resources, so for consistency I would prefer it. There is also `npm init video` which IMO reads cooler.

abhijit-hota:
>  do you see any benefit? 

Nothing as such. I personally prefer Yarn but most of the examples you provided in #266 show the `npm` script. If we're using yarn everywhere, then there's no added benefit.

>  There is also npm init video which IMO reads cooler.

Oh! Didn't know that.

JonnyBurger:
@abhijit-hota Got it! Makes sense.

A further iteration would be to have something like React Navigation where you can select your package manager: 

image

It may also default to NPM.

abhijit-hota:
Ah ,yes. Cool!

PR #288, opened 2021-04-20
Remotion three integration

olee:
Basic support layer for [@react-three/fiber](https://github.com/pmndrs/react-three-fiber) usage within remotion by providing 2 core features:
- RemotionThreeCanvas  
   Replacement for [@react-three/fiber](https://github.com/pmndrs/react-three-fiber) Canvas component which is compatible with remotion
- useVideoTexture hook which allows easily using VideoTexture in three by internally using the `<Video />` component

Other notable features:
- Provides a basic `<ThreeVideo />` component for easily rendering (fullsceen) videos
- Allows for powerful postprocessing using [@react-three/postprocessing](https://github.com/pmndrs/react-postprocessing) (internally using [postprocessing](https://github.com/vanruesc/postprocessing))

localhost_3000_three-basic


packages/example/.eslintrc
```diff
@@ -1,3 +1,3 @@
 {
-	"extends": "@remotion"
+	"extends": "@jonny"
```
JonnyBurger: This should be reverted, need to use the same config as the user will use. What problem did arise with it?

packages/example/src/ThreeShaderedVideo/ShaderedVideo.tsx
```diff
@@ -0,0 +1,99 @@
+import React from 'react';
```
JonnyBurger: Is this example supposed to work yet? I am only getting a black screen.

packages/example/src/ThreePostprocessing/SwirlEffect.tsx
```diff
@@ -0,0 +1,64 @@
+import {BlendFunction, Effect} from 'postprocessing';
+import React, {forwardRef, useMemo} from 'react';
+import {Uniform, Vector2} from 'three';
+
+const fragmentShader = `
+uniform vec2 offset;
+uniform float radius;
+uniform float angle;
+
+void mainUv(inout vec2 uv) {
```
JonnyBurger: I moved this into the example folder because most people don't need the swirl effect specifically. let's put all effects in example for now, otherwise the three package will get huge if we start shipping effects.

packages/three/src/RemotionThreeCanvas.tsx
```diff
@@ -0,0 +1,204 @@
...
+				left: 0,
+				overflow: 'hidden',
+				visibility: 'hidden',
+			}}
+		/>
+	);
+	return [threeDomNodesWrapper, remotionThreeContext] as const;
+}
+
+// Alternative implementation which does not directly call ReactDOM.render.
```
JonnyBurger: Still need to decide whether to use it - otherwise remove it

packages/example/src/ThreeShaderedVideo/ShaderedVideo.tsx
```diff
@@ -0,0 +1,113 @@
...
+			...rest
+		} = props;
+		const frame = useCurrentFrame();
+		const [size] = useThree((s) => [s.size] as const);
+		const videoTexture = useVideoTexture({src, startFrom, ...videoProps});
+		return (
+			<mesh
+				ref={ref}
+				scale={[size.width, size.height, 1]}
+				renderOrder={ignoreDepth ? -2 : undefined}
```
JonnyBurger: some of the values in here still seem very magical. What do they do and should we ship these values hardcoded?

packages/example/src/ThreeVideo/index.tsx
```diff
@@ -0,0 +1,20 @@
...
+import {Sequence} from 'remotion';
+
+const videoSrc =
+	'http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4';
+
+const Video = (): React.ReactElement => {
+	return (
+		<div style={{display: 'flex', width: '100%'}}>
+			<ThreeCanvas>
+				<Sequence layout="none" from={-460} durationInFrames={Infinity}>
```
JonnyBurger: The sequence does still throw some errors if you play around with it (parent not found). I will investigate this

packages/three/src/ThreeCanvas.tsx
```diff
@@ -0,0 +1,205 @@
...
+				left: 0,
+				overflow: 'hidden',
+				visibility: 'hidden',
+			}}
+		/>
+	);
+	return [threeDomNodesWrapper, remotionThreeContext] as const;
+}
+
+// Alternative implementation which does not directly call ReactDOM.render.
```
JonnyBurger: need to decide whether to use it or remove it

packages/three/src/useVideoTexture.tsx
```diff
@@ -0,0 +1,67 @@
...
+		});
+	}, [frame, render, videoTexture]);
+
+	useRemotionThreeDomNode(
+		<Video
+			{...videoProps}
+			ref={videoRef}
+			onLoadedData={onLoadedData}
+			crossOrigin="anonymous"
+			style={{ width: 800, height: 500 }}
```
JonnyBurger: hardcoded value

JonnyBurger:
Very very nice!

The canvas is very convienient and the video filters are unlocking whole new possibilities!

I did some cleanups and refactors. What is left to do is more cleanups (e.g. remove hardcoded values) and documentation of stuff (I started writing some docs).

I will also further contribute to this PR when I get time and investigate the sequence bug.

JonnyBurger:
@olee Should we as a first step just release the package with RemotionCanvas? This one seems pretty good for me to ship. We can then do video filters as a second step.

olee:
Sorry that I was a bit unavailable recently - had a lot of stuff going on at work 😅 
Anyway, I completely agree with you and think the current implementation should be ready to release as-is for people to try out and improve on.

JonnyBurger:
Thanks a lot for this PR! I have now merged the Canvas feature, and because it was messing with the types, removed the video filter for now. We can always retrieve the work by going back to the commit 98981eda94a19d045102aaecd307a0796c940b86.

Also I found when I wanted to use the <Stage> component from Drei that I got an exception because the composition is not wrapped in a Three <Canvas>. I did that and it started working without any hacks - so now it's much smaller and is just wrapping Remotion's contexts. I think that all the things that were possible before are also possible now, correct my if I am wrong.

In any case, thanks a lot for putting this much time into this awesome Remotion feature, I think that will be super popular going forward! And also the video filters I think we'll definitely continue to work on :)

packages/three/src/RemotionThreeCanvas.tsx
olee: If we do not squash the commits it would stay in the history and could be looked up again, so we might remove it

packages/example/src/ThreeShaderedVideo/ShaderedVideo.tsx
olee: It was working last time I checked... 🤔

packages/example/src/ThreeShaderedVideo/ShaderedVideo.tsx
olee: Well this is an example so...

olee:
Wait, you removed `useThreeVideo`?
At least this hook is definitely necessary and should be included in the code.
I had to do quite some adjustments to get the video handling working properly which this hook all takes care of.

Also the ThreeVideo component would be quite useful I think, however I have to agree that it was still more on a beta feature level.

As for the postprocessing stuff, I added this just to the examples for people to look into how to use this, however I can agree that it might not be definitely necessary,

PR #298, opened 2021-04-21
use yarn or npm given `yarn create` or `npm init`

### LINKED ISSUE #276: Using `yarn create video` installs the project using `npm`
When I initialize the project using the command `yarn create video` , I expect it to use Yarn as the package manager, using `yarn.lock` file. 

But it still installs all the projects' dependencies using npm and only the `create-video` binary is installed using Yarn.

Iamshankhadeep:
fixes #276 
now `create-video` understands that you are using `npm` or `yarn` as a package manager, default's to `npm` as a package manager.

### DIFF
--- a/packages/create-video/src/init.ts
+++ b/packages/create-video/src/init.ts
@@ -19,9 +19,17 @@ const askQuestion = (question: string) => {
 	});
 };
 
+const shouldUseYarn = (binaryPath: string): boolean => {
+	if (binaryPath.match(/.yarn/g)) {
+		return true;
+	}
+	return false;
+};
+
 xns(async () => {
 	const arg = process.argv[2];
 	let selectedDirname = arg?.match(/[a-zA-Z0-9-]+/g) ? arg : '';
+	const createVideoBinaryPath = process.argv[1];
 	while (selectedDirname === '') {
 		const answer =
 			(await askQuestion(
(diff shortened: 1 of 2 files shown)


JonnyBurger:
I was not able to get it working with your methodology. Maybe it's a problem of different yarn versions.

But I figured out in the yarn source code that they are adding some env variables when running in yarn and I am using this to detect. Can you verify on your end if this also works for you, before I merge?

Iamshankhadeep:
> I was not able to get it working with your methodology. Maybe it's a problem of different yarn versions.
> 
> But I figured out in the yarn source code that they are adding some env variables when running in yarn and I am using this to detect. Can you verify on your end if this also works for you, before I merge?

Can you just tell me how did you test my PRs?

Iamshankhadeep:
Yeah I did test it on my end and it's working. We are good to go, We can merge it.

JonnyBurger (approved):
Awesome!

JonnyBurger:
@Iamshankhadeep i just published a package myself based off your source code as well. Also don't know of a better way

Iamshankhadeep:
I see but are you using yarn 2 or yarn 1?

JonnyBurger:
Yarn 1

Iamshankhadeep:
The code that I suggested was working with yarn 1.22 I am not sure why did it work on your end.

packages/create-video/src/init.ts
```diff
@@ -76,10 +93,10 @@ xns(async () => {
 
 	console.log('Get started by running');
 	console.log(chalk.blue(`cd ${selectedDirname}`));
-	console.log(chalk.blue('npm start'));
+	console.log(chalk.blue(shouldUseYarn() ? 'yarn start' : 'npm start'));
 	console.log('');
 	console.log('To render an MP4 video, run');
-	console.log(chalk.blue('npm run build'));
+	console.log(chalk.blue(shouldUseYarn() ? 'yarn build' : 'npm run build'));
```
abhijit-hota, line 99: Do we really need to call the function thrice? (L71, L96, L99)
I we can refactor to something like 
```js
const shouldUseYarn = checkYarnUsage();
```

packages/create-video/src/init.ts
```diff
@@ -61,13 +68,23 @@ xns(async () => {
...
+		promise.stdout?.pipe(process.stdout);
+		await promise;
+	} else {
+		console.log('> npm install');
+		const promise = execa('npm', ['install'], {
+			cwd: outputDir,
+		});
+		promise.stderr?.pipe(process.stderr);
+		promise.stdout?.pipe(process.stdout);
+		await promise;
```
abhijit-hota, line 86: A lot of functionality is common. I propose the following refactor:
```js
const logInitOutput = shouldUseYarn ? '> yarn' : '> npm install';
const args = shouldUseYarn ? ['yarn', []] : ['npm', ['install']];

console.log(logInitOutput);
const promise = execa(...args, { cwd: outputDir });

promise.stderr?.pipe(process.stderr);
promise.stdout?.pipe(process.stdout);
await promise;
```

Variable names can be changed.

PR #307, opened 2021-04-23
Added support for playbackrate to video

dested:
This PR adds playbackRate to the Video component. It does this by simply adding playbackRate to the getCurrentTime function and multiplying the time by that. I also set the playback rate of the video element itself so the preview draws better (for the most part). 

While this method does work it has a drawback: you cannot really have a variable playback rate. 

For instance, let's say you wanted the video to speed up and slow down, maybe setting the playback rate to `1+Math.sin(frame)`. This would cause the multiplication to calculate the wrong frame since the current frame of the video does not take into account the playbackRate at all previous frames, only the playbackRate at the current frame (I hope this makes sense :-/). 

I don't think that should necessarily block this PR as it does add a lot of value, but it's something to consider as a future feature. The only way I can see supporting it is something like `playbackRate={(frame)=>1+Math.sin(frame)}`, and effectively summing the result for every frame up until the current frame. This means for very long videos you're going to be doing a lot of math. I'd love to hear recommendations on how to solve this in a less naive way though!

This PR does not support playback rate for audio as I was less certain about what exactly to modify, so I simply jam the playbackRate to be 1 for audio. It also does not have any tests as I do not know exactly how the test videos are being generated. Any support there would be helpful!

This fixes https://github.com/JonnyBurger/remotion/issues/232


JonnyBurger:
Hey @dested! Thanks a lot for jumping on this!

I totally agree with the approach here, let's start simple and only support a constant playback rate. and video. We can expand to a `playbackRate: (frame: number) => number` later without breaking backwards compatibility.

The most obvious issue to me seems that if a video has audio, that audio does not get speed up.
We need to introduce an [FFMPEG `atempo` filter](https://ffmpeg.org/ffmpeg-filters.html#atempo) to handle it. Interestingly FFMPEG only supports tempo in the 0.5-100 range, so we might have to daisy chain it (0.5 * 0.5) if you want for example 0.25.

Furthermore we need to mention it in the docs and add some basic validation (what if passing NaN, string (for plain JS users), negative values, Infinity to playbackRate, what if the supposedly constant rate is unexpectedly changing).

I am happy to also contribute to this when I get time and maybe @Iamshankhadeep can help too.

To answer your question, the tests can be run with the `npx jest` command from the root of the Remotion monorepo.

Thanks again for this PR and I'm confident we'll get it merged soon! 🙌

OlehShevtchuk:
I have a similar issue and hope that support for playbackrate control both to video and video with audio will be implemented in the near future🙂

dested:
Thanks @JonnyBurger. I am going to try my hand at atempo but it's a bit out of my depth. I will make updates to the documentation and tests soon!

JonnyBurger:
@dested Sounds good, and if you need help, ping me!

If you can't manage it, no problem, leave the PR and let me know, I'll try to get it merged in due time 😎

JonnyBurger:
@dested I am working on finishing this today, unless you say that you are coincidentially also working on it at the same time.

JonnyBurger:
Moved to #370 so CI will run

PR #310, opened 2021-04-24
Adding yarn.lock and package-lock.json

### LINKED ISSUE #309: dev server error: unexpected token `static in(easing) {`
## Error

image

## Reproduction Steps

1. created new project via `yarn create video`
2. `cd <directory_name>`
3. `yarn start`
4. see error in both console and browser

Iamshankhadeep:
Adding yarn.lock and package-lock.json.
fixes #309 issues similar like this.


JonnyBurger:
Nice! I am hesitant to merge this because now it will not by default install the newest Remotion version. Also the lockfiles seem to be different across npm versions which could potentially be a headache.

I should have described it better, but what I meant by pinning the versions was to remove the carets `^` in package.json. But this may also not be ideal as commands like `yarn upgrade` will not work for the user.

I'm not merging it right away but want to revisit this. Inputs appreciated...

Iamshankhadeep:
> Nice! I am hesitant to merge this because now it will not by default install the newest Remotion version. Also the lockfiles seem to be different across npm versions which could potentially be a headache.

I am also thinking the same it'll be better to not use the `lock` files. Do you think I should close this PR?

JonnyBurger:
Yes :( sorry for the effort which went to waste. But I also think it's better to not have them.

Another reason is that I want to make these repos template repos, aka you can just clone them from Github and install the dependencies yourself and get started, or click the fork button.

Two lockfiles would be a conflict.

PR #312, opened 2021-04-25
Update output

### LINKED ISSUE #237: after downloading asset, output is not nice
image

samuel3105:
Capture

make new output after downloading .

Also add some forgot types.


JonnyBurger (approved):

JonnyBurger:
Thanks for the PR! 💙 🙌 

I just made two changes: The Logging functions are meant for log levels (info, warn, error), not for colors. So I inlined the colors and replaced it with Log.Info calls.

Also I had to unlink the issue #237 - the issue I meant was that if a remote asset is downloaded, that log mixes with the other output. So while the output is now nicer, the original problem is not fixed by it. Nonetheless very useful!

PR #314, opened 2021-04-26
Fix input props issue

JonnyBurger:


JonnyBurger:
This was the wrong way to fix it, better PR incoming

PR #330, opened 2021-05-01
Interpolation improvements

j0nezz:
I did some mutation testing using Stryker and i stumbled over quite a few branches in the interpolation file which are never reached (and therefore can be arbitrarily mutated with all tests still passing).

- `inputMin === inputMax` doesn't need to be tested, because we require input ranges to be **strictly** monotonically increasing
- Removed all checks for Inifinity in input/output ranges as we only allow finite numbers.
- removed array length check in `checkValidInputRange` as it is already done in `checkInfiniteRange `
- removed `if (!(arr.length !== 2 || arr[0] !== -Infinity || arr[1] !== Infinity))` which is logically equivalent to `if(arr.length === 2 && arr[0] === -Infinity & arr[1] === Infinity)` => This very specific case is already covered by the following loop and I don't see the need for an extra check before.

I also improved the tests a tiny bit and everything is still passing. Hope it helps :)

### DIFF
--- a/packages/core/src/interpolate.ts
+++ b/packages/core/src/interpolate.ts
@@ -46,34 +46,14 @@ function interpolateFunction(
 		return outputMin;
 	}
 
-	if (inputMin === inputMax) {
-		if (input <= inputMin) {
-			return outputMin;
-		}
-
-		return outputMax;
-	}
-
 	// Input Range
-	if (inputMin === -Infinity) {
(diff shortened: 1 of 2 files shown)


JonnyBurger (approved):

JonnyBurger:
Amazing! Thanks a lot! 👏 👏

PR #332, opened 2021-05-03
Mutation testing to improve test suites (core)  /  add coverage badges

dydent:
This PR was done together with @calvin-f and @j0nezz .

In this batch of improvements we did some mutation testing using the Stryker Mutator. We found several mutants, showed no mercy and killed them ;) We added some hints regarding mutation testing to the `CONTRIBUTING.md` file, so that in the future other contributors can also test their quality of tests.
Since the coverage has grown quite a few percent during the last weeks, we decided to add some coverage badges to the core `Readme.md` file. The badges are updated after each test run.

As usual, we hope this helps and makes remotion more robust. If you have any change requests, let us know!


packages/core/package.json
```diff
@@ -5,7 +5,7 @@
 	"main": "dist/index.js",
 	"sideEffects": false,
 	"scripts": {
-		"test": "eslint src --ext ts,tsx && jest",
+		"test": "eslint src --ext ts,tsx && jest && istanbul-badges-readme --silent",
```
JonnyBurger: This is very cool!

I get an error `Error: Coverage file does not exist in ./coverage/coverage-summary.json`, I guess because no test coverage is being generated.

Also: The README of the root package currently gets copied to the core package when publishing. This is so both on the Remotion Github Repo and on the NPM package you can see the README.

So I am unsure if the test coverage of the core package should be at the top of the README. I would prefer to keep it clean and selectively only put the most important information in the README. However happy to put it somewhere else if this is possible.

JonnyBurger:
Awesome! Thanks a lot! 🙌 

Besides the comment about the badges, tests look good 😊  Thanks for another great contribution! 💙

packages/core/package.json
j0nezz: As discussed, we moved everything to `TESTING.md`. @calvin-f has written a script so we have zero dependencies for the badge generation👌🏻  Makes a lot more sense now!

JonnyBurger:
Looks very stable now! 🙌  thanks a lot! 💙

JonnyBurger (approved):

PR #358, opened 2021-05-09
Add note on Windows and inline JSON content

Maxou44:
I added a note about Windows and inlined JSON


JonnyBurger (approved):

JonnyBurger:
Thanks a lot!
I took the opportunity to also add this as a warning in the CLI if you stumble upon the error and to improve the docs 🙌  Hope from here on it will be smooth sailing!

PR #373, opened 2021-05-16
Feature/use hash as assets filename

sheunglaili:
use `sha-1` and first 10 characters as assets filename to prevent super long filename when using speech synthesis api with endpoint like `/speech/:text` and audio in data url format `data:audio/mpeg:base64,blahblahblah..........`


JonnyBurger:
Thanks a lot! appreciate reporting and fixing yourself! 💙 🙌 Really awesome.

I did tweak it a bit because I found SHA256 resolves differently based on the OS. So I used our own hashing function instead. Also added more tests.

I'm going to close this one but I copied it over to #374, so it will run the tests in our CI.

PR #374, opened 2021-05-16
use sha256 as assets filename

JonnyBurger:


packages/renderer/src/assets/download-and-map-assets-to-file.ts
```diff
@@ -63,8 +62,18 @@ export const downloadAndMapAssetsToFileUrl = async ({
...
 }): Promise<TAsset> => {
-	const {pathname} = new URL(localhostAsset.src);
-	const newSrc = path.join(webpackBundle, sanitizeFilename(pathname));
+	const {pathname, search} = new URL(localhostAsset.src);
+
+	const split = pathname.split('.');
+	const fileExtension =
+		split.length > 1 && split[split.length - 1]
+			? `.${split[split.length - 1]}`
+			: '';
```
sheunglaili, line 86: when using `/speech?text=blahblahblah` like tts API , this will have no file extension.

sheunglaili (commented):
Thanks for the quick update! I looked into your code a bit and it was nice 👍  yet I have little concern over the file extension part.

packages/renderer/src/assets/download-and-map-assets-to-file.ts
sheunglaili: can we possibly identify the remote assets file extension from the response `Content-type` header ?

JonnyBurger:
@sheunglaili all i do is add back the original file extension, could it be such a bad thing?

JonnyBurger:
Yes if the URL has no file extension then the file will also have none, but this was also the case before, so it should not introduce any new problems.

sheunglaili:
> Yes if the URL has no file extension then the file will also have none, but this was also the case before, so it should not introduce any new problems.

This work absolutely with existing flow. It's just give rise to problem when we are getting our audio assets from tts api. These tts api often has endpoint like : `/speech?text=blahblahblah` or `/speech/:text` which doesn't include a file extension in their endpoint. I reckon we could do a `HEAD` request and the mime type from `Content-type` response header to precisely reflect the extension to improve experience when working with tts api.

Iamshankhadeep:
> This work absolutely with existing flow. It's just give rise to problem when we are getting our audio assets from tts api. These tts api often has endpoint like : `/speech?text=blahblahblah` or `/speech/:text` which doesn't include a file extension in their endpoint. I reckon we could do a `HEAD` request and the mime type from `Content-type` response header to precisely reflect the extension to improve experience when working with tts api.

I see what you are trying to say but I did go through this [stackoverflow](https://stackoverflow.com/a/48704300/8737200) answer for `Content-Type` allowed value but I did not found any way to guess the actual file extension for the media. If you can provide me with some links I will be helpful.

sheunglaili:
> > This work absolutely with existing flow. It's just give rise to problem when we are getting our audio assets from tts api. These tts api often has endpoint like : `/speech?text=blahblahblah` or `/speech/:text` which doesn't include a file extension in their endpoint. I reckon we could do a `HEAD` request and the mime type from `Content-type` response header to precisely reflect the extension to improve experience when working with tts api.
> 
> I see what you are trying to say but I did go through this [stackoverflow](https://stackoverflow.com/a/48704300/8737200) answer for `Content-Type` allowed value but I did not found any way to guess the actual file extension for the media. If you can provide me with some links I will be helpful.

This package https://www.npmjs.com/package/mime-types provides a method `extension` to get the default extension from mime-type which is quite useful.

JonnyBurger:
Ideally I want to keep it as simple as possible. So far no problems have arisen from missing file extensions.

`ffmpeg` from what I understand is powerful enough to interpret files in the right way even if file extensions are missing. We even use the `ffprobe` binary in Remotion and from my experience in using it, it recognizes all possible scenarios and is based on reading the first few bytes of the file to detect the format.

I will only install a new dependency and import a MIME database if it shows that the current way is leading to problems. Until then, I think we are fine with this PR. Thanks a lot again @sheunglaili! 🙌

sheunglaili:
No worries! Thanks for taking my opinion into account 🙌🏻

JonnyBurger:
@sheunglaili Thanks for understanding! I also added a fix for #375 in the same PR 😊

Iamshankhadeep (approved):
lgtm.

PR #376, opened 2021-05-17
Minor test suite improvements

dydent:
This PR was done together with @calvin-f and @j0nezz with the goal of improving the test code quality of the core package.

A few notes on two things we changed:

- Fixed a bug in the function `isValidLogLevel` inside `log.ts`. Currently, the not listed log levels would be recognized as valid, even tho they should be recognized as invalid.
- The test file `get-media-time.ts` was missing the `.test` in the file name, so the tests were not running. We renamed it to `get-current-time.test.ts` and improved the tests so they would pass.

So we hope this PR helps, if you have some comments / change requests let us know :)


JonnyBurger:
Thanks a lot again! 🙌 

Indeed, there was a bug, thanks a lot for spotting and fixing it, alongside the various tests!
Also stupid from me to add tests that were not running 🤦  thanks for fixing that as well.

Issue #35, opened 2021-02-09
Include Transition components as Helpers


BrunnerLivio:
I‘ve seen you already created some awesome transitions in your example repositories, such as [Transition.tsx](https://github.com/JonnyBurger/remotion-trailer/blob/main/src/Transition.tsx) or [FadeTransition.tsx](https://github.com/JonnyBurger/remotion-trailer/blob/main/src/FadeTransition.tsx).
I think it could be a nice addition to have since it is something a lot of videos will need to use!

JonnyBurger:
True!

I'm thinking since they are opinionated, they shouldn't be released in core. It would be nice to publish them in some way. Maybe @remotion/transitions or maybe I release them in a separate repo personally.

I'm leaving this open until I'll do it which I plan to 👌

nitinthewiz:
From what I understand, transitions are used only in Sequences and not with Compositions, right? They certainly make sense to be a part of a separate package for those who need them... Just my two cents! 😃

JonnyBurger:
This is now WIP: https://github.com/remotion-dev/remotion/pull/1057

PR #386, opened 2021-05-20
Add Series component

### LINKED ISSUE #336: <Stagger> component
It will render multiple components in sequence, without having to provide both from and durationInFrames prop to Sequence

Iamshankhadeep:
Fixes #336

### DIFF
--- a/packages/core/src/series/index.tsx
+++ b/packages/core/src/series/index.tsx
@@ -48,7 +48,7 @@ const Series: FC & {
 
 			if (!castedChild || !castedChild.props.children) {
 				throw new TypeError(
-					`A <Series.Sequence /> component (${debugInfo}) was detected to not  have any children. Delete it to fix this error.`
+					`A <Series.Sequence /> component (${debugInfo}) was detected to not have any children. Delete it to fix this error.`
 				);
 			}
 
--- a/packages/core/src/test/series.test.tsx
+++ b/packages/core/src/test/series.test.tsx
@@ -168,7 +168,7 @@ test('Handle empty Series.Sequence', () => {
 					<Series.Sequence durationInFrames={10} />
 				</Series>
 			),
-		/A <Series.Sequence \/> component \(index = 1, duration = 10\) doesn't have any children\./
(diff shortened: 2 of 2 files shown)


JonnyBurger:
Thanks a lot!

It seems to work great, but I'd like to think more about the API design before merging it.

Specifically:

- Should we colocate the duration and the content? Right now you have to pass the component as children but the duration are passed at the top as an array. Should there be another component so you can colocate the duration and the content? Like

```
<Stagger.Container>
  <Stagger.Piece></Stagger.Piece>
  <Stagger.Piece></Stagger.Piece>
  <Stagger.Piece></Stagger.Piece>
</Stagger.Container>
```

- Naming: Is "Stagger" a good name? What about "Series", "Relay"? "Sequence" is already taken unfortunately...

- Handling edge cases: What if people don't put the same number of arguments as durations? Need to handle NaN, negative, Infinity, non-number values as well...

- Preventing user mistakes: What if someone puts a sequence as a children? They might get confused by the result. Should we forbid putting a sequence or use the syntax I posted above to make it more explicit?

- Transitions: I think a big usecase is that you can combine it with transitions. e.g. 20 images as a slideshow but with a fade transition in between. Currently it doesn't seem possible. Now that `measureSpring()` is out we can also start working on a `@remotion/transitions` package (#35) that ideally plays well with this component.

Once we are certain that our component is versatile and robust that it satisfies all these constraints, I'll merge it. Happy to hear feedback from you and other Remotion users!

Iamshankhadeep:
> ```
> <Stagger.Container>
>   <Stagger.Piece></Stagger.Piece>
>   <Stagger.Piece></Stagger.Piece>
>   <Stagger.Piece></Stagger.Piece>
> </Stagger.Container>
> ```
I like this api design. It's way more clean. User will pass duration prop on `<Stagger.Piece/>` component right?

> * Naming: Is "Stagger" a good name? What about "Series", "Relay"? "Sequence" is already taken unfortunately...

I am in support for `Series` it sounds cool :slightly_smiling_face:

PR #389, opened 2021-05-21
Add support for react-native-svg

wcandillon:


wcandillon:
Just tried it locally and it works great 🙌🏻

JonnyBurger:
@wcandillon I think having many of these PRs is not very scalable (not going to the moon), so the better way to go I think is if you put this in your `remotion.config.ts`:

```ts
Config.Bundling.overrideWebpackConfig((config) => {
	return {
		...config,
		resolve: {
			...config.resolve,
			alias: {
				...config.resolve.alias,
				'react-native-svg$': 'react-native-svg-web',
			},
		},
	};
});
```

That way not too much opinionated stuff get added to the config.

wcandillon:
nice thks a lot!

PR #391, opened 2021-05-24
Removed lingering BlueBall example

patcon:
I see why you updated away this example 😄


Iamshankhadeep (approved):
Thanks for noticing this typo and a PR to fix this merging.

JonnyBurger:
😂 😂 Yes indeed that was the initial name and yes at the last moment I changed it for obvious reasons! 
Thanks a lot for finding this leftover and fixing this 🙌

PR #394, opened 2021-05-24
Player doubleclick support

### LINKED ISSUE #347: Double click to fullscreen in `<Player />`

Iamshankhadeep:
fixes #347


JonnyBurger:
Thanks! Looks awesome, very well engineered the 200ms delay!
Could not find any bugs. Great quality 👏 

Only thing I changed is to disable the behavior by default. Since `controls` is also disabled by default, I would find it surprising that by default there are no controls but you can double click.

JonnyBurger (approved):

Iamshankhadeep:
I see that sounds awesome :fire:

PR #398, opened 2021-05-25
fix rendering on very long videos

### LINKED ISSUE #395: Maximum callstack size exceeded
Hi all. I tried to render a very long video today (8 hours), and when it began encoding, there was this following error:
```
RangeError: Maximum call stack size exceeded
    at Object.stitchFramesToVideo (...\node_modules\@remotion\renderer\dist\stitcher.js:66:32)
    at async Object.render (...\node_modules\@remotion\cli\dist\render.js:166:9)
    at async cli (...\node_modules\@remotion\cli\dist\index.js:34:9)
```
I tried looking through the stack trace to see where the call stack could have been maxed out, but currently don't understand. Was wondering if anyone else had run into this issue? Would it be simply because the video is too long? Thank you!

JonnyBurger:
Rendering on very long videos (e.g. 8 hours) is broken because `Math.max(...new Array(300000).fill(1))` throws a `Maximum call stack exceeded` error. Fixing this by using our own `min()` and `max()` functions 🔥 

Fixes #395, should ping dLamSlo8#0898 on Discord to let them know it's fixed afterwards

### DIFF
--- a/packages/renderer/src/min-max.ts
+++ b/packages/renderer/src/min-max.ts
@@ -1,0 +1,34 @@
+// Standard library Math.min and Math.max can throw
+// if array length is very long. Fixing this with own implementation
+
+export const min = (arr: number[]) => {
+	if (arr.length === 0) {
+		throw new Error('Array of 0 length');
+	}
+
+	let smallest = arr[0];
+	for (let i = 0; i < arr.length; i++) {
+		const elem = arr[i];
+		if (elem < smallest) {
+			smallest = elem;
+		}
+	}
+
+	return smallest;
+};
+
+export const max = (arr: number[]) => {
+	if (arr.length === 0) {
+		throw new Error('Array of 0 length');
+	}
+
+	let biggest = arr[0];
+	for (let i = 0; i < arr.length; i++) {
+		const elem = arr[i];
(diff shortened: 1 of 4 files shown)


Iamshankhadeep (approved):
lgtm

packages/renderer/src/min-max.ts
```diff
@@ -0,0 +1,34 @@
+// Standard library Math.min and Math.max can throw
+// if array length is very long. Fixing this with own implementation
+
+export const min = (arr: number[]) => {
+	if (arr.length === 0) {
+		throw new Error('Array of 0 length');
```
Iamshankhadeep, line 6: when `arr.length` is 0, it throws error and it's causing CI to fail. I guess best way to handle this condition is to return `0` here cause `Math.min([]) = 0`.

JonnyBurger: This would actually lead to some unintended evaluation - if you would render the audio, it would lead to `numberLength === "Infinity"`. This was a bug but it had no negative consequences. But thanks to the test we noticed it and I managed to improve upon it now 😇

Iamshankhadeep (approved):
awesome thanks

PR #400, opened 2021-05-27
create-video update and templates

### LINKED ISSUE #392: Plain JS starter template
A plain JS starter template for people who don't like using Typescript. In my view it should just be a separate repository marked as template so Github shows the tag "Use this template".

Iamshankhadeep:
fixes #392 

- [x] add all templates
- [x] add more logs
- [x] error and validations


JonnyBurger:
@Iamshankhadeep 

We can also add these templates now:

Three.JS: https://github.com/jonnyburger/remotion-template-three
Text to Speech: https://github.com/FelippeChemello/Remotion-TTS-Example

packages/create-video/src/init-v2.ts
```diff
@@ -0,0 +1,266 @@
...
+	}
+
+	// not in git tree, so let's init
+	try {
+		await execa('git', ['init'], {cwd: root});
+		!flags.silent && Log.info('Initialized a git repository.');
+
+		if (flags.commit) {
+			await execa('git', ['add', '--all'], {cwd: root, stdio: 'ignore'});
+			await execa('git', ['commit', '-m', 'Create a new Remotion app'], {
```
JonnyBurger: Let's not use "app" as the term for a project but "video"

packages/create-video/src/init-v2.ts
```diff
@@ -0,0 +1,266 @@
...
+	}
+};
+
+const resolveProjectRootAsync = async () => {
+	let projectName = '';
+	try {
+		const {answer} = await prompts({
+			type: 'text',
+			name: 'answer',
+			message: 'What would you like to name your app?',
```
JonnyBurger: App should be renamed to video throughout

packages/create-video/src/init-v2.ts
```diff
@@ -0,0 +1,266 @@
...
+					path.basename(path.resolve(name))
+				);
+				if (typeof validation === 'string') {
+					return 'Invalid project name: ' + validation;
+				}
+				return true;
+			},
+		});
+
+		if (typeof answer === 'string') {
```
JonnyBurger: If I don't type anything in, it will create a folder but just return and not create a project.

packages/create-video/package.json
```diff
@@ -19,11 +19,17 @@
...
+		"fs-extra": "^9.1.0",
+		"npm-package-arg": "^8.1.2",
+		"prompts": "^2.4.1",
 		"xns": "^2.0.7"
 	},
 	"devDependencies": {
 		"@jonny/eslint-config": "^2.1.233",
+		"@types/fs-extra": "^9.0.11",
 		"@types/node": "^15.0.1",
+		"@types/npm-package-arg": "^6.1.0",
```
JonnyBurger: npm-package-arg and @types/npm-package-arg seems unused now

packages/create-video/src/create-directory.ts
```diff
@@ -0,0 +1,79 @@
...
+	if (typeof name !== 'string' || name === '') {
+		return 'The project name can not be empty.';
+	}
+	if (!/^[a-z0-9@.\-_]+$/i.test(name)) {
+		return 'The project name can only contain URL-friendly characters (alphanumeric and @ . -  _)';
+	}
+	return true;
+}
+
+export function getConflictsForDirectory(
```
JonnyBurger: So if you enter a folder name that already exists it will trigger this function? Should we just simplify it and exit with an error if that directory already exists instead?

packages/create-video/src/init-v2.ts
```diff
@@ -0,0 +1,267 @@
...
+
+	await fs.ensureDir(projectRoot);
+
+	await assertFolderEmptyAsync(projectRoot, folderName);
+
+	return [projectRoot, folderName];
+};
+
+const isNodeVersionGreater = () => {
+	return process.versions.node >= '16.0.0';
```
JonnyBurger: Not necessarily the most stable way to check it... e.g `2.0.0` > `16.0.0` ==> true

Alternative way which doesn't rely on version checking:  `(fs.rmSync ?? fs.rmdirSync)()`

packages/create-video/src/init-v2.ts
```diff
@@ -0,0 +1,267 @@
...
+};
+
+export const init = async () => {
+	// let projectName = process.argv[2];
```
JonnyBurger: Let's remove it before release

JonnyBurger:
Besides the comments, I am pretty happy how this turned out :) Thanks a lot!

packages/create-video/src/init-v2.ts
Iamshankhadeep: got it :fire:

packages/create-video/src/init-v2.ts
Iamshankhadeep: okay okay :fire:

packages/create-video/package.json
Iamshankhadeep: yeah it is I will delete it. thanks for noticing. :fire:

packages/create-video/src/init-v2.ts
Iamshankhadeep: yup I will do that.

packages/create-video/src/init-v2.ts
Iamshankhadeep: okay that sounds awesome. :fire:

packages/create-video/src/create-directory.ts
Iamshankhadeep: yeah I think simplification is the way to go.

packages/create-video/package.json
JonnyBurger: Still here, can you remove it?

package.json
```diff
@@ -29,6 +29,7 @@
     "@react-three/fiber": "^6.1.2",
     "react": "^17.0.1",
     "react-dom": "^17.0.1",
+    "strip-ansi": "^3.0.1",
```
JonnyBurger: Why did it get installed in the root? When published on NPM the module will not be available, I think it will break...

packages/create-video/src/create-directory.ts
JonnyBurger: It should mean this file can be removed now right?

JonnyBurger (changes requested):
Thanks! I think we are almost there 🙂 just some things left to address..

JonnyBurger:
@Iamshankhadeep I am now happy with it! 🎉 

Before merging, I would like to test it on Windows quickly which I can do on Monday. Once that is also done, I will merge it and release it. No action required on your side. 99% of the way there, thank you so much 🙂

Iamshankhadeep:
awesome thanks a lot :fire:

JonnyBurger:
Just tested on Windows, works great! 🥳🥳🥳🥳🥳🥳
Thanks a lot, excited to ship this!!

JonnyBurger (approved):

Iamshankhadeep:
awesome exited for this too :fire:

PR #402, opened 2021-05-28
Warn about non-seekable media

JonnyBurger:
I've observed that when serving from a private Google Cloud Storage bucket, you cannot seek an MP3. We cannot do anything about it except provide some helpful info.


Iamshankhadeep (approved):
awesome.

PR #410, opened 2021-05-31
Apple ProRes support

wcandillon:

### DIFF
--- a/packages/cli/src/get-cli-options.ts
+++ b/packages/cli/src/get-cli-options.ts
@@ -120,6 +120,14 @@ const getAndValidatePixelFormat = (codec: Codec) => {
 	return pixelFormat;
 };
 
+const getAndValidateProResProfile = (actualCodec: Codec) => {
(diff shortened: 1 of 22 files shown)


JonnyBurger (approved):
Thanks a lot! Awesome idea 💙  🙌

wcandillon:
Amazing, it looks like the pixel format needs to be yuva444p10le, I'm getting: `Incompatible pixel format 'yuva420p' for codec 'prores_ks', auto-selecting format 'yuva444p10le'`

JonnyBurger:
@wcandillon We don't set `yuva420p` automatically, did you by any chance set it in your config file explicitly? If so, removing it may fix it.

wcandillon:
indeed, I just thought maybe that yuva444p10le should be an allowed value maybe? 🤷‍♂️

JonnyBurger:
@wcandillon Good point, made another issue about it!

PR #417, opened 2021-06-02
💚 Fix ProRes transparency support

### LINKED ISSUE #413: Support yuva444p10le as valid pixel format
> indeed, I just thought maybe that yuva444p10le should be an allowed value maybe? 🤷‍♂️

_Originally posted by @wcandillon in https://github.com/JonnyBurger/remotion/issues/410#issuecomment-852138819_

wcandillon:
fixes #413


packages/core/src/config/pixel-format.ts
```diff
@@ -8,6 +8,7 @@ const validOptions = [
 	'yuv420p10le',
 	'yuv422p10le',
 	'yuv444p10le',
+	'yuva444p10le'
```
wcandillon, line 11: This is not part of the fix, just added it.

JonnyBurger (commented):
Thanks a lot! 💙 🙌

JonnyBurger (approved):

PR #430, opened 2021-06-06
pass inputProps to component in preview

### LINKED ISSUE #428: —props doesn’t work in Preview Mode
As reported by @saftipierreminiggio on Discord

JonnyBurger:
Previously you could only get it using `getInputProps` but that is not consistent with how it works during rendering. Fixes #428


Iamshankhadeep (approved):
lgtm

PR #432, opened 2021-06-06
set environment variables earlier

### LINKED ISSUE #422: Environment variables get set too late
Although you can read them from a component, it cannot be read from the top-level.

JonnyBurger:
Fixes #422 

At first I thought that this technique would not work if you cached the webpack bundle, but I verified that it works even with cache enabled


Iamshankhadeep (approved):
lgtm

PR #434, opened 2021-06-08
avoid out of memory error

JonnyBurger:
calling Promise.all with over 800'000 array elements (if rendering an 8 hour video) might be the cause of the problem that leads to an out of memory problem. Doing it in chunks, that should help!


Iamshankhadeep (approved):
lgtm

PR #435, opened 2021-06-08
fix missing array dependencies

JonnyBurger:
Fix missing array dependencies. ESLint rule would warn about it


Iamshankhadeep (approved):
thanks

PR #443, opened 2021-06-12
Update spring.md

UmungoBungo:
Small fix - the config keys got spring() are all lowercase. When copying them into my code, I was getting an error that the property did not exist in the type.


JonnyBurger (commented):
Thanks! Good catch 🙌

JonnyBurger (approved):

PR #448, opened 2021-06-13
Fix link to three starter template

marcelmusiol:


JonnyBurger:
Thanks a lot for pointing this out and sending a PR! 🙌 

I just fixed it the other way around - I moved the repo to remotion-dev/template-three. So closing this but you still helped, thanks a lot! 🙂

PR #459, opened 2021-06-15
fix #457

### LINKED ISSUE #457: Error reporting in getVideoMetadata
I tried get the length of an mp4 video using the ```getVideoMetadata``` function and the render process threw an error. It took me a while to realize my mistake.

The problem was that the error had no indication what went wrong, all I saw was that the ```getCompositions``` call failed. After a bit of investigation, I could see that the call is rejected with [ev.error](https://github.com/remotion-dev/remotion/blob/main/packages/media-utils/src/get-video-metadata.ts#L15).

The problem is that it's undefined and the video errors are available under ```video.error```. So the Promise should be rejected with that instead.

I made a fiddle to show the difference: [https://jsfiddle.net/sashee/gtf9bock/1/](https://jsfiddle.net/sashee/gtf9bock/1/). You can see that ```e.error``` is ```undefined``` and ```video.error``` is a ```MediaError``` object with the details.

JonnyBurger:


JonnyBurger:
I verified that without it, it's a bug and with it, it's fixed!

PR #461, opened 2021-06-15
fix example

### LINKED ISSUE #453: `example` package is broken
Probably broke because of the latest change to the webpack config where we added fast refresh

JonnyBurger:
Fixes #453 

minor breaking change: it is expected that node modules are already compiled. we need to watch out for breakages but I think it's a good change, also speeds up bundling.

### DIFF
--- a/packages/bundler/src/webpack-config.ts
+++ b/packages/bundler/src/webpack-config.ts
@@ -163,6 +163,7 @@ export const webpackConfig = ({
 				},
 				{
 					test: /\.jsx?$/,
+					exclude: /node_modules/,
 					use: [
 						{
 							loader: require.resolve('esbuild-loader'),


packages/bundler/src/webpack-config.ts
```diff
@@ -163,6 +163,7 @@ export const webpackConfig = ({
 				},
 				{
 					test: /\.jsx?$/,
+					exclude: /node_modules/,
```
Iamshankhadeep, line 166: seriously this fixes the issue?

JonnyBurger: Yes, we try to add fast refresh to all node modules, which is probably not necessary anyway, and the fast refresh fails when doing so, since some node modules are too modern and don't include `module.exports` anymore. At least this is my understanding

Iamshankhadeep: Got it. Thanks for the awesome explanation. :fire:

PR #463, opened 2021-06-15
.MKV support

JonnyBurger:
This preset was created for Remotion Lambda, and is optimal for when concatenating multiple video clips into one.


Iamshankhadeep (approved):
lgtm. Thanks :fire:

PR #465, opened 2021-06-17
fix(docs): complete docusaurus ts setup

frencojobs:


JonnyBurger (commented):
Thanks a lot! 🙌 💙 

This is an issue which I observed but never knew how to solve exactly, so this is super appreciated! 🥳

JonnyBurger (approved):

PR #469, opened 2021-06-18
Small fixes in SSR doc page

JorgeArreolaS:
ssr doc page have one variable wrong (considering that frames and output video will be both on temp folder)
and import `path` is missing
Only that. 
Awesome btw


JonnyBurger:
Thanks a lot! 🙌  Good catch 🎣

PR #474, opened 2021-06-20
can render an audio with no inputs

JonnyBurger:
In lambda sometimes we need to generate a mp3 that has no sound. currently this is an error but it actually would make sense to support it.


Iamshankhadeep (approved):
LGTM. Thanks :fire:

PR #475, opened 2021-06-20
Fix render on js

### LINKED ISSUE #472: Rendering fails in plain JS project if no `remotion.config.js` exists
It should work without the config file, it should be optional.
Let's also test if the same bug occurs with .tsx file.

Iamshankhadeep:
Fixes #472


tests/rendering.test.ts
```diff
@@ -1,6 +1,7 @@
 import execa from "execa";
 import fs from "fs";
 import path from "path";
+import { complexSubtract } from "../packages/media-utils/dist/fft/complex";
```
JonnyBurger: Don't import from `dist` folders

packages/example/remotion.config.js
```diff
@@ -0,0 +1,4 @@
+import {Config} from 'remotion';
+
+Config.Rendering.setConcurrency((require('os').cpus() || {length: 1}).length);
```
JonnyBurger: `require('os').cpus()` should never return a falsy value, so why is it needed?

tests/rendering.test.ts
```diff
@@ -311,3 +312,40 @@ test("Dynamic duration should work", async () => {
...
+      "render",
+      "packages/example/src/ts-entry.tsx",
+      "framer",
+      outputPath,
+    ],
+    {
+      reject: false,
+    }
+  );
+  console.log(task.stderr,'err');
```
JonnyBurger: Remove console.log

JonnyBurger (changes requested):
The bug was actually that the project would not start if there is no `remotion.config.js`. But here we just add one, is the bug fixed?

Also need to remove the out.mp4 file before merging.

packages/example/remotion.config.js
Iamshankhadeep: cause on my system this was giving error  `os.cpus: Cannot read property 'length' of undefined` so I added this code.

tests/rendering.test.ts
Iamshankhadeep: okay thanks for noticing.

tests/rendering.test.ts
Iamshankhadeep: okay :fire:

Iamshankhadeep:
> The bug was actually that the project would not start if there is no `remotion.config.js`. But here we just add one, is the bug fixed?

Yeah I tested and the bug is fixed. It works now.

JonnyBurger:
Ah I see we return now. Thanks! 🔥

packages/example/remotion.config.js
JonnyBurger: Can you reproduce this in the console?

```
$ node
> require('os').cpus().length
```

Node.JS docs don't mention that this can happen: https://nodejs.org/api/os.html#os_os_cpus
So this is a Node.JS bug or what? 🤔

JonnyBurger (approved):

PR #476, opened 2021-06-23
Add emoji support to Dockerfile

brandonpearcy:
For context, there is only one emoji font supported by Alpine Linux: https://wiki.alpinelinux.org/wiki/Emojis
WIthout this change, emojis render as [?] characters.


JonnyBurger (approved):
Makes a lot of sense! Thanks a lot for the PR 🙌 That's a nice improvement 💙

Issue #317, opened 2021-04-27
[Idea] Detect/fix font-display: swap


sashee:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%2450%20Funded-%2300A156.svg" />](https://issuehunt.io/r/JonnyBurger/remotion/issues/317)

I've just got bit by the ```font-display=swap``` property. It's mentioned in [the docs](https://www.remotion.dev/docs/fonts) that:

> Google Web Fonts by default appends ?display=swap to their URLs. Make sure to remove it to ensure the video renders correctly if you have a slow internet connection.

But it turns out it's not just keeping in mind to remove the query parameter when I copy-paste the font URL as 3rd-party dependencies could also hardcode it in their code ([example](https://github.com/marp-team/marp-core/blob/main/themes/gaia.scss#L16)).

This got me thinking how to solve this automatically.

As the rendering process has access to the network requests the page makes ([ref](https://pptr.dev/#?product=Puppeteer&version=v9.0.0&show=api-pagesetrequestinterceptionvalue-cachesafe)), it could warn if there is a request that goes to the Google Web Fonts URL with ```?display=swap```. This would be a helpful warning message during the rendering. Maybe it's even possible to do this using a Service Worker in the preview.

But it can go further. The rendering could also check the *results* of CSS requests and see if there is a ```font-display: swap``` in them. This would make it not specific to Google Web Fonts. On the other hand, it might flag false positives.

But it can go further still. I think there is no need for ```font-display: swap``` during rendering *ever*. So the browser could remove that from all CSS responses (maybe even from inline CSS). This would solve this problem for good, and a hidden hardcoded ```?display=swap``` wouldn't result in flickering.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $50.00)

- [<img src='https://avatars.githubusercontent.com/u/1629785?v=4' alt='jonnyburger' width=24 height=24> jonnyburger](https://issuehunt.io/u/jonnyburger) ($50.00)

### Submitted pull Requests
- [#501 fix: await document.fonts.ready when rendering video](https://issuehunt.io/r/JonnyBurger/remotion/pull/501)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/JonnyBurger/remotion/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

[...]

PR #486, opened 2021-06-30
chore: create font-display: swap issue demo

arthurdenner:
Relates to [this comment](https://github.com/remotion-dev/remotion/issues/317#issuecomment-871690225).

## Steps to reproduce

- Clone this branch (arthurdenner:feat/wait-for-fonts)
- Run `npm run build -- --watch` in the `packages/renderer` folder
- Run `npm run render` in the `packages/example` folder

This will (hopefully) give you a video with a delayed font swapping, like the one below.
PS: You may need to run the `render` two times - maybe some cache in the build process?

<details><summary>Video</summary>

https://user-images.githubusercontent.com/13774309/124334814-9702ff00-db98-11eb-8a41-f64a572332a2.mp4

</details>

- Uncomment the `await page.evaluateHandle('document.fonts.ready');` line ([ref](https://github.com/arthurdenner/remotion/blob/feat/wait-for-fonts/packages/renderer/src/render.ts#L131));
- Verify the `npm run build -- --watch` output was updated just in case
- Run `npm run render` in the `packages/example` folder

This will (hopefully) give you a video *without* a delayed font swapping, like the one below, because we postpone rendering frames until fonts are ready.
PS: You may need to run the `render` two times - maybe some cache in the build process?

<details><summary>Video</summary>

https://user-images.githubusercontent.com/13774309/124334972-0e389300-db99-11eb-840e-f4cdf3696d0e.mp4

</details>

You can comment and uncomment the `await page.evaluateHandle('document.fonts.ready');` line a few times.
I was able to reproduce the bug and verify the proposed fix many times but sometimes I had to `npm run render` twice as mentioned above.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -4,12 +4,18 @@ import {checkMultipleRemotionVersions} from './multiple-versions-warning';
 
 declare global {
 	interface Window {
-		ready: boolean;
 		getStaticCompositions: () => TCompMetadata[];
 		remotion_setFrame: (frame: number) => void;
 		remotion_collectAssets: () => TAsset[];
 		remotion_isPlayer: boolean;
 		remotion_imported: boolean;
+		remotion_handlesReady: boolean;
+		remotion_fontsReady: boolean;
+	}
+	interface Document {
+		fonts: {
+			ready: Promise<void>;
+		};
 	}
 }
 
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -13,8 +13,8 @@ declare global {
 		remotion_fontsReady: boolean;
 	}
 	interface Document {
-		fonts: {
-			ready: Promise<void>;
+		fonts?: {
+			ready?: Promise<void>;
 		};
 	}
 }
--- a/packages/core/src/ready-manager.ts
+++ b/packages/core/src/ready-manager.ts
@@ -1,7 +1,16 @@
 import {getRemotionEnvironment} from './get-environment';
 
 if (typeof window !== 'undefined') {
-	window.ready = false;
+	window.remotion_handlesReady = false;
+	window.remotion_fontsReady = false;
+
+	document.fonts.ready
+		.then(() => {
+			window.remotion_fontsReady = true;
+		})
+		.catch(() => {
+			window.remotion_fontsReady = true;
+		});
 }
 
 let handles: number[] = [];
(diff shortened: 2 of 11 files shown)


JonnyBurger:
This is great! I had no idea about this API. Absolutely perfect. 

Also I found this by coincidence: https://github.blog/2021-06-22-framework-building-open-graph-images/

Shows how to correctly await document.fonts.ready

JonnyBurger:
@arthurdenner @sashee I pushed a commit that I think solves the problem: [`3189a90` (#486)](https://github.com/remotion-dev/remotion/pull/486/commits/3189a9056fb8e977f68a1139a1069607ac170658)

Do you want to take a look?

arthurdenner:
@JonnyBurger, have you tried the steps in this PR?

Only `await page.evaluateHandle('document.fonts.ready');` solves the issue I think. Maybe it wasn't clear but I only left the line commented in this PR so you could uncomment while going through the steps to verify the fix yourselves 😅 

Hook up the same API into Remotion internals seems like a valid option too, you definitely know better where the usage of the API is more secure/stable/whatever 😄

JonnyBurger:
@arthurdenner Hmm, ok I'm trying to understand it. So basically I created the same solution but it's not fully reliably and it gets stuck sometimes? 

Or is there no issue at all and this PR proposes a fix to be merged?

JonnyBurger:
By the way, I ran the render 100 times via a bash loop and it worked well, no getting stuck and it always waited for the font! Looking good

arthurdenner:
The changes you did seem to do the same effect as just `await page.evaluateHandle('document.fonts.ready');` does.

This PR is just a `chore` and not a `fix` because it contains more changes to build a reproducible scenario using one of the examples.
The purpose was to follow the steps in the description so you could verify `await page.evaluateHandle('document.fonts.ready');` works.

If we verify that it works fine (because there may be edge cases or other relevant parts of the codebase I'm not aware of), then I'd open a proper PR to fix it.

JonnyBurger:
Thanks for explaining! And sorry for getting lost and not understanding correctly, was indeed quickly switching between various issues 😌

In that case, I would suggest we just add that one line of code and merge it, what do you say? Just did a test and seemed reliable.

arthurdenner:
No worries, I know there is a lot going on in the repo and maybe I wrote too much text to explain this. 😆 

It's great that you could validate that change. Can I open a PR with that line only and if we find any issues, we can try your changes?

JonnyBurger:
Sounds good! Thanks a lot :)

arthurdenner:
Closing this in favor of #501.

PR #488, opened 2021-07-04
Refactoring and Twoslash Update

frencojobs:
I noticed quite a few things that needed to be done and fixed them up in the upstream twoslash via https://github.com/shikijs/twoslash/pull/57 and in this PR too.

This PR does -

- Outdate the requirements of override files
- Add a default copy-to-clipboard button with default a11y/i18n settings and a fix to a bug that current plugin in the docs have
- Highlight non twoslash codes

| Now      | Then |
| ----------- | ----------- |
|  now-yaml   |  then-yaml | 

- Better CSS for line highlights (the approach I used in the last PR is hacky and is overlapping over texts making them a bit faint in appearance)

| Now      | Then |
| ----------- | ----------- |
| now | then | 

*The above are done via upstream PR and this PR does -*

- Make titles together with the code blocks (possible due to upstream PR)

| Now      | Then |
| ----------- | ----------- |
| title-now | title | 

- Delete the format script I added last PR because I found out that it doesn’t work and even break the codes
- (Mostly manually) formatted all the docs

Formatting the code is trivial but the inconsistencies are bugging me so I did it. The codes are formatted based on your `.prettierrc` in the core package. But I changed two things, `semi` and `useTabs`. I opted out of the semicolons because some of the codes are without it and some are with it, so I chose not to include it to at least reduce byte. Tabs are switched to two spaces because default tab size in browsers I think are about 8 spaces, and make the codes so far indented and most code blocks won’t fit in.

Because manually formatted, the places where it makes sense as an example are carefully avoided and not formatted. For example, the interpolate function explanation’s comment `// — v —- v —-` and quite a few codes are *carefully* avoided.

Note that future documentations must be written to follow the same style because prettier or any formatted won’t work(or properly) inside markdown files.

👇 Bugs like these also should be fixed.

447A72E9-E34C-406D-8C62-E68D6007FE24

The documentation format fixes are quite a lot of tiny fixes to be reviewed. (I've revised the changes once but I might've missed anything.) So take your time. Let me know if anything. Byee.


JonnyBurger (approved):

JonnyBurger:
This is so great! Thanks for working on such an awesome project, and sending me another PR! Lots os cleanups and little improvements that I just learned about :) 🙌

PR #498, opened 2021-07-08
Freeze support

### LINKED ISSUE #381: [Idea] Add paused support for video
I record screencasts and I usually want them to play a bit, then pause so that I can explain a concept, then play a bit more, then pause again. I made a Webpack loader that splits an input video at keyframes and also outputs still images ([ref](https://github.com/sashee/video-loader)). This works great, but I'm looking for a more Remotion-friendly solution.

With the support for changing the speed of the Video element and the start-end arguments I can take care of the video scenes part. But AFAIK there is no way to show a single frame of a video as an image. It would be great if the Video element could support the paused attribute.

Iamshankhadeep:
fixes #381


JonnyBurger:
Thanks! Very nice idea with the demo on the docs page! 🙌 💙 

The component is perfect, I only did minor touchups. Let's ship it 🚀

JonnyBurger (approved):

PR #501, opened 2021-07-09
fix: await document.fonts.ready when rendering video

### LINKED ISSUE #317: [Idea] Detect/fix font-display: swap

arthurdenner:
Closes #317. Some discussion and context can be found at #486.

`document.fonts.ready` is a Promise that resolves when all of the fonts of a page are loaded ([MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Document/fonts)).

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#317: [Idea] Detect/fix font-display: swap](https://issuehunt.io/repos/274495425/issues/317)
---
</details>

### DIFF
--- a/packages/renderer/src/render.ts
+++ b/packages/renderer/src/render.ts
@@ -115,6 +115,7 @@ export const renderFrames = async ({
 
 		const site = `http://localhost:${port}/index.html?composition=${compositionId}`;
 		await page.goto(site);
+		await page.evaluateHandle('document.fonts.ready');
 		page.off('pageerror', errorCallback);
 		return page;
 	});
--- a/packages/renderer/src/render.ts
+++ b/packages/renderer/src/render.ts
@@ -115,7 +115,6 @@ export const renderFrames = async ({
 
 		const site = `http://localhost:${port}/index.html?composition=${compositionId}`;
 		await page.goto(site);
-		await page.evaluateHandle('document.fonts.ready');
 		page.off('pageerror', errorCallback);
 		return page;
 	});
--- a/packages/renderer/src/seek-to-frame.ts
+++ b/packages/renderer/src/seek-to-frame.ts
@@ -12,4 +12,5 @@ export const seekToFrame = async ({
 		window.remotion_setFrame(f);
 	}, frame);
 	await page.waitForFunction('window.ready === true');
+	await page.evaluateHandle('document.fonts.ready');
 };


JonnyBurger:
Thanks a lot! I moved the validation so it happens after each frame, I guess that is a possibility that a font comes in only at a later point.

I will run another 100 iterations of my test and approve it if it looks all good :)

You are also eligible for a bounty. Would you like to receive it via IssueHunt? 
The caveat is that you can only cash out if you have at least $100, so I am happy to also just send it via crypto or PayPal.

arthurdenner:
I think it makes sense to move the handle, great catch. And yes, I'm fine with IssueHunt.

JonnyBurger:
Cool! Thanks a lot again this is a really awesome enhancement! 🙌  Appreciate it much more than just the one line that it is in the end :)

JonnyBurger (approved):

PR #503, opened 2021-07-09
Change reference from remotion.config.ts to remotion.config.js

c0:
I found that using `.ts` didn't pick up the config file, but `.js` worked.


JonnyBurger:
Thanks! I'll consider it a bug report, because `remotion.config.js` is only used if the entry file is also a JS file. A bit confusing, I'll use your PR as a starting point to improve on it.

c0:
Ohh, that makes sense now. 👍

JonnyBurger:
Thanks for pointing it out, I will close it in favor of fb2f77d63430257eac7a3e32c079adad4d188562 where we check for existence of remotion.config.ts (first prio) or remotion.config.js (second prio).

c0:
🙌

PR #509, opened 2021-07-14
[Docs] Updates Default From Argument Value To 0

thorbenprimke:
I was reading the documentation and noticed that it says 1 for the default value for the `from` argument. 
Code has it as 0: https://github.com/remotion-dev/remotion/blob/main/packages/core/src/spring/measure-spring.ts#L8

Thanks for making Remotion!


JonnyBurger:
Thank you! 🙌 💙  Yes, it's obviously wrong!

JonnyBurger (approved):

PR #511, opened 2021-07-15
Prevent seekbar from loosing position after resizing browser window

Zayco:
When resizing the browser window after initial rendering of the player component the size variable (from useElementSize) does not get updatet resulting in the seekbar position no longer being in sync with the mouse pointer.

This PR adds an option to also trigger the useElemetSize hook when the browser window gets resized. It is optional to prevent unnecessaryre-renders when used in components that's not dependant on this event to function correctly.

There is probably a more elegant way to solve this but I am new to hooks and typescript in general so feel free to refactor.

### DIFF
--- a/packages/cli/src/editor/components/Canvas.tsx
+++ b/packages/cli/src/editor/components/Canvas.tsx
@@ -13,7 +13,7 @@ export const Container = styled.div`
 export const Canvas: React.FC = () => {
 	const ref = useRef<HTMLDivElement>(null);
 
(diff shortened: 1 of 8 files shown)


JonnyBurger (approved):
Thanks a lot for the PR! 🙌 💙  Appreciate you reporting and fixing it right away. I can see the problem, I fixed a few type errors and also enabled the window resize wherever the `left` property is being used because there the bug can also happen.

PR #517, opened 2021-07-17
feat: support <source> as <Video> children

### LINKED ISSUE #364: Allow <source> elements to be placed inside a <Video> or <Audio> element
We should rely mostly on browser mechanisms, but it will require some tweaks from our side on the video and audio components to achieve this.

mikaoelitiana:
fixes #364

### DIFF
--- a/packages/core/src/test/video.test.tsx
+++ b/packages/core/src/test/video.test.tsx
@@ -56,6 +56,18 @@ describe('Render correctly with props', () => {
 			)
 		).not.toThrow();
 	});
+	test('It should render Video with children source', () => {
+		expect(() =>
+			render(
+				<Wrapper>
+					<Video>
+						<source src="test-path-1" />
+						<source src="test-path-2" />
+					</Video>
+				</Wrapper>
+			)
+		).not.toThrow();
+	});
 	test('It should render Video with endAt props', () => {
 		expect(() =>
 			render(
--- a/packages/core/src/video/VideoForRendering.tsx
+++ b/packages/core/src/video/VideoForRendering.tsx
@@ -65,7 +65,7 @@ const VideoForRenderingForwardFunction: React.ForwardRefRenderFunction<
 	});
 
 	useEffect(() => {
-		if (!props.src) {
+		if (!props.src && !props.children) {
 			throw new Error('No src passed');
 		}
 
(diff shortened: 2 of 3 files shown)


JonnyBurger:
Thanks a lot for the PR! 😃 There's a small backlog of PRs just now, so I might need a few days to review it. Looking forward to it though!

packages/core/src/test/video.test.tsx
```diff
@@ -56,6 +56,18 @@ describe('Render correctly with props', () => {
 			)
 		).not.toThrow();
 	});
+	test('It should render Video with children source', () => {
```
arthurdenner, line 59: From what I've tested, the implementation in this PR still requires the user to pass the `src` prop.

If I understood correctly, we want to enable the user to choose between the `src` and the `children` props as the native tags do, right?

JonnyBurger: Yes I agree with what @arthurdenner said. Still needs some work, if necessary we can support you!

packages/core/src/video/VideoForRendering.tsx
```diff
@@ -65,24 +65,26 @@ const VideoForRenderingForwardFunction: React.ForwardRefRenderFunction<
...
-			id,
-			frame: absoluteFrame,
-			volume,
-			isRemote: isRemoteAsset(getAbsoluteSrc(props.src)),
-			mediaFrame: frame,
-			playbackRate: playbackRate ?? 1,
-		});
+		if (props.src) {
+			registerAsset({
+				type: 'video',
```
JonnyBurger, line 78: We need to fix it, if you pass a <source> now, then the audio of the video will not play in the final video.

Do we know of a reliable way to determine the source of the video if children have been passed?

packages/core/src/video/VideoForRendering.tsx
```diff
@@ -95,14 +97,15 @@ const VideoForRenderingForwardFunction: React.ForwardRefRenderFunction<
 		frame,
 		absoluteFrame,
 		playbackRate,
+		props.children,
```
JonnyBurger, line 100: ESLint warning is probably correct, the dependency array should include a value that makes the useEffect trigger when the children change.

mikaoelitiana: Let me check that

packages/core/src/video/VideoForRendering.tsx
mikaoelitiana: as `<video>` may have multiple `<source>`, should we take the `src` of the first child?

arthurdenner: @mikaoelitiana, when I was exploring this feature some weeks ago, I had the feeling that we can leverage the DOM events on the `audio` and `video` tags to find which source was chosen.
Then, we can pass that source to Remotion internals (as we currently do with `props.src` only).

The [`loadedmetadata`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/loadedmetadata_event) event fires when the browser has chosen a source and [`mediaElement.currentSrc`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/currentSrc) will contain the chosen source.

PR #519, opened 2021-07-21
Update fonts doc to reflect new behavior

### LINKED ISSUE #508: Document new font loading behavior
Currently we still suggest that you should disable font display swap, however from version 2.2 it's not necessary anymore, nothing has to be done. We should update the documentation but still mention that it's necessary for people still on older versions.

JonnyBurger:
Fixes #508


arthurdenner (approved):

Iamshankhadeep (approved):
LGTM

PR #527, opened 2021-07-30
Cycle browser tabs

JonnyBurger:
While rendering, in a 100ms interval, we cycle through the browser tabs and focusing them.

This is in the hope of increasing rendering speed and preventing tabs getting throttled because they become 'inactive'. 

Related PR: #176


Iamshankhadeep (approved):
This is a plus. 🔥 🔥

PR #550, opened 2021-08-06
Use `degit` in create-video

### LINKED ISSUE #546: New project from template includes the original template git history
After running `npm init video` or `yarn create video` the resulting project is just a cloned git repository of the template with the full git history of it.

It is expected that the created project has a clean history or no git set in place at all.

It would be better to use something like [degit](https://www.npmjs.com/package/degit), which is faster as it does not download the full git history (it downloads the `.tar.gz` file of repo and unpack it).

Iamshankhadeep:
Fixes #546

### DIFF
--- a/packages/create-video/package-lock.json
+++ b/packages/create-video/package-lock.json
@@ -1,21 +1,30 @@
 {
 	"name": "create-video",
-	"version": "2.1.9",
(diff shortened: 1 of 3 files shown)


JonnyBurger (approved):
Thank you! So much faster wow! 😲

PR #552, opened 2021-08-08
fix: config is given higher priority than cli flags

Iamshankhadeep:
While testing #527 found out that `remotion.config.ts` is given higher priority than remotion cli flags. 
This PR fixes that.

### DIFF
--- a/packages/cli/src/render.tsx
+++ b/packages/cli/src/render.tsx
@@ -58,14 +58,14 @@ export const render = async () => {
 	const file = parsedCli._[1];
 	const fullPath = path.join(process.cwd(), file);
 
-	parseCommandLine();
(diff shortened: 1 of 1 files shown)


JonnyBurger:
This is a bit finicky, it's because --overwrite is now treated as boolean, meaning that if it does not get specified, it defaults to false and then it gets cancelled.

I fixed it with my latest commit, though the drawback is `--overwrite=false` will now keep overwrite=true.

I think that's the smaller problem though, CI is passing and I am merging this.

JonnyBurger (approved):

Iamshankhadeep:
Ahh I see now why's CI is failing. Got it

PR #554, opened 2021-08-10
Shared audio tags for fixing autoplay issues on mobile

JonnyBurger:
Problem: If you play an audio file from say 10 seconds on, it will only then get mounted and on Mobile Safari, the autoplay policy does not allow it to get played without user interaction.

Solution: Mount 5 audio tags at the beginning and make them play once the user, with interaction, triggers the video to play. Once an audio tag gets mounted, move the props and ref to the pre-mounted audio tag, so it doesn't violate the audio policy

Also: If video cannot be played because of autoplay policy, we mute it and then play it. We should document this behavior.


Iamshankhadeep (approved):
LGTM.

PR #555, opened 2021-08-10
ffmpeg.wasm experimentation

JonnyBurger:
doesn't work yet at all and not sure if it ever will work so good that we'll ship it 😄

Currently we face this issue during stitching https://github.com/ffmpegwasm/ffmpeg.wasm/issues/241


JonnyBurger:
Closed in favor of the new Rust-based compositor.
I prototyped H264 encoding inside of it and it is way more promising.

In the browser we can use VideoEncoder API which should also perform much more reliably.

PR #572, opened 2021-08-24
respect log level everywhere

### LINKED ISSUE #567: Config setLevel does not work
# Bug Report 🐛

The `config.setLevel(X)` does not work. Or at least I'm not getting the desired output on the terminal.
I have tried both, setting it up from the `remotion.config.ts` file and from the command line.

On remotion.config.ts:
```Config.Log.setLevel('error');```

```"build": "remotion render src/index.tsx App out.mp4 --overwrite --crf 1 --log error",```

I'm expecting it to be "Silent except error messages" when the log level is set to "error".

Besides that, the correct value is "warn", not "warning" as the docs estipulate:
https://www.remotion.dev/docs/cli#--log
https://www.remotion.dev/docs/config/#setlevel

Thanks!

JonnyBurger:
Fixes #567 where progress bars would bypass log level.

### DIFF
--- a/packages/cli/src/get-filename.ts
+++ b/packages/cli/src/get-filename.ts
@@ -35,25 +35,25 @@ export const getOutputFilename = ({
 
 	if (extension === null && !imageSequence) {
 		if (codec === 'h264' || codec === 'h265') {
-			Log.info('No file extension specified, adding .mp4 automatically.');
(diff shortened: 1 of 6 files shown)


Iamshankhadeep (approved):
LGTM 🔥

PR #575, opened 2021-08-27
validate video has even dimensions

JonnyBurger:
Using ESLint plugin done ✅ 

And runtime check is also in place ✅


Iamshankhadeep (approved):
LGTM 🔥 . thanks

PR #578, opened 2021-08-28
`@remotion/media-utils`: Add `visualizeAudioWaveform()` and `createSmoothSvgPath()` APIs

### LINKED ISSUE #421: Audio wave visualization
There are two types of audio visualization, with bars and with wave, and we only support 1 right now. We should add the wave style for completeness too.

Iamshankhadeep:
fixes #421.


JonnyBurger:
TODOs:

- [ ] Update the `Audio visualization` page to mention the new API and give guidance on which API to choose for what (audio vs. music)
- [ ] New `normalize` property was added to getWaveformPortion and not documented. Will result in type error for people who upgrade. What shall we do?
- [ ] Better explain examples and maybe copy it over to the `Audio visualization` page
- [ ] Insert new voiceover (I ordered from Fiverr)

PR #597, opened 2021-09-19
feat: multiple sources in Audio and Video components

arthurdenner:
This is my take at #364 from a few months ago.
The approach is to leverage the HTML elements to make it work.

I stopped working on it as someone else was assigned but their PR is stuck for some time now.
I'm opening this as a draft because it needs to be rebased and tested in more scenarios.
I plan to get back to it soon but feel free to have a look and provide feedback in the meantime if you're interested.


packages/core/src/use-media-in-timeline.ts
```diff
@@ -58,23 +58,27 @@ export const useMediaInTimeline = ({
 	}, [duration, startsAt, volume, mediaVolume]);
 
 	useEffect(() => {
-		if (!mediaRef.current) {
+		const tagName = mediaType === 'audio' ? '<Audio>' : '<Video>';
+
+		if (!mediaRef.current || !hasMetadata) {
 			return;
```
JonnyBurger, line 65: We are silently returning if no metadata, and then we are not checking for `currentSrc` in the next statement. Should we swap the order of these statements?

packages/core/src/video/VideoForRendering.tsx
```diff
@@ -110,7 +119,7 @@ const VideoForRenderingForwardFunction: React.ForwardRefRenderFunction<
 			return getMediaTime({
 				fps: videoConfig.fps,
 				frame,
-				src: props.src as string,
+				src: videoRef.current.currentSrc,
```
JonnyBurger: Does it still work if the source gets changed?

We could test it by rendering 1 second with 1 video, then rendering 1 second with another video.

It's raising my eyebrow because src is not in the useEffect dependencies anymore.

JonnyBurger (changes requested):
Thanks a lot! This looks quality!

- Just a few comments that still need to be confirmed to not cause issue.
- Can we add example code to https://remotion.dev/docs/audio and https://remotion.dev/docs/video pages?
- I removed the `.skip()` from the tests, if it now fails, they need to be fixed

packages/core/src/use-media-in-timeline.ts
arthurdenner: I changed the approach to throw an error if no `src` attribute or child `source` is found inside `useMediaHasMetadata`.
I think that makes the check for `currentSrc` redundant, but I'm not sure... it's tricky 😕

packages/core/src/video/VideoForRendering.tsx
arthurdenner: Thanks for pointing it out. I've deconstructed `currentSrc` outside of the hook and added it to the dependencies array.

arthurdenner:
@JonnyBurger, I've updated the PR and fixed the broken tests.

However, there is still a bug in the preview that I didn't manage to solve (related to the preview) 😞 
Please read the description and check the video with a reproduction.

JonnyBurger:
@arthurdenner Thanks a lot! Need some time to review but will surely do in the coming week!

packages/core/src/test/audio-for-rendering.test.tsx
```diff
@@ -43,18 +44,37 @@ describe('Register and unregister asset', () => {
...
 		const props = {
 			src: 'test',
 			muted: false,
 			volume: 50,
 		};
-		const {unmount} = render(
+
+		Object.defineProperty(
+			global.window.HTMLMediaElement.prototype,
+			'currentSrc',
```
JonnyBurger, line 58: Clever! 🤓

PR #599, opened 2021-09-20
Fix Node.JS 14 Windows ENOENT bug 

JonnyBurger:


JonnyBurger:
does not fix the issue 🤔

Issue #518, opened 2021-07-21
By default, put the video in an `out` folder


JonnyBurger:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%24112%20Funded-%2300A156.svg" />](https://issuehunt.io/r/JonnyBurger/remotion/issues/518)

This simplifies the .gitignore template by allowing to just ignore the `out` directory instead of unignoring the .mp4 files in the `src` folder and also think it's nice to have a dedicated folder.

cc @aVileBroker who came up with the idea

---

🎃  This issue is eligible for **Hacktoberfest**!
In addition to helping you get a contribution towards your [Hacktoberfest](https://hacktoberfest.digitalocean.com/) reward, Remotion pays a $100 bounty for it!

**Read**: [How to contribute to Remotion](https://github.com/remotion-dev/remotion/blob/main/CONTRIBUTING.md)

- Let us know when you are interested in this issue, and we'll assign you!
- In order to prevent issues from getting stuck, try to progress the issue by opening a PR and adding commits. If your progress is stale for 7 days, another user can ask to take over the issue.

**Approval criteria for this PR:**
- Change `packages/cli/src/render.tsx` so that if you render the video into a folder, and that folder doesn't exist, you create it. The following command should work in the `example` folder: `npx remotion render src/index.tsx react-svg any/nested/folder/video.mp4`
- Go through all the templates we support, and instead of rendering to `out.mp4`, render to `out/video.mp4` (adjust the script in `package.json`). Make a separate PR for each of those templates!
- Also remove the `**.mp4` references in the `.gitignore` of the templates, and add `out` instead.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $112.00)

- [<img src='https://avatars.githubusercontent.com/u/1629785?v=4' alt='jonnyburger' width=24 height=24> jonnyburger](https://issuehunt.io/u/jonnyburger) ($112.00)

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/JonnyBurger/remotion/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

ahmadrosid:
Hi @JonnyBurger I would like to work on this issue.

JonnyBurger:
@ahmadrosid sure! Assigning you.

ahmadrosid:
[...]

PR #618, opened 2021-10-01
feat: render video to out folder

ahmadrosid:
## Description
Fixes https://github.com/remotion-dev/remotion/issues/518

## How to test?
- Build package
```bash
npm run build
```
- Go to example folder
```bash
cd packages/example
```
- Run render
```bash
../cli/remotion-cli.js render src/index.tsx react-svg video.mp4
```
- Render at sub folder
```bash
../cli/remotion-cli.js render src/index.tsx react-svg nested/video.mp4
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ lerna-debug.log
 .DS_Store
 coverage
 packages/example/.env
+packages/example/out
--- a/packages/cli/src/render.tsx
+++ b/packages/cli/src/render.tsx
@@ -166,6 +166,8 @@ export const render = async () => {
 		Internals.perf.logPerf();
 	}
 
+	let outputLocation = absoluteOutputFile
+
 	if (shouldOutputImageSequence) {
 		Log.info(chalk.green('\nYour image sequence is ready!'));
 	} else {
@@ -173,6 +175,15 @@ export const render = async () => {
 			throw new TypeError('CRF is unexpectedly not a number');
 		}
 
+		const outFolder = path.resolve(process.cwd(), './out');
+		outputLocation = absoluteOutputFile.replace(process.cwd(), outFolder)
+
+		if (!fs.existsSync(outFolder)) {
+				fs.mkdirSync(outFolder, {
+						recursive: true,
+				});
+		}
+	
 		const stitchingProgress = createProgressBar();
 
 		stitchingProgress.update(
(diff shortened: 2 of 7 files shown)


JonnyBurger (changes requested):
Thanks! This is a great start to make the folder if it doesn't exist, but the idea is to not always put it in an out folder, only by default.

So the hardcoded out path should be reverted, and instead the templates should be updated. See the acceptance criteria of the original issue again and let me know when you have questions!

ahmadrosid:
> Thanks! This is a great start to make the folder if it doesn't exist, but the idea is to not always put it in an out folder, only by default.
> 
> So the hardcoded out path should be reverted, and instead the templates should be updated. See the acceptance criteria of the original issue again and let me know when you have questions!

Sure @JonnyBurger updated, do we need to changes the example render from
```bash
remotion render src/index.tsx react-svg video.mp4
```
To
```bash
remotion render src/index.tsx react-svg out/video.mp4
```

JonnyBurger:
@ahmadrosid Yes would be nice, but most importantly a PR to each of the templates 😊

ahmadrosid:
> @ahmadrosid Yes would be nice, but most importantly a PR to each of the templates 😊

Here's PR to the templates @JonnyBurger 
- https://github.com/remotion-dev/template-helloworld/pull/4
- https://github.com/remotion-dev/template-three/pull/1

JonnyBurger:
Thanks! Almost there, what about the templates for plain JS, Still and Text-To-Speech?

JonnyBurger:
We also just merged a new template, `remotion-dev/template-empty` which would be nice!

ahmadrosid:
@JonnyBurger  Sure let me add more PR to templates for plain JS and remotion-dev/template-empty.

For the template-still is already have that build folder I believe.
- https://github.com/remotion-dev/template-still/blob/main/package.json#L8
- https://github.com/remotion-dev/template-still/blob/main/.gitignore#L6

And I can't find repo for this template `Text-To-Speech`.

ahmadrosid:
New PRs here:
- https://github.com/remotion-dev/template-helloworld-javascript/pull/3
- https://github.com/remotion-dev/template-empty/pull/1

JonnyBurger (approved):

JonnyBurger (approved):

JonnyBurger:
All shipped! Thanks a lot @ahmadrosid! 🎉

PR #619, opened 2021-10-01
Add getContainerNode to PlayerRef methods

evancloutier:
Adds a `getContainerNode` method to any instance of a `PlayerRef`. This will allow consumers to retrieve the container `HTMLDivElement` and handle any interaction events they'd like to manually (ex. click, double click events).

I haven't added an example into the documentation (I can if you'd like), but it would look something like:

```tsx twoslash {15}
import {useEffect, useRef} from 'react'
import {Player, PlayerRef} from '@remotion/player'
import {MyComposition} from './MyComposition'

const MyComp: React.FC = () => {
  const playerRef = useRef<PlayerRef>(null);

  useEffect(() => {
    function listener() {
      console.log("I have been clicked");
    }

    if (playerRef.current) {
      const container = playerRef.current.getContainerNode();

      container.addEventListener("click", listener);

      return () => {
        container.removeEventListener("click", listener);
      };
    }
  }, []);

  return (
    <Player
      ref={playerRef}
      durationInFrames={30}
      compositionWidth={1080}
      compositionHeight={1080}
      fps={30}
      component={MyComposition}.
    />
  )
}
```

### DIFF
--- a/packages/docs/docs/player.md
+++ b/packages/docs/docs/player.md
@@ -198,6 +198,10 @@ Pauses the video if it's playing. Plays the video if it's paused.
 
 Gets the current postition expressed as the current frame. Divide by the `fps` you passed to get the time in seconds.
 
+### `getContainerNode()`
+
+Gets the container HTMLDivElement of the Remotion Player. This method is useful if you'd like to handle Player click events yourself with `addEventListener`.
+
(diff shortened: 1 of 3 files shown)


JonnyBurger (approved):
LGTM! Thanks

Issue #616, opened 2021-10-01
Template "Empty"


JonnyBurger:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%24112%20Funded-%2300A156.svg" />](https://issuehunt.io/r/JonnyBurger/remotion/issues/616)

For those people who already know Remotion, it might be useful to skip the "Hello World" and start with an "Empty" template. This template should be the default Typescript template but a blank Canvas.

---

🎃  This issue is eligible for **Hacktoberfest**!
In addition to helping you get a contribution towards your [Hacktoberfest](https://hacktoberfest.digitalocean.com/) reward, Remotion pays a $100 bounty for it!

**Read**: [How to contribute to Remotion](https://github.com/remotion-dev/remotion/blob/main/CONTRIBUTING.md)

- Let us know when you are interested in this issue, and we'll assign you!
- In order to prevent issues from getting stuck, try to progress the issue by opening a PR and adding commits. If your progress is stale for 7 days, another user can ask to take over the issue.

**Approval criteria for this PR:**
- Make a new repo, `template-still`, which should contain the same code as https://github.com/remotion-dev/template-helloworld
- Remove the Hello World code and only have 1 composition, 1280x720 and empty.
- Remove the SSR code and dependencies.
- Adapt the README if necessary.
- Transfer the repo to the `remotion-dev` organization.
- Make a PR to add it to the list of templates.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $112.00)

- [<img src='https://avatars.githubusercontent.com/u/1629785?v=4' alt='jonnyburger' width=24 height=24> jonnyburger](https://issuehunt.io/u/jonnyburger) ($112.00)

### Submitted pull Requests
- [#620 add template-empty to the template list](https://issuehunt.io/r/remotion-dev/remotion/pull/620)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/JonnyBurger/remotion/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

aneeshrelan:
I'd like to give this a try. Can you please assign this to me?

Iamshankhadeep:
@aneeshrelan I am assigning it to you. all the best

PR #620, opened 2021-10-01
add template-empty to the template list

aneeshrelan:
Implements #616

### DIFF
--- a/packages/create-video/src/init.ts
+++ b/packages/create-video/src/init.ts
@@ -39,6 +39,11 @@ const FEATURED_TEMPLATES: TEMPLATES[] = [
 		name: 'FelippeChemello/Remotion-TTS-Example',
 		description: 'Turns text into speech and makes a video',
 	},
+	{
+		shortName: 'Empty',
+		name: 'remotion-dev/template-empty',
+		description: 'An empty canvas to get you started'
+	}
 ];
 
 function padEnd(str: string, width: number): string {


aneeshrelan:
@Iamshankhadeep @JonnyBurger I might need some help in transferring the repository. A simple transfer throws a permission error. The template is available at: [template-empty](https://github.com/aneeshrelan/template-empty) if you want to take a look and provide feedback

JonnyBurger:
Thanks a lot! It looks 90% done :D Just a few requests:

- We don't commit the lockfile to the templates, so it works with npm and yarn, so for this reason please remove `package-lock.json`.
- Can you also remove `express`, `@types/express` dependencies as well as `Dockerfile` and `.dockerignore`. (this is the minimalistic template)
- In `.github/workflows/render-video.yml` the two inputs should be removed as well, it's only used in the Hello World

Otherwise perfect! What permission error do you get when transferring the repo?

aneeshrelan:
Thank you for the feedback. Resolved all, please check :D

For the error, when I click on the "Transfer ownership" in the repo settings and mention the `remotion-dev` organisation. It throws an error saying "You don’t have the permission to create public repositories on remotion-dev". I might be missing something

JonnyBurger:
@aneeshrelan Ah ok, probably a GitHub thing. In that case, transfer it to `JonnyBurger` and then I'll transfer it again!

aneeshrelan:
@JonnyBurger Transferred. Please check

JonnyBurger (approved):
Thanks a lot! 🙌🏼

Issue #609, opened 2021-09-29
`useCurrentTime()` hook


marcusstenbeck:
# Feature Request 🛍️

Time-oriented helpers, for example `useCurrentTime()`.

## Use Case

I understand the technical reasons why most of everything in Remotion revolves around frames, but there's a good reason to use time in most things that should be re-usable. If a template or helper uses frames then it's locked to that frame rate, so I suggest that there should be helpers for working with time without having to consider frame rate.

## Possible Solution

At the very least Remotion should own a hook named `useCurrentTime()`.

```js
function useCurrentTime() {
  const config = useVideoConfig();
  const currentFrame = useCurrentFrame();

  return 1000 * (currentFrame / config.fps);
}
```

LoganArnett:
Opened a PR around this with an attempt to use minimal overhead

marcusstenbeck:
I'll let this issue rest for a bit, even though I still think time is a better measure from the human perspective. I understand the desire to use whole-number frame rates from a technical standpoint.

PR #628, opened 2021-10-07
feat: adding in a useCurrentTime hook

LoganArnett:
This PR is addressing #609 by adding a `useCurrentTime` hook.

This hook utilizes the `useVideo` and `useCurrentFrame` hooks in an attempt to calculate the current time with minimal overhead.

I added basic documentation in a new markdown file as well which is based on the `useCurrentFrame` docs as it is an almost identical hook in concept.

### DIFF
--- a/packages/core/src/use-current-time.ts
+++ b/packages/core/src/use-current-time.ts
@@ -1,13 +1,9 @@
-import { useCurrentFrame } from './use-frame';
-import { useVideo } from './use-video';
+import {useCurrentFrame} from './use-frame';
+import {useVideoConfig} from './use-video-config';
 
 export const useCurrentTime = (): number => {
-  const video = useVideo();
-  const currentFrame = useCurrentFrame();
+	const {fps} = useVideoConfig();
+	const currentFrame = useCurrentFrame();
 
-    if (video?.fps && currentFrame) {
-      return 1000 * (currentFrame / video.fps);
-    }
-
-    return 0;
+	return 1000 * (currentFrame / fps);
 };


packages/core/src/use-current-time.ts
```diff
@@ -0,0 +1,9 @@
+import {useCurrentFrame} from './use-frame';
+import {useVideoConfig} from './use-video-config';
+
+export const useCurrentTime = (): number => {
+	const {fps} = useVideoConfig();
+	const currentFrame = useCurrentFrame();
+
+	return 1000 * (currentFrame / fps);
```
JonnyBurger, line 8: This will return the current time in miliseconds. The doc examples assume seconds. Which unit it the proposed one?

packages/docs/docs/use-current-time.md
```diff
@@ -0,0 +1,38 @@
...
+```tsx twoslash
+import {Sequence, useCurrentFrame, useCurrentTime} from 'remotion'
+
+const Title = () => {
+  const time = useCurrentTime() // 25
+  return <div>{time}</div>
+}
+
+const Subtitle = () => {
+  const time = useCurrentTime() // 15
```
JonnyBurger, line 17: The time would actually not return `15`, since the `from` prop that is set below always accepts frames, not seconds. So the hook here would return whatever is the result of `(25 - (10 / fps))`.

Maybe this can be a new source of confusion for users?

JonnyBurger:
Sorry but I've decided to not add this feature to the core of Remotion. I have commented in #609 to give a few reasons why.
Still, thanks a lot for this contribution which is still very much appreciated and helped with the decision making. I hope it's not upsetting that we are closing it!

Issue #534, opened 2021-08-03
Needing to pass `compositionId` and `config` to renderFrames() is redundant


JonnyBurger:
This could be a 3.0 breaking change

PR #629, opened 2021-10-07
[v3] chore: making compositionId optional to start

LoganArnett:
This PR is addressing #534 with a graceful backward compatibility option instead of needing to be a breaking change. This is a possible way around introducing the breaking change while introducing the non-redundant behavior since we already have the ability to grab the composition ID within the config if we pass it.

For the breaking change version I can of course remove the compositionId option all together and remove its reference in the `render.ts` file with `config.id` being the only option.


JonnyBurger:
Yes of course! 😃 I'm also adding it to the other PR as well so hopefully both count!

Appreciate the nice words and your contributions!

PR #630, opened 2021-10-07
use onError property in every usage of `renderStill` and `renderFrames` in docs

### LINKED ISSUE #500: Recommend onError() handler for renderFrames() in docs
Also should check if unhandled exceptions are reported correctly, someone on Discord had a bit of struggle with it

danilhendrasr:
Closes #500 

Before I continue working on the rest of the docs, I wanted to know if what I did with commit https://github.com/remotion-dev/remotion/commit/814b0429c82df7679c6d212f2d5029cdf221fb08 is enough? or maybe I should improve it?

### DIFF
--- a/packages/docs/blog/2021-08-11-remotion-2-3.md
+++ b/packages/docs/blog/2021-08-11-remotion-2-3.md
@@ -91,6 +91,9 @@ await renderStill({
   composition,
   webpackBundle: bundleLocation,
   output: '/tmp/still.png',
+  onError: (error) => {
+    console.error("The following error occured when rendering the still: ", error.message)
+  },
   inputProps: {
     custom: 'data'
   }
--- a/packages/docs/docs/parametrized-rendering.md
+++ b/packages/docs/docs/parametrized-rendering.md
@@ -102,6 +102,13 @@ await renderFrames({
   webpackBundle: bundled,
   onStart: () => undefined,
   onFrameUpdate: (f) => undefined,
+  onError: (info) => {
+    if (info.frame === null) {
+      console.error('Got error while initalizing video rendering', info.error)
+    } else {
+      console.error('Got error at frame ', info.frame, info.error)
+    }
+  }
   parallelism: null,
   outputDir: framesDir,
   inputProps: {
--- a/packages/docs/docs/render-frames.md
+++ b/packages/docs/docs/render-frames.md
@@ -139,9 +139,9 @@ const renderFrames = (options: Partial<Parameters<typeof rf>[0]>) => {}
 renderFrames({
   onError: (info) => {
     if (info.frame === null) {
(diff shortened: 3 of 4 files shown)


packages/docs/docs/ssr.md
```diff
@@ -65,6 +65,11 @@ const start = async () => {
     // Get's called after bundling is finished and the
     // actual rendering starts.
     onStart: () => console.log('Rendering frames...'),
+    // Handle errors in your React code
+    onError: (error) => {
+      console.error(`Could not render frame ${error.frame} properly`);
+      console.error(`${error.error.message}`);
+		},
```
JonnyBurger, line 75: The closing bracket here is not aligned with the rest

packages/docs/docs/ssr.md
```diff
@@ -65,6 +65,11 @@ const start = async () => {
     // Get's called after bundling is finished and the
     // actual rendering starts.
     onStart: () => console.log('Rendering frames...'),
+    // Handle errors in your React code
+    onError: (error) => {
+      console.error(`Could not render frame ${error.frame} properly`);
```
JonnyBurger: `error.frame` can be `null`. This happens if the error occurred before seeking to the initial frame.
It would be nice if another comment could be added.

The word "properly" can also be removed, in that case it is a hard error not just a heads-up.

packages/docs/docs/ssr.md
danilhendrasr: That's strange :/ in my editor it's properly aligned though.

image

packages/docs/docs/ssr.md
danilhendrasr: I've found an example of using `onError` in [render-frames.md](https://github.com/remotion-dev/remotion/blob/d41e15d4122c76fb61c8e9b034f15905cb845c46/packages/docs/docs/render-frames.md#L139-L148). So I changed my code to match that example. Is it correct?

danilhendrasr:
As for `renderStill`s, I couldn't find anywhere in the docs where it seems I need to add `onError` usage.
Or do I need to add it [here](https://github.com/remotion-dev/remotion/blob/d41e15d4122c76fb61c8e9b034f15905cb845c46/packages/docs/docs/render-still.md#L43-L50) and [here](https://github.com/remotion-dev/remotion/blob/d41e15d4122c76fb61c8e9b034f15905cb845c46/packages/docs/blog/2021-08-11-remotion-2-3.md#L90-L97)?

JonnyBurger:
Yes, those two would be good to update! I am also not aware of any other occurences in the code.

packages/docs/docs/ssr.md
JonnyBurger: It's because of spaces vs. tabs :) I'll fix it

Issue #603, opened 2021-09-24
Remotion Preview: Play sub range of longer clip (working area)


marcusstenbeck:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%24112%20Rewarded-%237E24E3.svg" />](https://issuehunt.io/r/JonnyBurger/remotion/issues/603)

# Feature Request 🛍️

Be able to set an in/out point for the looping playback that is shorter than the viewed composition.

Here's a screenshot from Adobe After Effects illustrating their "Work Area" feature.

![image](https://user-images.githubusercontent.com/616906/134678902-78e1c6a4-f902-4bcd-9793-ae833a82cd9b.png)

## Use Case

A lot of the time I'm fine-tuning animations in a longer sequence and I just want to set an in/out point, a working area, while I'm doing that.

## Suggested scope

- Set in/out points by placing the playhead at a certain position and then click a UI button to set in and out point.
- Dragging is nice, but totally not a must-have, so skip it in first version
- When switching comps it's not necessary to remember the work area, nice to have, but not a must-have

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $112.00)

- [<img src='https://avatars.githubusercontent.com/u/1629785?v=4' alt='jonnyburger' width=24 height=24> jonnyburger](https://issuehunt.io/u/jonnyburger) ($112.00)

### Submitted pull Requests
- [#631 Add feature sub range (work area) for clip](https://issuehunt.io/r/remotion-dev/remotion/pull/631)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/JonnyBurger/remotion/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

JonnyBurger:
I like that idea!

arthurdenner:
Very cool idea indeed! Thanks for the suggestion!

JonnyBurger:
🎃  This issue is eligible for **Hacktoberfest**!
In addition to helping you get a contribution towards your [Hacktoberfest](https://hacktoberfest.digitalocean.com/) reward, Remotion pays a $100 bounty for it!

**Read**: [How to contribute to Remotion](https://github.com/remotion-dev/remotion/blob/main/CONTRIBUTING.md)

- Let us know when you are interested in this issue, and we'll assign you!
[...]

PR #631, opened 2021-10-07
Add feature sub range (work area) for clip

ankeshp03:
Resolve/Add feature #603 

This PR adds the feature to put in/out points (work area) for clips.

remotion

### Suggested Scope:
- [x] Set in/out points by placing the playhead at a certain position and then click a UI button to set in and out points.
- [x] When switching comps it's not necessary to remember the work area, nice to have, but not a must-have
- [ ] Dragging is nice, but totally not a must-have, so skip it in the first version

### Points to remember:
- In and Out points cannot be the same, the minimum gap is **1 frame**
- In-point cannot be greater than out-point and vice versa.
- Playhead cannot be set in position outside the working area (area between in and out points). 
   Playhead will automatically start from in-point on resuming play.
- In/Out point buttons on the preview toolbar are toggle buttons.

### Blocker in Scope
 - Implementing drag is challenging since we already have a ```TimelineDragHandler``` which changes the playhead position on drag. Drag for work area might affect the ```TimelineDragHandler``` functionality.


JonnyBurger:
Woah that looks great! Thanks a lot. I'm now gone for two days, expect some delay in reviewing, can't wait to look at it though 👀

Issue #254, opened 2021-04-17
Implement `loop` property for audio properly


JonnyBurger:

Maxou44:
I can help with ffmpeg filters if needed 😉
But I don't have much knowledge of the remotion code base

JonnyBurger:
@Maxou44 If you know how to implement loops with trimming support (start and end marker) and can give an example, it will help the implementation move forward faster for sure 😊

JonnyBurger:
as mentioned in #381 it would actually be cooler to have a `<Loop>` high order component so you can loop anything, not just audio and video! It's probably easier to implement and more versatile. So I am now leaning more towards this.

sohamsshah:
Hey @JonnyBurger ! Can I take up this issue?

JonnyBurger:
@sohamsshah Sure thing! Good luck 🎃

bpedersen:
I would like to take this issue if there is no one working on it

PR #638, opened 2021-10-13
Issue/254 loop component

bpedersen:
Created Loop component for https://github.com/remotion-dev/remotion/issues/254
* Created exampled of looped in example package
* Created tests
* Created documentation in docs package

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#254: `<Loop>` component](https://issuehunt.io/repos/274495425/issues/254)
---
</details>

### DIFF
--- a/packages/core/src/test/loop-validation.test.tsx
+++ b/packages/core/src/test/loop-validation.test.tsx
@@ -47,7 +47,7 @@ describe('Loop-validation render should throw with invalid props', () => {
 							<Loop durationInFrames={'1'}>hi</Loop>
 						</Wrapper>
 					),
-				/The "durationInFrames" prop of the \<Loop \/\> component must be a number, but you passed a value of type string/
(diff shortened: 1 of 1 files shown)


JonnyBurger:
Thanks a lot! Looking very good! 😊 I started reviewing it, and making some changes.

IMO, the desired behavior for durationInFrames=20, loop=2 is that the component loops two times, then disappears. So I made this change, do you agree with it?

Another thing I changed is to disallow durationInFrames=Infinity. That would mean it never loops!

Also I used Sequences again, so we rely less on Internals. It also allows us to display it more nicely in the timeline. I'm adding a "x3" badge to the timeline if you use the Loop component and started implementing it.

Appreciate it so much that you added tests and docs, and an example. Makes it so much easier to process! 😊

I'm doing some more refinements tomorrow for the timeline and will then merge it. No actions are needed from your side, unless you have some inputs. Thanks again for an excellent PR 😃

PR #661, opened 2021-10-21
localStorage error on Next.js

VirenMohindra:

### DIFF
--- a/packages/core/src/config/input-props.ts
+++ b/packages/core/src/config/input-props.ts
@@ -2,9 +2,25 @@ import {getRemotionEnvironment} from '../get-environment';
 
 export const INPUT_PROPS_KEY = 'remotion.inputProps';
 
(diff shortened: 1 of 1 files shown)


packages/core/src/config/input-props.ts
```diff
@@ -4,6 +4,10 @@ export const INPUT_PROPS_KEY = 'remotion.inputProps';
 
 export const getInputProps = () => {
 	if (getRemotionEnvironment() === 'rendering') {
+		if (typeof window === 'undefined') {
+			return;
```
arthurdenner:
```suggestion
			return {};
```
Based on the rest of the implementation, we should return an empty object.

VirenMohindra: Fixed, thanks @arthurdenner

PR #665, opened 2021-10-24
Add local FFmpeg option (Resolve #635)

### LINKED ISSUE #635: Using local ffmpeg
I would like to be able to use local, not just ffmpeg that exists in the PATH.

# Feature Request 🛍️

## Use Case
I'm working on an application that uses Remotion in Electron to generate videos from templates. So I would like to ask users to provide ffmpeg and place it in tmpDir and use it to render the video. (Due to licensing issues with ffmpeg)

## Possible Solution
Add an option to Config to specify the path to ffmpeg, like `setBrowserExecutable`.
https://remotion.dev/docs/config#setbrowserexecutable

If there is a way to still use local ffmpeg, please let me know.
Thanks.

ayatkyo:
## Description
PR to resolve #635
Add option to use local FFmpeg

## How to use

Add `--ffmpeg-executable` flag to command

```bash
remotion render <entry-file> <composition-id> <output-location> --ffmpeg-executable=<ffmpeg-executable>
```

Pass `ffmpegExecutable` option to `stitchFramesToVideo` when using remotion programmatically

```ts
await stitchFramesToVideo({
  // Specify FFmpeg executable here
  ffmpegExecutable: './supports/ffmpeg.exe',
})
```

### DIFF
--- a/packages/renderer/src/stitcher.ts
+++ b/packages/renderer/src/stitcher.ts
@@ -1,8 +1,6 @@
 import execa from 'execa';
-import {resolve} from 'path';
 import {
 	Codec,
-	Config,
 	FfmpegExecutable,
 	ImageFormat,
 	Internals,
@@ -27,7 +25,6 @@ import {resolveAssetSrc} from './resolve-asset-src';
 import {validateEvenDimensionsWithCodec} from './validate-even-dimensions-with-codec';
 import {validateFfmpeg} from './validate-ffmpeg';
 
-// eslint-disable-next-line complexity
 export const stitchFramesToVideo = async (options: {
 	dir: string;
 	fps: number;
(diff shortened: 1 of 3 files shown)


packages/renderer/src/stitcher.ts
```diff
@@ -78,12 +78,11 @@ export const stitchFramesToVideo = async (options: {
 	const isAudioOnly = encoderName === null;
 	const supportsCrf = encoderName && codec !== 'prores';
 
-	if (options.ffmpegExecutable) {
-		Config.Output.setFfmpegExecutable(resolve(options.ffmpegExecutable));
-	}
```
ayatkyo: I think this line is required, but we need to put it above `await validateFfmpeg();`.

Sorry, my bad miss reading `validateFfmpeg` and don't test server side rendering before making PR 😅

PR #699, opened 2021-11-14
Fix `seekTo` method not calling `ended` method

JonnyBurger:


JonnyBurger:
Published as `2.6.0-alpha.3+04636f3fa`

PR #704, opened 2021-11-18
renderMedia() API

JonnyBurger:


JonnyBurger:
Looks like frames may be added to FFMPEG in the wrong order. We need to ensure the right order before merging

PR #707, opened 2021-11-19
black background like in preview for transparent pixels

### LINKED ISSUE #702: Background in Blank template is white in export
If a background is not specified, the fallback color in the preview is black but it's white in the export.

Would be nice to have the same fallback colors. I tripped on this today. 

Workaround is to set a background in the composition.

JonnyBurger:
Fixes #702

### DIFF
--- a/packages/renderer/src/screenshot-dom-element.ts
+++ b/packages/renderer/src/screenshot-dom-element.ts
@@ -24,8 +24,11 @@ export const screenshotDOMElement = async ({
 
 	if (imageFormat === 'png') {
 		await page.evaluate(() => {
-			document.body.style.background =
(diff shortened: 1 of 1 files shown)


packages/renderer/src/screenshot-dom-element.ts
```diff
@@ -24,7 +24,8 @@ export const screenshotDOMElement = async ({
 
 	if (imageFormat === 'png') {
 		await page.evaluate(() => {
-			document.body.style.background = 'transparent';
+			document.body.style.background =
+				imageFormat === 'png' ? 'transparent' : 'black';
```
krlk89: line 25 already has `imageFormat === 'png'` check, isn't it redundant here?

PR #715, opened 2021-11-28
Enhancement: Update lambda upgrade docs and make removing lambda functions optional

VirenMohindra:
Small fix, but not sure if I need to extend the documentation or add additional reference to explain the rationale. For instance, adding this line after the step might help but also overcomplicate the upgrade process:

Refer to the second deploy exception in the [`FAQ`](/docs/lambda/faq#do-i-need-to-deploy-a-function-for-each-render) where one can run multiple lambda functions with different versions in the same region.

I thought maybe adding (optional) should be good enough, would appreciate some input regardless


JonnyBurger:
Thanks 😁 👍

PR #726, opened 2021-12-13
Fix asset types declaration

zoontek:
Hello 👋

I started using Remotion, but noticed you overwrite asset types declarations (for imports).
Problem: the value type is set to `any`, which make some typescript-eslint rules break.

Example with the `no-unsafe-assignment` rule:

issue-remotion

This PR set the exported value to `string`, as the import returned value will be a `string` URL.

### DIFF
--- a/packages/core/src/asset-types.ts
+++ b/packages/core/src/asset-types.ts
@@ -63,8 +63,4 @@ declare module '*.woff2' {
 	export default src;
 }
 
-declare module '*.css' {
-	const src: string;
-	export default src;
-}
-
+declare module '*.css';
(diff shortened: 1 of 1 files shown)


JonnyBurger:
@zoontek Thanks a lot, this is a good PR! 😃 I have to revert the .css declaration, because what is a common pattern

```
import styles from "./success.module.css";

style={styles.header}
```

After that, the CI should also be green, and I'll merge it!
Btw, I also use your libraries quite often, glad you found Remotion!

JonnyBurger (approved):

PR #766, opened 2022-01-09
expose `deviceScaleFactor` of puppeteer

### LINKED ISSUE #716: Expose `deviceScaleFactor` from puppeteer to create pixel perfect videos
Remotion is a great tool, especially for creating interface mockups/demo videos of either existing code or by being faster at writing code. While creating a demo video of a mobile web software I faced an issue regarding the resolution, as I needed the video dimensions to exactly match those of a smartphone. Creating the mobile phone scenes with a greater width/height would work as well but is unnecessarily complex and needs going through a lot of things. By providing direct access to the `deviceScaleFactor` parameter (maybe also other puppeteer options) this could be improved on.

# Feature Request 🛍️

Expose the `deviceScaleFactor` parameter of puppeteer for more granular control of rendering resolution. While being at it, maybe find a general solution to expose puppeteer parameters directly.

## Use Case

Generally good to be able to control all aspects of the rendering output. The use case for the `deviceScaleFactor` is about creating pixel perfect demo videos of application interfaces for demo videos/other content. 

## Possible Solution

In my opinion this would be either a setting for the rendering process or more generally a .env flag. I would love to implement this myself, but would need a little bit of help poking in the right direction! :)

Iamshankhadeep:
fixes #716


JonnyBurger:
What do you think of instead it being a parameter of a composition, it is a render option?
I was thinking an option like `--scale=2`

It's useful for when you want to export multiple densities for retina and non-retina for example

Compare with Figma:
image

Iamshankhadeep:
@JonnyBurger Sorry I did not saw your comment, yeah I was thinking of exposing it via cli, but when I start reading the code base I just implemented it as a parameter of a composition, I can totally expose it via cli, I'll do that and it does make more sense. Thanks 🔥

JonnyBurger:
Would be cool! Very useful for P5 animations which can be blurry at 1x, @lukezirngibl is waiting for this to land 😄

PR #774, opened 2022-01-13
Enhancement/add retention period for cloudwatch

VirenMohindra:
Hey, we came across this issue while trying to reduce our technical debt by going through logs for older remotion lambdas. Unfortunately two weeks had already passed (the current default retention period for CloudWatch logs is 14 days) and we couldn't view the logs.

This PR adds `cloudWatchLogRetentionPeriodInDays` as an optional param which can be passed to `deployFunction`. One can also pass it in via the CLI with `--retention-period`.

Points to note:
1. Naming convention can be improved because `--retention-period` sounds like the retention period for the entire lambda.
2. I had to explicitly type this line `as a number` since its an optional param: `retentionInDays: options.cloudWatchLogRetentionPeriodInDays as number,`. If there are better fixes please let me know!

I've added relevant documentation wherever needed and made the test suites pass to the best of my ability. Looking forward to our feedback :)


packages/docs/docs/lambda/cli/functions.md
```diff
@@ -55,6 +62,10 @@ Memory size in megabytes. Default: <DefaultTimeout /> seconds.
 
 Does not create a CloudWatch log group.
 
+### `--retention-period`
+
+Retention period for the CloudWatch Logs. Default: <DefaultLogRetention /> days.
```
tsejerome: Should we specify it as an integer for clarity?

tsejerome:
> 1. Naming convention can be improved because `--retention-period` sounds like the retention period for the entire lambda.

For clarity, Should we name it as smth like `--cloudwatch-logs-retention-period` / `--cloudwatch-logs-retention`?
@VirenMohindra  as we discussed offline there's no retention period for lambda but possibly a potential confusion for first time lambda user

JonnyBurger:
Thank you for a PR that contains everything!
Very much appreciate this contribution and the quality of it 😊 

I left it at `retention-period` as you probably anyway consult the documentation for it and it's shorter to type.

PR #777, opened 2022-01-15
feat: adds presignUrl and getS3Object helpers

imlegend19:
Adds the following helpers to Remotion's lambda API:

1. `presignUrl`: Returns a presigned URL for public access of an object stored in Remotion's S3 bucket.
2. `getS3Object`: Retrieves an object stored in Remotion's S3 bucket.


JonnyBurger (approved):

JonnyBurger:
Brilliant PR! Thanks a lot for the contribution 😊  💙 

I decided to remove the `getS3Object` API and instead implement getAwsClient(): https://remotion-git-fork-imlegend19-lambda-testing-mahen-remotion.vercel.app/docs/lambda/getawsclient

Your original intention, how you get a buffer of the object, is shown in the documentation page.

The reason for it is that there are probably many lesser-used operations that are desired, and instead of implementing all of them, we just expose the whole SDK and write examples for them, to save time and enable full flexibility.

PR #799, opened 2022-01-23
spring.md, + from & to defaults

tomByrer:
I assumed the defaults were 0 & 1, but had to [check the code](https://github.com/remotion-dev/remotion/blob/5b5893862d1070ea65114e894c9a2b80bde9a789/packages/core/src/spring/index.ts) anyway.


JonnyBurger:
Thanks sir! Good to make that clear.

JonnyBurger (approved):

PR #806, opened 2022-01-28
Recursively discover lockfiles

SkyezerFox:
Improves `bundler/src/get-package-manager.ts` to recursively ascend the file tree looking for valid lockfiles.

### DIFF
--- a/packages/bundler/src/get-package-manager.ts
+++ b/packages/bundler/src/get-package-manager.ts
@@ -18,27 +18,35 @@ export const lockFilePaths: LockfilePath[] = [
 	},
 ];
 
-export const getPackageManager = (): PackageManager | 'unknown' => {
+const discoverPkgManager = (currentPath: string): PackageManager | "unknown" => {
+	if (currentPath === "/") {
+		return "unknown";
+	}
+
 	const existingPkgManagers = lockFilePaths.filter((p) =>
 		fs.existsSync(path.join(process.cwd(), p.path))
 	);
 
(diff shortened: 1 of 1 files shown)


SkyezerFox:
Wow, https://github.com/remotion-dev/remotion/pull/806/commits/c8078e0db67142e82040defede9ce58548cffa53 proves i'm an idiot, thanks for the spot! I need more coffee...

JonnyBurger:
Hey, thanks a lot for filing the issue and sending a PR right away!

I already fixed this bug just a few hours earlier, and will make a new release shortly. (#803)

After some thinking, I decided to not merge this PR as it is because it could introduce two edge cases:
- On Windows, the root is not `"/"` but something like `C:\\` and the function will be an infinite loop
- On macOS, if we traverse up to the root of the filesystem, there will be permission popups (`Allow "node" to access the user directory?`)

The intention of this package manager detection feature is for it to power the `npx remotion upgrade` command. However, if the user is in a monorepo, maybe it is not wise for the feature to run the upgrade in the subpackage. Usually you want to update the package JSON and then run the installation command in the root. In that case it might be preferrable if the upgrade command fails.

I propose we close the PR as the original bug is fixed, unless you want to work out the edge cases.
Nonetheless thanks a lot for your report and initiative which I love 😊 🙌

SkyezerFox:
Okay! I wasn't sure if returning `unknown` had side effects or not in such a project, but if that will work fine, closing sounds sensible. Edge case fixing is boring!

PR #814, opened 2022-02-02
docs: Clarification of Player.props.component

SkyezerFox:
Small change to `@remotion/player` documentation to clarify that `Player.props.component` shouldn't take a `Composition` component, but the video component itself!

Spent the last 2 hours trying to work out why my video wasn't displaying, figured it could do with some clarification to prevent someone from doing the same!

### DIFF
--- a/packages/docs/docs/player/api.md
+++ b/packages/docs/docs/player/api.md
@@ -11,6 +11,8 @@ A React component which takes the following props:
 
 Pass a React component in directly **or** pass a function that returns a dynamic import. Passing neither or both of the props is an error.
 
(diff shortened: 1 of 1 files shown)


JonnyBurger (commented):
Thanks a lot! And sorry you had this problem happen to you 😌 what did you do?

`component={<Composition/>}`? 

We could add some validation for this as well.

please ignore CI failure, we are sorting it out right now ☮️

JonnyBurger (approved):

SkyezerFox:
Yeah, it wasn't particularly clear in the examples whether `MyComp` was a composition or a component - could be both! API doc should make it a little more obvious!

SkyezerFox:
Validation would be cool - I can do a PR for that tomorrow if I'm not too busy!

PR #820, opened 2022-02-03
feat(player): Add validation to Player.props.component

SkyezerFox:
Simple check to validate `Player.props.component` is not a `<Composition />` element, as rendering this has no visible result. Test included!


SkyezerFox:
It would be nice to check the validity of a function with signature `() => <Component />`, but that has potential side effects, so I don't think it's possible... likewise with lazy components.

JonnyBurger:
Thanks a lot! Quality!! 🙌

You are right, we cannot test for everything, but I added one more misunderstanding that can happen. Hope it helps someone!

JonnyBurger (dismissed):

SkyezerFox:
@JonnyBurger looks like my two commits from the closed PR the other day made it into this one? d044907b74cc548d9d84ca17bc628669bedc2e5b and c8078e0db67142e82040defede9ce58548cffa53. Looks like my forked repo is messy!

SkyezerFox:
> @JonnyBurger looks like the two commits from the closed PR the other day made it into this one? [d044907](https://github.com/remotion-dev/remotion/commit/d044907b74cc548d9d84ca17bc628669bedc2e5b) and [c8078e0](https://github.com/remotion-dev/remotion/commit/c8078e0db67142e82040defede9ce58548cffa53). Looks like my forked repo is messy!

Could've sworn I reset the fork to upstream but apparently not... what do you reckon is the best course of action?

JonnyBurger:
@SkyezerFox Oh you are right. Very strange, I think it has something to do with me using GitHub Desktop to check it out. It just added two commits to push. Don't know why, but I'll just remove those and then merge it. Thanks for catching this super quickly!

JonnyBurger (approved):

PR #834, opened 2022-02-09
remove angle as default renderer

### LINKED ISSUE #442: Debug excessive memory consumption during stitching
While rendering a 8 hour video:

> Hi, so it failed again. In my most recent run, I logged the amount of heap that was being used per loop going through the chunks, and it steadily increased until there was no more heap left. I had set ``--max-old-space-size` to 10gb, and it went through 785 iterations. <@!701030996430159872> how many chunks are there relative to the number of frames? It seems like just number of frames / 1000, but let me know if I'm correct. 

> If that's the case, I may be able to crank the heap size to 12-14gb and maybe get away with it, but right now it just seems like maybe there's literally too much data being downloaded.

### LINKED ISSUE #821: macOS Monterey
This is just to see if anyone else is having issues rendering on macOS Monterey? 

After upgrading to it, I just tried rendering a video and it fails miserably at about the 3000 frame mark, crashing my system. That's regardless if I set concurrency to just 1. The minute I start rendering, memory usage on my system climbs to 13G, that's even if I quit all other applications that are a little heavier on the memory usage. The memory pressure continues to rise until at about 3000 frames rendered, it climbs to over 15Gb and then my system with 16G RAM freezes up and crashes. 

I heard that Monterey 12.1 has some memory leak issues, so I upgraded to 12.2, but that didn't help at all.

I suspect it may be Chromium that's the problem. Not sure, though.

JonnyBurger:
I found `angle` to be eating all the RAM when rendering a longer video.
The new default is `null`. With the `--gl` flag and equivalent Node.JS APIs, users may be able to specify `angle` again.

Fixes #442 
Fixes #821

### DIFF
--- a/packages/core/src/config/chromium-flags.ts
+++ b/packages/core/src/config/chromium-flags.ts
@@ -3,11 +3,11 @@ import {
 	validateOpenGlRenderer,
 } from '../validation/validate-opengl-renderer';
 
-export const DEFAULT_OPENGL_RENDERER: OpenGlRenderer = 'angle';
+export const DEFAULT_OPENGL_RENDERER: OpenGlRenderer | null = null;
 
 let chromiumDisableWebSecurity = false;
 let ignoreCertificateErrors = false;
-let openGlRenderer: OpenGlRenderer = DEFAULT_OPENGL_RENDERER;
+let openGlRenderer: OpenGlRenderer | null = DEFAULT_OPENGL_RENDERER;
 let headlessMode = true;
 
 export const getChromiumDisableWebSecurity = () => chromiumDisableWebSecurity;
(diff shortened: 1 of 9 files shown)


alexweininger:
Did this memory leak also impact rendering images? I'm only rendering images (.png) with remotion and am trying to diagnose constantly increasing memory usage from the docker container running remotion.

JonnyBurger:
@alexweininger I think so unfortunately. With the exact same renderer, we noticed that with `angle` as the renderer, the Google Chrome memory increased consistently, whereas with the default renderer it did not happen. We therefore suspect a bug in Google Chrome. A potential workaround is to close and reopen the Google Chrome browser periodically: https://www.remotion.dev/docs/renderer/open-browser

PR #850, opened 2022-02-22
add gif output support

### LINKED ISSUE #718: Render as GIF
Provide GIF as an output format

Iamshankhadeep:
fixes #718


JonnyBurger:
I thought about it and I'm feeling that it will be common to render a video at 30fps but have the gif at a lower FPS, like 15 FPS. Should we add a flag for it? `--gif-fps`, or just `fps`? Maybe if you convert a 30 fps video into a 20 fps video, that will result in `useCurrentFrame` not returning an integer, that would lead to some edge cases. Maybe we should force the user to select a frame rate that is divisible by the original FPS. It could be a flag `--skip-n-frames=1` where only every second frame would be rendered (15fps), and then `skip-n-frames=2` would render in 10 fps.

The other thought I had: There are loopable and non-loopable GIFs. It would be nice to configure that as well!

I did not think about it too much yet 😃  So don't take it as concrete suggestions. But we should think about these details.

PR #878, opened 2022-03-12
feat: get compositions from CLI

### LINKED ISSUE #719: Get compositions from CLI
$ npx remotion comps src/index.tsx
my-comp-1 my-comp-2 

then you can script it with a bash loop to render all compositions at once

Slashgear:
Fix #719

- added new command in the CLI
- added documentation in website
- added documentation in help command


JonnyBurger:
Thanks a lot! This is a lovely feature 😊
I made it a bit richer - my plan is to, when v3 is launched, allow it to accept the new `-q` flag and in that case only print the IDs in a space-separated list. This has the benefit that you can loop over the IDs using a bash script 😊

JonnyBurger (approved):

PR #881, opened 2022-03-13
WIP: Demonstrate and fix a  bug in audio mixer

JonnyBurger:


JonnyBurger:
Fixed in another way in 3.0.4

PR #885, opened 2022-03-15
chore: define bugs package.json configuration

Slashgear:
## Why

Some public and published packaged don't have the `bugs` attribute in their `package.json`.
This is documented feature of npm which is a good practice to define.
Here is the PR that make sure define it for all.

## TODO

- [ ] I don't really understand why my package.json format doesn't seems to be correct 😅 

## Another idea

Found out that packages does not have `keywords` define.
It could be really nice to add some of them for `remotion` visibility on NPM

### DIFF
--- a/packages/babel-loader/package.json
+++ b/packages/babel-loader/package.json
@@ -16,10 +16,10 @@
 	"repository": {
 		"url": "https://github.com/remotion-dev/remotion"
 	},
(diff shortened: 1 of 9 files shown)


JonnyBurger (approved):
This is great! Thanks a lot 😃

Issue #730, opened 2021-12-15
Compositions grouping


gorhom:
Introduce compositions grouping on the side bar.

# Feature Request 🛍️

The request is to provide a way to group compositions on the side bar which will help users to focus on selected groups that they are working on.

## Use Case

For mid to large scale projects, we would en up with a lot of compositions that makes it difficult to find them, for example:

having multiple compositions that serve under the same category like:

![image](https://user-images.githubusercontent.com/4061838/146276293-aca8bf7a-eef4-4d9a-8a61-b2a5c4e020f3.png)

## Possible Solution

1. we could introduce a new grouping component that users has to wrap their compositions under like: 

```tsx
...
<Group id="video" open={true}>
  <Composition
    id="video-1"
  />
  <Composition
    id="video-2"
  />
</Group>
...
```

2. extract group id from compositions id prefix `{id}-name-name}` and then group every composition that include the group id as a prefix. this i have already done in [my fork](https://github.com/gorhom/remotion) which i will be happy to polish and submit a pr if needed ;)

https://user-images.githubusercontent.com/4061838/146276907-290a0445-3300-4b51-bdef-e7f768c5f4df.mov

Finally, thanks a lot for this amazing library @JonnyBurger

JonnyBurger:
Thanks a lot for the feature request @gorhom and sorry for responding only now!

I like the idea in general and would be interested in implementing a `<Folder />` component:

```tsx
export const Video = () => {
   return (
     <Folder name="Video Testing">
		<Composition
				id="video-testing-mp4"
				component={VideoTesting}
				width={1080}
				height={1080}
				fps={30}
				durationInFrames={100}
				defaultProps={{
					codec: 'mp4',
				}}
			/>
			<Composition
				id="video-testing-webm"
				component={VideoTesting}
				width={1080}
				height={1080}
				fps={30}
				durationInFrames={100}
				defaultProps={{
					codec: 'webm',
				}}
			/>
      </Folder>
      <Composition
				id="audio-testing"
				component={VideoTesting}
				width={1080}
				height={1080}
				fps={30}
				durationInFrames={100}
				defaultProps={{
					codec: 'webm',
				}}
	/>
   )
}
```

I prefer this explicit way rather than doing it automatically using the prefix. Would this also work for you?

tomByrer:
> interested in implementing a `<Folder />` component

Seems good for me.
[...]

PR #890, opened 2022-03-21
Composition grouping via Folder

kave-me:
Adds Composition grouping #730 requested feature


kave-me:
https://user-images.githubusercontent.com/27765198/159203253-39a81b1b-0816-4cdd-b2f6-adfacd743768.mp4

One issue that I have is I don't know why when I move my cursor on the sidebar the state of the expanded folder changes.

JonnyBurger:
Thanks for making this PR! Great start! We can iterate from here. Some feedback:

- Instead of hovering, I think clicking to expand / collapse would be better. It is more often seen in file tree structures like windows explorer / finder so it will be more familiar. also, it will be easier to implement for sure!

- To render the tree, you could to the following:
1. Create an empty array of list items
2. Loop over the compositions
3. If the composition is not in a folder, add it to the array you created
4. If the composition is in a folder, add a folder to the array you created, if the folder is not already in there.
5. If the composition is in a folder and the folder is already in the array of list items, add it to the children of the folder you already created.
6. Render through all list items, if it is a composition, render a `CompositionSelectorItem`, otherwise render a `<Folder>` component. The Folder component should also re-use the `CompositionSelectorItem` instead of re-writing it!

Using that algorithm, we can keep the order of the folders and make sure each folder only gets rendered once.

The data structure could look something like this:

```
[
	{type: 'folder', folderName: 'my-folder', children: [
		{type: 'composition', composition: {id: 'my-comp' /*duration, fps, width... */}}
	]},
	{type: 'composition', composition: {id: 'my-comp-2'}
]
```

(making it easy to add nested folders in the next step!)

JonnyBurger (approved):
Thanks a lot Kave for kicking it off! Finally managed to finish it, a great feature for Remotion 🙂  🙌

PR #916, opened 2022-04-07
fix #894 Make spawned child process detached on windows

### LINKED ISSUE #894: The "Open in Editor" functionality does not work on Windows
The functionality is in 2 places: In the Error overlay and after the project creation in the CLI.

In the latter case, I tried to open is and it would not open it. To be investigated.

Devinpennings:
When the promise of `init()` in `bin.js` of the `create-video` package resolves, `process.exit()` is called. This causes the spawned child process, containing the editor, to exit as well.

To fix this, I changed the spawned child process to be detached for 'open in editor' on windows

### DIFF
--- a/packages/bundler/src/error-overlay/react-overlay/utils/open-in-editor.ts
+++ b/packages/bundler/src/error-overlay/react-overlay/utils/open-in-editor.ts
@@ -167,7 +167,15 @@ export const getDisplayNameForEditor = (
 		return null;
 	}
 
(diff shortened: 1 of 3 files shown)


JonnyBurger:
Brilliant! Confirmed working on Windows now 🤩

JonnyBurger (approved):

JonnyBurger:
Thank you! 💙

PR #948, opened 2022-04-26
refactor: alias method getAudioDuration to getAudioDurationInSeconds

### LINKED ISSUE #940: getAudioDuration() should include a unit
It should be `getAudioDurationInSeconds()` or include a JSDoc comment that it is in seconds

Slashgear:
### Why

Close #940 

I did not want to include a breaking change here, so here is my proposition to just alias the method.

### How

- [x] update documentation
- [x] update examples


JonnyBurger (approved):
Thanks a lot! 🙌

PR #958, opened 2022-04-30
chore: update turborepo version to latest

Slashgear:
- update turborepo version to latest


JonnyBurger (approved):
Excellent! Thanks a lot 😁  Gonna dig in what's new now!

PR #1011, opened 2022-05-26
Render as GIF

Iamshankhadeep:
fixes https://github.com/remotion-dev/remotion/issues/718


packages/cli/src/get-cli-options.ts
```diff
@@ -235,6 +235,8 @@ export const getCliOptions = async (options: {
 			Internals.getChromiumOpenGlRenderer() ??
 			Internals.DEFAULT_OPENGL_RENDERER,
 	};
+	const skipNFrames = Internals.getSkipNFrames();
```
JonnyBurger: Bug with --skip-n-frames=1:
The GIF plays twice as fast if you use it

packages/cli/src/parse-command-line.ts
```diff
@@ -114,6 +120,10 @@ export const parseCommandLine = (
 		Config.Puppeteer.setChromiumHeadlessMode(false);
 	}
 
+	if (parsedCli['every-nth-frame']) {
+		Config.Rendering.setEveryNthFrame(parsedCli['every-nth-frame']);
```
JonnyBurger: We must forbid users setting this property when not rendering a GIF, as it will mess up the audio output

packages/cli/src/parse-command-line.ts
```diff
@@ -92,6 +94,10 @@ export const parseCommandLine = (
 		);
 	}
 
+	if (parsedCli.loop) {
+		Config.Rendering.setLoop(parsedCli.loop);
```
JonnyBurger: For DX, let's forbid this property when not using a GIF codec

packages/renderer/src/render-frames.ts
```diff
@@ -62,6 +62,7 @@ type RenderFramesOptions = {
 	parallelism?: number | null;
 	quality?: number;
 	frameRange?: FrameRange | null;
+	everyNthFrame: number;
```
JonnyBurger: Must make this optional before release to keep backwards compatibility

packages/cli/src/progress-bar.ts
```diff
@@ -94,14 +95,17 @@ export const makeStitchingProgress = ({
...
 	stage,
+	codec,
 }: StitchingProgressInput) => {
 	const progress = frames / totalFrames;
 	return [
 		`(${steps.indexOf('stitching') + 1}/${steps.length})`,
 		makeProgressBar(progress),
 		stage === 'muxing'
 			? `${doneIn ? 'Muxed' : 'Muxing'} audio`
-			: `${doneIn ? 'Encoded' : 'Encoding'} video`,
```
JonnyBurger, line 104: We should skip mixing and adding an audio track when making a GIF

packages/renderer/src/validate-fps-for-gif.ts
```diff
@@ -0,0 +1,9 @@
+import {Codec} from 'remotion';
+
+export const validateFpsForGif = (codec: Codec, fps: number) => {
+	if (codec === 'gif' && (fps > 50 || fps < 1)) {
```
JonnyBurger: Unused and should apply after the recalculated GIF (after `everyNthFrame`) is taken into account

packages/renderer/src/stitch-frames-to-video.ts
```diff
@@ -206,6 +207,7 @@ export const spawnFfmpeg = async (
 		options.onProgress?.(Math.round(totalFrameProgress));
 	};
 
+	// TODO: Don't render audio if GIF
```
JonnyBurger: Disable audio if rendered as a GIF

packages/renderer/src/get-extension-from-codec.ts
```diff
@@ -27,6 +27,8 @@ export const getFileExtensionFromCodec = (
 			return 'webm';
 		case 'vp9':
 			return 'webm';
+		case 'gif':
```
JonnyBurger, line 29: On Lambda, the output works but looks horrible.
Also it only takes the first chunk - should we encode the chunks to another codec?

Also combineVideos() must then take `loop` and `skip-n-frames` into account

JonnyBurger: Command to experiment with in `packages/example`:

pnpx remotion lambda render testbed framer --out-name=ggg.gif ggg.gif --codec=gif --log=verbose

packages/lambda/src/cli/commands/render/render.ts
```diff
@@ -66,6 +66,7 @@ export const renderCommand = async (args: string[]) => {
 		quality,
 		scale,
 		everyNthFrame,
+		// loop=0 results in infinite
```
Iamshankhadeep: `loop=0` should result in infinite loop.

- -1 no loop (plays once)
- 0 infinite loop (default)
- 1 loop once (plays 2 times)
- 2 loop twice (plays 3 times) 

[source](https://superuser.com/a/1607134/1665061)

JonnyBurger: I see - very confusing from FFMPEG -  shall we change this to not inherit the FFMPEG options directly?

`loop=0` I think is not a intuitive syntax for people to make it loop indefinitely 🤣

What would make sense IMO

loop = 0 ==> Don't loop
loop = 1 ==> Play once
loop = 2 ==> Play twice
loop = null ==> Loop indefinitely

Iamshankhadeep: by default gif will play one time, so this will turn out something like this right?
loop = 0 ==> ~Don't loop~ plays once ( as gif needs to be played once )
loop = 1 ==> ~Play once~ plays twice ( as one play for default + 1 play for the loop ) 
loop = 2 ==> ~Play twice~ plays 3 times ( as one play for default + 2 play for the loop ) 
loop = null ==> Loop indefinitely

right? @JonnyBurger

JonnyBurger: You are of course right @Iamshankhadeep 🙈

packages/lambda/src/cli/commands/render/render.ts
```diff
@@ -104,6 +107,8 @@ export const renderCommand = async (args: string[]) => {
 		timeoutInMilliseconds: puppeteerTimeout,
 		chromiumOptions,
 		scale,
+		loop,
+		everyNthFrame,
```
JonnyBurger, line 110: Rendering on Lambda with everyNthFrame = 5 does not yet work

JonnyBurger: Failing test case added

packages/lambda/src/functions/launch.ts
```diff
@@ -105,17 +105,20 @@ const innerLaunchHandler = async (params: LambdaPayload, options: Options) => {
...
 
-	const frameCount = RenderInternals.getDurationFromFrameRange(
+	const frameCount = RenderInternals.getFramesToRender(
 		realFrameRange,
-		0
+		comp.durationInFrames,
+		params.everyNthFrame
 	);
 
+	// TODO: Validate if with everyNthFrame, this makes sense
```
JonnyBurger: Does not work with everyNthFrame yet

JonnyBurger:
@Iamshankhadeep I am planning to ship it tomorrow, any last changes or suggestions, let me know!

PR #1037, opened 2022-06-10
remove express dependency

JonnyBurger:


tomByrer:
+1; less is more
& thanks for remembering the overlay!

PR #1044, opened 2022-06-13
fix little spelling mistake

DMC-CMD:
...animations are a beautiful way...
instead of animations are beautiful way


JonnyBurger (approved):
Thanks! 🙌

PR #1046, opened 2022-06-14
cleanup assets folder after every render

JonnyBurger:
some doubts:

- [ ] what if asset gets downloaded during getCompositions() and then deleted, then renderMedia() still thinks it is in cache
- [ ] what if render fails or user cancels it, it should still clean up


JonnyBurger:
Already implemented in another PR

PR #1049, opened 2022-06-15
improve DX of SSR.md example

tomByrer:
More  newbie help:
Extra troubleshooting;: logs steps (instead of long pause) & filename feedback.
filename based on compositionId

tested in: https://github.com/tomByrer/remotion-simple-demo


JonnyBurger (approved):
Thanks a lot! Makes sense. Did small tweaks (dont hardcode .jsx, object shorthand)

tomByrer:
Welcome :)
I keep forgetting the 'object shorthand', thanks for remembering.

PR #1065, opened 2022-06-21
Inline puppeteer source code and get rid of dependencies

JonnyBurger:


tomByrer:
Thanks!  It was a bit confusing for me trying to read the Puppeteer parts; many little functions.

PR #1073, opened 2022-06-25
New concurrency model

JonnyBurger:
Many frames on one page


JonnyBurger:
Exploring different strategies for speeding up renders!

PR #1079, opened 2022-06-30
Remove depreciated renderVideoOnLambda() from docs

yuvraj108c:
- Replaced the depreciated function `renderVideoOnLambda()` with `renderMediaOnLambda()` in the docs.


JonnyBurger (approved):
Thanks a lot! 🙌  How did we miss this!

PR #1088, opened 2022-07-03
Updated instructions acc. to AWS Console Upadates

Just-Moh-it:
1. The "Common use cases" section was renamed to just "Use Cases"

Screenshot 2022-07-03 at 10 26 18 AM

1. Some sections have also been removed in "Create role" section

Screenshot 2022-07-03 at 10 27 53 AM

1. Added a more defined location for the "Add Inline Policies" option. It took me half a minute to figure out (really un-usual location in the console)


JonnyBurger:
Thanks a lot for the update! 🙌 💙

PR #1094, opened 2022-07-05
allow <Sequence> to have a style

JonnyBurger:
Does not work if `layout="none"` is specified


tomByrer:
So now `<Sequence>` may act like a full background color/image?  Or do you still have to specifically `width:100%;height: 100%`?

JonnyBurger:
@tomByrer It works like an AbsoluteFill, but now also allows the style to be overriden

PR #1108, opened 2022-07-11
Fix <OffthreadVideo> race condition leading to "Cannot read .trim() of undefined

JonnyBurger:


JonnyBurger:
@hfellerhoff Thanks a lot for reporting! New release coming tomorrow morning!

PR #1156, opened 2022-07-22
Fix Skia template installation in docs for yarn and pnpm

byCedric:
Here you go! I don't think this is right :)

image

image


JonnyBurger (approved):
Thanks a lot! 🙌

PR #1158, opened 2022-07-22
Fix minor docs issues

cfhull:
The markup was rendering incorrectly for the new `downloadBehavior`. I think the backticks were types before the hashes, which apparently confuses the parser.

Previous:
image

New:
image

### DIFF
--- a/packages/docs/docs/lambda/rendermediaonlambda.md
+++ b/packages/docs/docs/lambda/rendermediaonlambda.md
@@ -165,7 +165,7 @@ How the output file should behave when accessed through the S3 output link in th
 Either:
 
 - `{"type": "play-in-browser"}` - the default. The video will play in the browser.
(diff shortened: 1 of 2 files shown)


JonnyBurger:
Thanks a lot! It's because the character is ` ` vs ` ` (Option + Space instead of Space).
I've raised the issue, but we'd have to change the markdown spec to fix it... in the meanwhile I installed an editor plugin that highlights the bad character.

Appreciate you finding an instance of the bug and sending a PR! 🙌

JonnyBurger:
CI is failing due to the new pnpm version being rolled out - no need to do anything from your side!

cfhull:
> CI is failing due to the new pnpm version being rolled out - no need to do anything from your side!

You

cfhull:
> Thanks a lot! It's because the character is ` ` vs ` ` (Option + Space instead of Space). I've raised the issue, but we'd have to change the markdown spec to fix it... in the meanwhile I installed an editor plugin that highlights the bad character.
> 
> Appreciate you finding an instance of the bug and sending a PR! 🙌

Good to know! Thanks for adding this feature. Our clients will appreciate the one-click downloads once we go live.

PR #1162, opened 2022-07-24
Update Skia instructions for rendering

wcandillon:
@JonnyBurger part of this PR, this section might need to be updated: https://www.remotion.dev/docs/gpu#the-way-forward-chrome-98 and based on what is the source of truth there I can update my PR as well. Right now it is a bit confusing.


JonnyBurger (commented):
Thx!!

JonnyBurger (approved):

JonnyBurger:
The template in `npx create video --skia` already has `gl=angle` by default in it's config file!

PR #1163, opened 2022-07-24
Improve consistency on Rendering with Skia

wcandillon:
The previous version was implying that the angle option was GPU emulation.


JonnyBurger (approved):

JonnyBurger:
Makes sense! Good catch thank you 🙌

JonnyBurger (approved):

PR #1195, opened 2022-08-02
feat: adds privacy support for deploying s3 site

imlegend19:
Support for deploying the private s3 site.


JonnyBurger (changes requested):
It needs to be clear that this does not work out of the box and that the user needs to set up a private CloudFront URL. Some basic instructions for that would be nice too!

alexey-pelykh:
@imlegend19 do you plan to further advance this PR?

PR #1196, opened 2022-08-02
feat: add @remotion/lottie package

arthurdenner:
Hey! 👋🏽 

As discussed on Discord, here's a slightly better version of what I've done in the [test repo](https://github.com/arthurdenner/remotion-lottie-test).

Although it's possible to provide the path to a JSON file to `lottie-web`, I've opted to support the `animationData` only, which puts an extra task on the developers but have the following benefits:

- We don't need to manage when the data is ready to continue the render
- There won't be any flickering during render due to mount/unmount (in case of sequences)
- There won't be any flickering during render due to frames not being loaded fast enough
> Please, feel free to leave your input in this decision

There are a few examples copied from the test repo. I tried to improve them a bit but I must admit they lack creativity a bit 😅 I mainly was focused on having different use cases/examples:

- Importing `animationData` from a file
- Importing one animation from a URL
- Importing multiple animations from different URLs using Promise.all
- Using Loops and Sequences to build animations based on markers

Some next steps:

- Create an example with the Player (I guess it's essential to test with it, right?)
- Write documentation about the package itself (e.g. a README hehe)
- Write documentation explaining the decisions taken, fetching data, etc

I didn't explore other properties and methods from [`AnimationItem`](https://github.com/airbnb/lottie-web/blob/master/index.d.ts#L6), e.g. `setDirection`, `setSubframe`, `renderer` as I'm not so familiar with Lottie. Hopefully, others can contribute to these areas or we can explore them later.

Every feedback is appreciated, please leave yours!
And thanks for the interest in moving my POC into an official package!


JonnyBurger:
I agree with the decision to only support animationData! We can still create abstractions in the future to make make the API more convienient. 

One change I will make is that `lottie-web` is a peer dependency. It will allow users to upgrade lottie themselves or choose a specific version.

Going to write some basic docs now!

JonnyBurger (approved):
Awesome work! I even tried the After Effects workflow and it worked very nicely! I also wrote 6 documentation pages to cover everything one might wanna know.

For me this is ready to ship, unless you find something off!

packages/docs/docs/lottie/getlottiemetadata.md
```diff
@@ -0,0 +1,211 @@
...
+    },
+  ],
+  markers: [],
+};
+
+// @filename: Animation.tsx
+import { animationData } from "./animation";
+
+// ---cut---
+import { getLottieMetadata } from "@remotion/lottie";
```
arthurdenner, line 159: Although `@remotion/lottie` is declared on the `package.json`, this is still throwing an error for me locally. It happens on every package when the package is imported. Does it happen for you? Any idea why?

<details>
<summary>Screenshot</summary>

image


</details>

packages/docs/docs/lottie/index.md
```diff
@@ -0,0 +1,70 @@
...
+</Tabs>
+
+You can now start using the [`<Lottie>`](/docs/lottie/lottie) component in your Remotion project.
+
+## Supported features
+
+- Playing Lottie animations using `lottie-web`
+- Change the speed of the animation
+- Playing remote files
+- Determinining dimensions and duration of a Lottie animation
```
arthurdenner:
```suggestion
- Determining dimensions and duration of a Lottie animation
```

packages/example/src/Lottie/Cybertruck/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+	const [handle] = useState(delayRender);
+
+	useEffect(() => {
+		// Credits: https://lottiefiles.com/11643-tesla-cybertruck
+		fetch('https://assets2.lottiefiles.com/packages/lf20_RqpTFh.json')
+			.then((res) => res.json())
+			.then(setAnimationData);
+	}, []);
+
+	useEffect(() => {
```
arthurdenner: Should we align the examples and the documentation regarding using two effects or just one? I personally prefer to break them but if there's no risk in calling `continueRender` after setting the state instead of after the component rerenders, I'm fine.

packages/example/src/Lottie/Halloween/Balloons.tsx
```diff
@@ -0,0 +1,44 @@
+import {Lottie} from '@remotion/lottie';
+import React from 'react';
+import {interpolate, Sequence, useCurrentFrame, useVideoConfig} from 'remotion';
+import balloonsAnimation from './balloons.json';
```
arthurdenner: We didn't document importing from a file directly (which works fine I think). However, should we update this example to use `staticFile`?

packages/lottie/src/Lottie.tsx
```diff
@@ -0,0 +1,108 @@
...
+export const Lottie = ({
+	animationData,
+	className,
+	loop,
+	playbackRate,
+	style,
+}: LottieProps) => {
+	if (typeof animationData !== 'object') {
+		throw new Error(
+			'animationData should be provided as an object. If you only have the path to the JSON file, load it and pass it as animationData. See https://remotion.dev/link-tbd for more information.'
```
arthurdenner: We need to update this error message. I put `https://remotion.dev/link-tbd` as a placeholder, which page should we link it to?

packages/lottie/src/index.ts
```diff
@@ -0,0 +1,3 @@
+export {getLottieMetadata} from './get-lottie-metadata';
+export {Lottie, LottieProps} from './Lottie';
```
arthurdenner:
```suggestion
export {Lottie} from './Lottie';
export type {LottieProps} from './Lottie';
```

packages/lottie/src/Lottie.tsx
```diff
@@ -0,0 +1,108 @@
...
+import lottie from 'lottie-web';
+import type {CSSProperties} from 'react';
+import {useEffect, useRef, useState} from 'react';
+import {continueRender, delayRender, useCurrentFrame} from 'remotion';
+import type {LottieAnimationData} from './types';
+import {getNextFrame} from './utils';
+import {validateLoop} from './validate-loop';
+import {validatePlaybackRate} from './validate-playbackrate';
+
+export interface LottieProps {
```
arthurdenner, line 11: Are types preferred in the Remotion codebase?

arthurdenner (commented):
@JonnyBurger, just a few comments and questions to address and I think we're good to go! 🚀

packages/lottie/src/Lottie.tsx
JonnyBurger: Thanks! I pointed it to the Lottie example snippet and added links for all 3 ways to do it

packages/lottie/src/Lottie.tsx
JonnyBurger: It doesn't really matter IMO! So this is fine for me

packages/lottie/src/index.ts
JonnyBurger: 🤝

packages/example/src/Lottie/Halloween/Balloons.tsx
JonnyBurger: Yes good idea - `staticFile` has some benefits which we'd like to promote over importing directly

JonnyBurger: Implemented that!

packages/docs/docs/lottie/getlottiemetadata.md
JonnyBurger: Hmm, it needs a `pnpm i` from the root again even though this is a local dependency and depending on if caching went wrong it might also need a `pnpm run clean` from the root

packages/example/src/Lottie/Cybertruck/index.tsx
JonnyBurger: It should always work if you call it directly because Remotion also waits for the JavaScript event loop to complete. I updated it!

PR #1209, opened 2022-08-09
remove mode="continuous" from skia because not needed anymore

JonnyBurger:
Remove the mode="continuous" but require at least 0.1.138 (not released at this point)


wcandillon:
it doesn't look like I'm allowed to edit this file but I tried 139 on Remotion and it's kind of a change changer:
* No more continuous rendering
* Support retina display (this one is embarrassing 😅)
* Fix issue where we would never unmount any components within a Canvas 🫣

JonnyBurger:
Let's ship it

PR #1217, opened 2022-08-10
feat(lottie): add support for animation direction

arthurdenner:
Seems to be enough to subtract the computed frame from the total frames for the correct result.

1. The values expected on [`setDirection`](https://github.com/airbnb/lottie-web#setdirectiondirection) are `1` and `-1`. Should we use the same or rename them to `forward` and `reverse` instead?

2. Should we rename the property to something other than `direction`? We renamed `speed` to `playbackRate` for example, hence the question.


JonnyBurger (changes requested):
Thanks a lot! 🙌

I think "forwards" and "backwards" would be nicer. We can make this a TypeScript union type `"forwards" | "backwards"`.
It also needs a mention in the docs in the /docs/lottie/lottie page.

arthurdenner:
> I think "forwards" and "backwards" would be nicer.

Apologies for the nitpick comment but it seems like "forwards" and "backwards" are British English while "forward" and "backward" are American English. Usually code is on American English (e.g. color vs colour).

JonnyBurger:
That's nice attention to convention! US English is also fine of course!

JonnyBurger:
LGTM! Can I merge it?

arthurdenner:
> LGTM! Can I merge it?

Yep! 🚀

JonnyBurger (approved):

JonnyBurger:
Thanks a lot for adding this 🙌🏼

PR #1231, opened 2022-08-12
docs: add postprocessing example to visualizeAudio

marcusstenbeck:
This adds postprocessing example of how to match the values from [`WebAudio AnalyserNode`](https://developer.mozilla.org/en-US/docs/Web/API/AnalyserNode)


JonnyBurger (approved):
Very valuable! Thanks a lot 🙌

PR #1242, opened 2022-08-17
docs: add links to W3C spec

marcusstenbeck:
This just adds a link to the specific place in the W3C where the `getByteFrequencyData()` algorithm is explained.

### DIFF
--- a/packages/docs/docs/visualize-audio.md
+++ b/packages/docs/docs/visualize-audio.md
@@ -93,9 +93,10 @@ const params = {
 // ---cut---
 /**
  * This postprocessing step will match the values with what you'd
- * get from WebAudio's AnalyserNode.
+ * get from WebAudio's `AnalyserNode.getByteFrequencyData()`.
  *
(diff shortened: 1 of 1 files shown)


tomByrer:
Would link to MDN be better?  I find W3 specs hard to read.
https://developer.mozilla.org/en-US/docs/Web/API/AnalyserNode/getByteFrequencyData

marcusstenbeck:
@tomByrer I updated the MDN link to go to [`getByteFrequencyData() on MDN`](https://developer.mozilla.org/en-US/docs/Web/API/AnalyserNode/getByteFrequencyData). However, MDN doesn't have the information about the post-processing. So for anyone curious about where the post-processing algorithm comes from the information is only found in the W3C spec.

This is the specific part of how `getByteFrequencyData()` should process the raw FFT values before returning data to the user.

image

JonnyBurger:
🤝 Thanks a lot for the polish!

PR #1244, opened 2022-08-19
docs: multi-process rendering with localStorage caching requires the data API response to stay the same

marcusstenbeck:
You might want to word it differently, but someone on Discord had an issue that was caused by this.


JonnyBurger:
Thanks a lot for that!
I noticed that the page makes some false statements as well (data is not fetched every frame if you do it right), so I rewrote parts of it. The importance of idempotency has been added!

PR #1268, opened 2022-08-27
feat: self-contained component for embedding player in iframe

marcusstenbeck:
I updated the example with a self-contained component so there's no need to remember to add the classname.


JonnyBurger (approved):
🔥

PR #1270, opened 2022-08-28
docs: fix type issue by ignoring a type check

marcusstenbeck:
Fixes so typings work. If there's a better way let me know.
Also removes an unused variable and corrects a typo in the component name.


JonnyBurger:
I approve of the removal of playerRef, thanks for catching that... other than that the snippet does not cause any type errors for me!

image

JonnyBurger:
Nevermind, I see it now - I could also reproduce the type error when using the component. I managed to implement ref forwarding though, at least! closing in favor of 7a579d4e063ff0a42b68d9cfb9ffe0989ff62206 ca2b93d105575cf45b19306b6db26fcfab6732de, sorry for the messy commit process!

PR #1282, opened 2022-09-04
fix: variable name

marcusstenbeck:
`otherProps` does not exist, and because of the `@ts-expect-error` it slipped through. I changed it to `props`. :)


JonnyBurger:
Thanks for catching that! 🙌

PR #1285, opened 2022-09-04
AbsoluteFill: add ref forwarding

satelllte:
This is rarely used but I think it's still worth to fill this gap since the component is highly reusable

### DIFF
--- a/packages/core/src/AbsoluteFill.tsx
+++ b/packages/core/src/AbsoluteFill.tsx
@@ -1,15 +1,8 @@
-import React, {useMemo, forwardRef} from 'react';
+import React, {forwardRef, useMemo} from 'react';
 
-/**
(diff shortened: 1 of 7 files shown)


JonnyBurger (dismissed):
Super useful! In fact I wanted to do this yesterday as well and had to work around it!
Thanks a lot for implementing this 🙌

JonnyBurger:
Also implemented it for `<Sequence>` and added docs to it btw!

satelllte:
> Super useful! In fact I wanted to do this yesterday as well and had to work around it!
Thanks a lot for implementing this 🙌

> Also implemented it for <Sequence> and added docs to it btw!

Nice, thanks @JonnyBurger. Waiting for them both to be live 🚀 🛰️

PR #1291, opened 2022-09-05
Remove the second browserExecutable argument

B3nnjoe:
By some reason there's two "browserExecutable" arguments in the arguments section. One of them should not be there??


JonnyBurger:
Thanks a lot! Good catch!

PR #1327, opened 2022-09-19
getGifDurationInSeconds() + Fix GIF fill modes

JonnyBurger:


packages/example/src/GifTest/gif-duration.tsx
```diff
@@ -0,0 +1,34 @@
...
+	}, [giphy]);
+
+	return (
+		<AbsoluteFill
+			style={{
+				backgroundColor: 'white',
+				fontSize: 100,
+			}}
+		>
+			Check the console!
```
khanzzirfan, line 29: I would just add a local state and then reflect the state value here , instead of `check the console`

JonnyBurger: Thanks, just added that with another commit on top of main

PR #1331, opened 2022-09-20
Change architecture from example to use the default one

FelippeChemello:


JonnyBurger (approved):
Makes a lot of sense! Thanks!

PR #1333, opened 2022-09-22
automatic root directory

patsalv:
Testcase now uses automatically determined root in order to no longer have C:\ and D:\  clashes


JonnyBurger:
Should only use the custom root on Windows, also please remove log and tsconfig change!

JonnyBurger:
Thanks!

PR #1334, opened 2022-09-22
Install browser with codecs on Windows if not installed

patsalv:
if no fitting browser is installed, thorium gets loaded from s3 instead of chromium on windows


JonnyBurger:
Perfect, thanks!

JonnyBurger:
Thanks a lot, nice improvement

PR #1336, opened 2022-09-22
 move  gifuct-js package in @remotion/gif  from devDependencies to dependencies

gexin1:
When I delete all the packages and upgrade to version 3.2.22, the reinstallation depends on completion, and the startup project prompts that gifact js is missing


JonnyBurger:
Oops, super good catch!
Thanks a lot and another release should be out today!

PR #1351, opened 2022-09-29
Clean temporary image frames when rendering ends or fails

gexin1:
Clean temporary image frames when rendering ends or fails


JonnyBurger:
Sorry for the late review and thanks a lot for the PR! Great find 🙏

PR #1368, opened 2022-10-04
feat(env): load new env variable on update

### LINKED ISSUE #1363: Load new env variables if `.env` has changed
[Issuehunt badges](https://issuehunt.io/r/remotion-dev/remotion/issues/1363)

🕛 This issue is currently assigned to @Slashgear and not for taking

---

## Acceptance criteria

- If the environment variables changed while `npx remotion preview command` is running, inject them into the browser and reload the server
- Use a file watcher if `.env` only if the file exists.
- Only inject variables that start with `REMOTION_` (use `getProcessEnv()` to filter out acceptable)
- Use the `EventSource` logic (search `new-input-props` in the codebase) to push a notification to the client that the environment variables have changed and apply a full reload
- Figure out how to update the environment variables via Webpack. Maybe it is not possible, in that case you have to write some code that updates the variables on page load.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $164.00)

- [remotion-dev remotion-dev](https://issuehunt.io/u/remotion-dev) ($112.00)
- [jonnyburger jonnyburger](https://issuehunt.io/u/jonnyburger) ($52.00)

### Submitted pull Requests
- [#1368 feat(env): load new env variable on update](https://issuehunt.io/r/remotion-dev/remotion/pull/1368)
---

### Tips

</details>

Slashgear:
## Why

Close #1363

## How

- Add new event type for preview app
- Handle new behaviour for onUpdate for getProcessEnv
- Handle filewatcher on `.env` file

## Todo

- [x] I still need to check that everything works correctly but I think this would do the job.
- Do we need to update documentation about env variable ? 
- Do we need to add new tests on that ?

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#1363: Load new env variables if `.env` has changed](https://issuehunt.io/repos/274495425/issues/1363)
---
</details>


JonnyBurger (changes requested):
The reload when the file changes is beautifully done! Very nice!
The main point of the PR does not yet work though - when opening the preview and entering `process.env` in the console, the old environment variables are still present.

Let me know if you have an idea on how to solve this, or if I should help come up with something!

Slashgear:
So I need to flush the previous variable before reinjecting the new one.

- Run 1 variable A is defined
- User removes variable A and add a variable B
- Refresh of the preview app
- A is not defined anymore and B is defined

That what you mean ?

JonnyBurger:
@Slashgear Yes, that would be the desired behavior! Since we currently don't make use of environment variables in Node.JS 
 during the preview but just in the browser, it needs to be re-injected into Webpack or added into a <script> tag when loading `index.html`, this would be a relatively simple solution actually.

Another possibility: We currently use DefinePlugin by Webpack, and I found a way to dynamically change the values: https://webpack.js.org/plugins/define-plugin/#runtime-values-via-runtimevalue We could add a dependency on the `.env` file.

JonnyBurger:
My preferred solution would be if there is a refactor and the environment variables get added to `index-html.ts` like we do for inputProps, since then we use the same pattern for both and we don't rely on Webpack in case we want to support more bundlers in the future.

JonnyBurger (dismissed):

JonnyBurger:
Implemented the deluxe version and also added handling for adding / removing a .env file 😇

PR #1372, opened 2022-10-05
Rotato 3D style video

alwex:
thanks again for remotion, this is the video used on the app paywall, using a 3D rendered phone


packages/docs/src/data/showcase-videos.tsx
```diff
@@ -1245,6 +1245,26 @@ export const showcaseVideos: ShowcaseVideo[] = [
...
+	submittedOn: new Date("09-30-2022"),
+	links: [
+		{
+			type: "website",
+			url: "https://blastworkout.app",
+		},
+	],
+	author: {
+		"url": "https://blastworkout.app",
+		"name": "Mad Mustache Company>"
```
JonnyBurger: Is this deliberate that there is a `>` at the end?
I also saw it in the other PR and removed it there 🤔

alwex: Oups definitely not deliberate, thanks a lot for removing it in the previous PR, I'll pay more attention next time.

PR #1374, opened 2022-10-06
Ffmpeg autoinstall

patsalv:
Ffmpeg gets downloaded and installed automatically if not yet existing


.github/workflows/push.yml
```diff
@@ -18,11 +18,10 @@ jobs:
           node-version: ${{ matrix.node_version }}
       - run: |
           corepack enable
-      - uses: Iamshankhadeep/setup-ffmpeg@v1.1
+      - uses: JonnyBurger/setup-ffmpeg@main
```
JonnyBurger: Good for testing, we cannot forget to remove it before we merge

packages/renderer/src/ensure-presentation-timestamp.ts
```diff
@@ -55,7 +56,7 @@ export const ensurePresentationTimestamps = async (
 	// If there is no file extension for the video, then we need to tempoa
 	const output = await getTemporaryOutputName(src);
 
-	await execa('ffmpeg', [
+	await execa(await getExecutableFfmpeg(null), [
```
JonnyBurger: This was a bug before where the ffmpegExecutable option was not respected, would be nice if we fixed it in the same PR

packages/renderer/src/ffmpeg-flags.ts
```diff
@@ -57,3 +62,89 @@ export const getFfmpegVersion = async (options: {
...
+export const downloadFfmpeg = async (): Promise<void> => {
+	// implement callback instead
+	function onProgress(downloadedBytes: number, totalBytes: number) {
+		console.log(
+			'Downloading ffmpeg: ',
+			toMegabytes(downloadedBytes) + '/' + toMegabytes(totalBytes)
+		);
+	}
+
+	const os = require('os');
```
JonnyBurger: Convention: should be an import at the top of the file

packages/renderer/src/ffmpeg-flags.ts
```diff
@@ -57,3 +62,89 @@ export const getFfmpegVersion = async (options: {
...
+	const path = require('path');
+	const destinationPath =
+		os.platform() === 'win32'
+			? path.resolve(process.cwd(), 'node_modules/.ffmpeg/ffmpeg.exe')
+			: path.resolve(process.cwd(), 'node_modules/.ffmpeg/ffmpeg');
```
JonnyBurger: - Lot of redundancy here, would be better to make a function that returns the folder name and one that returns the absolute path, and then use it across the PR, for writing and reading the file
- More precise than process.cwd() would be `remotionRoot` which you should be able to access by passing it as an argument through the functions. `remotionRoot` is like process.cwd() but it handles the case that the user started remotion from a directory not adjacent to a package.json

packages/renderer/src/ffmpeg-flags.ts
```diff
@@ -57,3 +62,89 @@ export const getFfmpegVersion = async (options: {
...
+			'https://remotion-ffmpeg-binaries.s3.eu-central-1.amazonaws.com/ffmpeg-linux-amd64';
+	}
+
+	try {
+		const totalBytes = await _downloadFile(url, destinationPath, onProgress);
+		onProgress(totalBytes, totalBytes);
+		if (os.platform() !== 'win32') {
+			fs.chmodSync(destinationPath, '755');
+		}
+	} catch (error) {
```
JonnyBurger: Lets remove the try/catch and let the global error handler handle it

packages/renderer/src/ffmpeg-flags.ts
```diff
@@ -57,3 +62,89 @@ export const getFfmpegVersion = async (options: {
...
+		console.log(error);
+	}
+};
+
+const getFfmpegBinaryFromNodeModules = () => {
+	const os = require('os');
+	const isWin = os.platform() === 'win32';
+	const path = require('path');
+	if (isWin) {
```
JonnyBurger: redundancy that can be be eliminated together with all the other instances of this path

packages/renderer/src/ffmpeg-flags.ts
```diff
@@ -57,3 +62,89 @@ export const getFfmpegVersion = async (options: {
...
+
+// should check if ffmpeg is installed. If installed, return "ffmpeg" else return path to ffmpeg.exe in node modules
+export const getExecutableFfmpeg = async (
+	ffmpegExecutable: FfmpegExecutable | null
+) => {
+	const os = require('os');
+	const isWin = os.platform() === 'win32';
+	const path = require('path');
+	if (await binaryExists('ffmpeg', ffmpegExecutable)) {
+		return 'ffmpeg';
```
JonnyBurger: Bug: if a custom ffmpegExecutable is provided and it exists, we still return "ffmpeg"

packages/renderer/src/prespawn-ffmpeg.ts
```diff
@@ -128,7 +129,7 @@ export const prespawnFfmpeg = async (options: PreSticherOptions) => {
 		? options.ffmpegOverride({type: 'pre-stitcher', args: ffmpegString})
 		: ffmpegString;
 
-	const task = execa(options.ffmpegExecutable ?? 'ffmpeg', finalFfmpegString);
+	const task = execa(await getExecutableFfmpeg(null), finalFfmpegString);
```
JonnyBurger: Don't ignore options.ffmpegExecutable

packages/renderer/src/validate-ffmpeg.ts
```diff
@@ -35,15 +37,38 @@ export const binaryExists = async (
 	}
 };
 
+export const ffmpegInNodeModules = (): Promise<boolean> => {
+	const path = require('path');
+	const fs = require('fs');
```
JonnyBurger: Remove redundancies

packages/renderer/src/validate-ffmpeg.ts
```diff
@@ -35,15 +37,38 @@ export const binaryExists = async (
...
-	const ffmpegExists = await binaryExists('ffmpeg', customFfmpegBinary);
+	// binaryExists should also for node_modules
+	const ffmpegExists =
+		(await binaryExists('ffmpeg', customFfmpegBinary)) ||
+		(await ffmpegInNodeModules());
 	if (!ffmpegExists) {
+		console.log('Platform: ', os.platform());
+		if (
+			os.platform() === 'darwin' ||
+			os.platform() === 'win32' ||
```
JonnyBurger: Windows on ARM is also not supported, might lead to error on that platform

JonnyBurger (changes requested):

packages/renderer/src/merge-audio-track.ts
```diff
@@ -86,9 +94,12 @@ const mergeAudioTrackUnlimited = async ({
 	]
 		.filter(truthy)
 		.flat(2);
-
-	const task = execa(ffmpegExecutable ?? 'ffmpeg', args);
-
+	console.log('before task in merge-audio-tracks');
```
JonnyBurger: Remove console.logs that were added

packages/renderer/src/ensure-presentation-timestamp.ts
```diff
@@ -55,7 +57,7 @@ export const ensurePresentationTimestamps = async (
 	// If there is no file extension for the video, then we need to tempoa
 	const output = await getTemporaryOutputName(src);
 
-	await execa('ffmpeg', [
+	await execa(await getExecutableFfmpeg(null, remotionRoot), [
```
JonnyBurger: Should accept and respect the ffmpegExecutable option

packages/renderer/src/render-media.ts
```diff
@@ -382,37 +386,40 @@ export const renderMedia = ({
...
 			return Promise.all([
-				stitchFramesToVideo({
-					width: composition.width * (scale ?? 1),
-					height: composition.height * (scale ?? 1),
-					fps,
-					outputLocation: absoluteOutputLocation,
-					internalOptions: {
-						preEncodedFileLocation,
-						imageFormat,
+				stitchFramesToVideo(
```
JonnyBurger: I forgot to say: `stitchFramesToVideo` is also a public API and should therefore not accept the `remotionRoot` parameter (list of public APIs is here: https://www.remotion.dev/docs/renderer)

packages/renderer/src/validate-ffmpeg.ts
```diff
@@ -35,15 +37,34 @@ export const binaryExists = async (
...
+	customFfmpegBinary: string | null,
+	remotionRoot: string
 ): Promise<void> => {
-	const ffmpegExists = await binaryExists('ffmpeg', customFfmpegBinary);
+	// binaryExists should also for node_modules
+	const ffmpegExists =
+		(await binaryExists('ffmpeg', customFfmpegBinary)) ||
+		ffmpegInNodeModules(remotionRoot);
 	if (!ffmpegExists) {
+		console.log('Platform: ', os.platform());
```
JonnyBurger: Remove console.logs

packages/renderer/src/ffmpeg-flags.ts
patsalv: should be resolved with d403c4749c879707c67a960e4baddc26199b87b7

packages/renderer/src/ffmpeg-flags.ts
```diff
@@ -1,19 +1,29 @@
...
 
 export type FfmpegVersion = [number, number, number] | null;
 
-export const getFfmpegBuildInfo = async (options: {
-	ffmpegExecutable: string | null;
-}) => {
+export const getFfmpegBuildInfo = async (
+	options: {
+		ffmpegExecutable: string | null;
+	},
```
JonnyBurger: Better to either make an object with options, or multiple arguments rather than mixing them

JonnyBurger: (I realize now i did this myself 😂)

packages/renderer/src/ffmpeg-flags.ts
```diff
@@ -51,16 +126,178 @@ export const parseFfmpegVersion = (buildconf: string): FfmpegVersion => {
...
+			listeners[url] = [];
+		}
+
+		listeners[url].push((src) => resolve(src));
+	});
+};
+
+const onProgress = (downloadedBytes: number, totalBytesToDownload: number) => {
+	console.log(
+		'Downloading FFmpeg: ',
```
JonnyBurger: Or FFprobe

packages/renderer/src/ffmpeg-flags.ts
```diff
@@ -51,16 +126,178 @@ export const parseFfmpegVersion = (buildconf: string): FfmpegVersion => {
...
+	const inNodeMod = ffmpegInNodeModules(remotionRoot, 'ffprobe');
+
+	if (inNodeMod) {
+		return inNodeMod;
+	}
+
+	return downloadBinary(remotionRoot, url, 'ffprobe');
+};
+
+export const getExecutableFfmpeg = async (
```
JonnyBurger: This and getExecutableFfprobe has the exact same logic, so i'd propose to have only 1 function

packages/renderer/src/validate-ffmpeg.ts
```diff
@@ -1,22 +1,39 @@
...
+			statSync(localExecutable);
+			existsMap[name] = true;
+		} catch (err) {
+			existsMap[name] = false;
+		}
+
+		return existsMap[name];
+	}
+
+	if (name === 'ffprobe' && localExecutable) {
```
JonnyBurger: The logic for ffprobe and ffmpeg appears to be the same -> just 1 if

packages/renderer/src/validate-ffmpeg.ts
```diff
@@ -94,8 +141,6 @@ export const validateFfmpeg = async (
 			);
 		}
 
-		throw new Error('FFmpeg not found');
+		process.exit(1);
```
JonnyBurger: revert this (process.exit will quit the whole server that one might be running, without any chance to try/catch {} the error)

packages/renderer/src/ffmpeg-flags.ts
patsalv: fixed in 086616814f4f09589b3e8fc027f839604f5a3bd4 :)

packages/renderer/src/validate-ffmpeg.ts
patsalv: reverted in 8aadb05ee7734313947c7a4b3daa48b3cf9c0ebd

packages/renderer/src/validate-ffmpeg.ts
patsalv: resolved in 417c039192d45204f123b8d9179c4d09c2c60311

JonnyBurger:
Have to postpone it a bit because while I was double checking stuff I realized that it does break the Lambda integration.

JonnyBurger (approved):
Thanks 🙌

Issue #1247, opened 2022-08-21
Allow to pass FFMPEG bitrate parameters


JonnyBurger:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%24167%20Rewarded-%237E24E3.svg" />](https://issuehunt.io/r/remotion-dev/remotion/issues/1247)

**🎃 This issue is part of our Hacktoberfest campaign!**
📚 [Read more about Hacktoberfest here](https://remotion.dev/hacktoberfest)
🕛 This issue is currently assigned to @Slashgear and up for taking!

---

## Problem

User story from Band on Discord:

> I have a related use case. We are also publishing to Instagram via the ig-user/media API (https://developers.facebook.com/docs/instagram-api/reference/ig-user/media#reel-specifications). We’ve had a lot of flakiness with IG throwing vague errors for video uploads in the past, so we have validators that catch discrepancies between the IG specs and the video we are attempting to upload.
> While the test videos rendered by remotion seem to upload fine, The team is hesitant to remove any validators, since we can point to Instagram as being the source of the issues if we are matching their requirements. It would be convenient if I could add -ab 128 to the ffmpeg call so it passes our validation and everyone is happy.

## Hacktoberfest Acceptance criteria

- Allow to customize `-b:v` and `b:a` parameters using a new option for `stitchFramesToVideo()` and `renderMedia()`
- If using `renderMediaOnLambda()`, the option should also be passed
- If using the `npx remotion lambda render` or the `npx remotion render` commands, new command line flags should be introduced to use the feature from the CLI as well.
- All APIs that accept the parameter should have their API page updated.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $167.00)

- [<img src='https://avatars.githubusercontent.com/u/85344006?v=4' alt='remotion-dev' width=24 height=24> remotion-dev](https://issuehunt.io/u/remotion-dev) ($167.00)

### Submitted pull Requests
- [#1380 feat(ffmpeg): add possibility to override video output bitrate](https://issuehunt.io/r/remotion-dev/remotion/pull/1380)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/remotion-dev/remotion/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

gexin1:
[...]

PR #1380, opened 2022-10-07
feat(ffmpeg): add possibility to override video output bitrate

Slashgear:
## Why

Related to issue #1247

We want to let remotion users to be able to override the output bitrate for audio and video 
- in the CLI
- in lambda configuration
- in JS api

## How

Handle 2 new options for rendering videos with remotion
- `audioBitrate`
- `videoBitrate`

## Todo
- [x] Implementation of the feature
- [x] Tests
- [x] Documentation update


packages/renderer/src/stitch-frames-to-video.ts
```diff
@@ -349,7 +351,8 @@ export const spawnFfmpeg = async (
 		codec === 'h264' ? ['-movflags', 'faststart'] : null,
 		audioCodecName ? ['-c:a', audioCodecName] : null,
 		// Set max bitrate up to 1024kbps, will choose lower if that's too much
-		audioCodecName ? ['-b:a', '512K'] : null,
+		audioCodecName ? ['-b:a', options.audioBitrate || '512K'] : null,
+		options.videoBitrate ? ['-b:v', options.videoBitrate] : null,
```
JonnyBurger: We already have `-b:v` on line 349, the command should not be duplicated

packages/renderer/src/stitch-frames-to-video.ts
```diff
@@ -278,7 +280,7 @@ export const spawnFfmpeg = async (
 				audioCodecName,
 				// Set bitrate up to 320k, for aac it might effectively be lower
 				'-b:a',
-				'320k',
+				options.audioBitrate || '320k',
```
JonnyBurger: Convention in Remotion codebase: Use `??` operator

packages/lambda/src/api/render-media-on-lambda.ts
```diff
@@ -48,6 +48,8 @@ export type RenderMediaOnLambdaInput = {
 	downloadBehavior?: DownloadBehavior | null;
 	muted?: boolean;
 	overwrite?: boolean;
+	audioBitrate?: string | null;
+	videoBitrate?: string | null;
```
JonnyBurger, line 53: When we support it on Lambda, the bitrate should be passed in `functions/renderer.ts` as well.
It should *not* be passed to combineVideos(), because that function does not re-encode the video.

JonnyBurger (changes requested):

JonnyBurger:
I think this is a good approach!
We just need docs, a command line flag and an option for the config file.

Fine-tuning questions:
- Which codecs do support this option?
- Should we adopt FFMPEG notation (it's fine IMO) or accept a number which then gets transformed to FFMPEG notation?


FFMPEG from my experience is not precise with this option. You may signal a desired bitrate but FFMPEG might do something completely different. We should note this in the docs.

packages/renderer/src/stitch-frames-to-video.ts
Slashgear: Maybe I could add this rule in the linting configuration in a dedicate PR ? 💡

JonnyBurger:
Thanks a lot, reviewing tomorrow!

JonnyBurger (changes requested):
Works!

But I got some insights:

- [x] `--crf` overwrites `--video-bitrate`, so they should never be used together 
- [x] ProRes does not support `--video-bitrate` and also not `--crf`, this should throw an error
- [x] Same with audio codecs, they should reject `--video-bitrate` in `validateQualitySettings`
- [x] The documentation does not mention which syntax to use, e.g. `512K` or `1M`

Slashgear:
@JonnyBurger Ok I will work on that during the end of the week :)

packages/docs/docs/cli/render.md
```diff
@@ -59,6 +59,16 @@ Specify a location for a dotenv file. Default `.env`. Available in v2.2 and late
 
 [`h264` or `h265` or `png` or `vp8` or `vp9` or `mp3` or `aac` or `wav` or `prores` or `h264-mkv`](/docs/config#setcodec). If you don't supply `--codec`, it will use the H.264 encoder. Available since v1.4.
 
+### `--audio-bitrate`
+
+Specify the output bitrate for the generated audio.
+Beware, ffmpeg may, depending on the codec chosen, not take this parameter into account.
+
+### `--video-bitrate`
```
JonnyBurger, line 72: Should also be documented in renderMedia() and stitchFramesToVideo()

Slashgear:
>TODO: ProRes does not support --video-bitrate and also not --crf, this should throw an error
>TODO: Same with audio codecs, they should reject --video-bitrate in validateQualitySettings

Hi @JonnyBurger , not sure what to do with this ones. I am not sure to understand what are the cases I should cover.

Is the documentation update ok ?

JonnyBurger (approved):
Thanks a lot! Very excited for this feature to land! Appreciat the contribution! 🙌 💙

JonnyBurger:
@Slashgear I forgot to ask, do you remember how long you spent on this PR? (estimated)

Reviewing your other PR soon 😃

Issue #1300, opened 2022-09-08
Loop argument on Video, Audio and OffthreadVideo


B3nnjoe:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%24167%20Rewarded-%237E24E3.svg" />](https://issuehunt.io/r/remotion-dev/remotion/issues/1300)

**🎃 This issue is part of our Hacktoberfest campaign!**
📚 [Read more about Hacktoberfest here](https://remotion.dev/hacktoberfest)
🕛 This issue is currently assigned to @DerrykBoyd 

# Feature Request 🛍️
A loop argument on `<Video/>` and `<OffthreadVideo>`
Since now you need to use the `<Loop/>` and it needs to re-download the video for every iteration. (Correct me if im wrong.)

## Acceptance criteria

- Accept a prop `loop` on the `<Video>`, `<OffthreadVideo>` and `<Audio>` components.
- It should work for development and rendering versions of the component.
- Mount the video/audio tag as normal and once the video is ready, get it's duration using a ref.
- Once the duration is read, use recursion - if the loop property is there, it should render the same component again but without the loop property. For example: `<Video loop src="video.mp3" />` will become `<Loop><Video src="video.mp3" durationInFrames={xxx}/></Loop>`
- Don't use the APIs in `@remotion/media-utils` - the components in `remotion` should have no dependencies.
- It should also work if the `src` of a video/audio changes!
- It should use the `delayRender()` / `continueRender()` pattern to ensure that the duration is determined before the render starts.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $167.00)

- [<img src='https://avatars.githubusercontent.com/u/1629785?v=4' alt='jonnyburger' width=24 height=24> jonnyburger](https://issuehunt.io/u/jonnyburger) ($167.00)

### Submitted pull Requests
- [#1382 Add loop prop for Video and Audio components](https://issuehunt.io/r/remotion-dev/remotion/pull/1382)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/remotion-dev/remotion/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

JonnyBurger:
Thanks for filing, this is a valid issue!
It will not be slower if you use Loop though, Remotion will not download it multiple times. Still I think it would be a nicer API!

[...]

PR #1382, opened 2022-10-08
Add loop prop for Video and Audio components

DerrykBoyd:
closes: #1300 

@JonnyBurger sorry for the incomplete PR, I figured this might be the best way for some async communication and hopefully get me unstuck here.

After following the general steps in the issue, here is what the current code outputs in the preview, there is a looping video, but there is also a full length video that seems to render the last frame on top of the looping videos

image

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#1300: Loop argument on Video, Audio and OffthreadVideo](https://issuehunt.io/repos/274495425/issues/1300)
---
</details>


JonnyBurger (changes requested):
No problem, this is the way to go!

videoDuration is a ref, which means if it gets changed, the component will not re-render. I changed it to a state now.
Next problem was with the `ref`'s, there is a `videoRef` that should be passed to the native video element and a `ref` that should be passed to the `<Video>` component. When I fixed that, the time would not sync anymore so I realized a refactor needs to be done - at the very top of the Video component, we loop the whole `<Video>` component (not `<VideoForDevelopment>`) if we have the duration, which we get from a callback from the children.

It works now as intended! What's left to do is to apply the same logic to the audio code. For OffthreadVideo, I realized that there is no video tag during rendering from which we can get the duration, so my proposal is that for now, we don't support the loop property for OffthreadVideo.

Documentation is also needed :)

DerrykBoyd:
> No problem, this is the way to go!
> 
> videoDuration is a ref, which means if it gets changed, the component will not re-render. I changed it to a state now. Next problem was with the `ref`'s, there is a `videoRef` that should be passed to the native video element and a `ref` that should be passed to the `<Video>` component. When I fixed that, the time would not sync anymore so I realized a refactor needs to be done - at the very top of the Video component, we loop the whole `<Video>` component (not `<VideoForDevelopment>`) if we have the duration, which we get from a callback from the children.
> 
> It works now as intended! What's left to do is to apply the same logic to the audio code. For OffthreadVideo, I realized that there is no video tag during rendering from which we can get the duration, so my proposal is that for now, we don't support the loop property for OffthreadVideo.
> 
> Documentation is also needed :)

Thanks for this @JonnyBurger, super helpful. 

I added similar logic for Audio. I did run into one issue where looped audio failed to play if the duration of the audio file was longer than the sequence duration. I added the `maxDuration` prop as a fix, but let me know if I should do that another way.

I see the tests are failing, not quite sure why, can you give me some help there?

JonnyBurger:
Thanks a lot @DerrykBoyd! It looks all good to me.

I'm ready to merge, but the thing about there being an error if the audio duration is longer than the composition did not make sense to me and I could not reproduce it, even when I removed the maxDuration logic. Could you elaborate or give me the green light if you also cannot reproduce the error anymore?

Also for statistics purposes, we are asking everybody: How long do you estimate you spent on this issue?

Cheers 🎉

JonnyBurger:
Thanks a lot for pointing it out, I found another two edge cases because of this:

- `loop` should be applied before `startFrom`, otherwise it muted that example
- When the duration is fetched, it should be stored globally so when it gets unmounted and mounted again, we do not have to recalculate it

The warning

> Could not play audio due to following error: DOMException: The play() request was interrupted because the media was removed from the document

can now be ignored, if the duration is calculated we do a re-render, yes, but it plays the audio now.

DerrykBoyd:
Thanks for helping me through this one @JonnyBurger, a few things were a bit of stretch as I'm still getting to grips with the internals of the package. I learned a bit though!

JonnyBurger (approved):
These components are super fragile, you did a good job! Thanks a lot 💙 🙌
Shipping 🚀

Issue #1389, opened 2022-10-11
`<Thumbnail>` component


JonnyBurger:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%24223%20Rewarded-%237E24E3.svg" />](https://issuehunt.io/r/remotion-dev/remotion/issues/1389)

**🎃 This issue is part of our Hacktoberfest campaign!**
📚 [Read more about Hacktoberfest here](https://remotion.dev/hacktoberfest)
⏲️ This issue is currently assigned to @Slashgear! 
💰 Thanks to [CodeChem](https://codechem.com) for sponsoring this issue!

Currently, the `@remotion/player` package has a `<Player>` component for playing a video on the web.
A common task is to display a thumbnail of a video somewhere on the page, and to accomplish that, one could render a Player and disable all interactivity and freeze it.

I propose a new `<Thumbnail>` component that gets shipped in the Player package that only displays a still image.
The implementation could be similar to the Player, but without controls. Also logic from the`<Thumbnail>` component from the preview could be used: https://github.com/remotion-dev/remotion/blob/b579cc25eea72d1123f7de6bd75bd2605b579bbc/packages/cli/src/editor/components/Thumbnail.tsx

## Acceptance criteria

- Implement a new `<Thumbnail>` component that fits the description above and export it in the `<Player>` package
- Should have  at least props for dimensions and `style` like for `<Player>`
- Write a documentation page describing all the props and an example snippet.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $223.00)

- [<img src='https://avatars.githubusercontent.com/u/85344006?v=4' alt='remotion-dev' width=24 height=24> remotion-dev](https://issuehunt.io/u/remotion-dev) ($223.00)

### Submitted pull Requests
- [#1405 feat(player): implement Thumbnail component for player](https://issuehunt.io/r/remotion-dev/remotion/pull/1405)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/remotion-dev/remotion/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

Slashgear:
Hi @JonnyBurger Could I take this one after the on that I almost finished.
I already work on few rendering sujet but working on remotion component is really something I want to try :)

JonnyBurger:
[...]

PR #1405, opened 2022-10-16
feat(player): implement Thumbnail component for player

Slashgear:
## Why

#1389 

## How

- add Thumbnail component
- add page of documentation in API, is this ok ? 

## Todo

- [x] documentation update
- [x] example in player-example package

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#1389: `<Thumbnail>` component](https://issuehunt.io/repos/274495425/issues/1389)
---
</details>


Slashgear:
No worry @JonnyBurger ! 

Answering those question would help me a lot on figuring if I am in the correct way of integrating it.

JonnyBurger:
Answering them on my phone, in case I can unblock you :) 

- Do I need to add it to the player example page ?
> It would be nice to have 1 example of it!

- Do Thumbnail have to handle a composition or a component like the player ?
> Just a component like the Player!

- How could I share logic between Player and Thumbnail? Could I create composable between them?
> I imagine with sharing math functions and contexts

- Is their is way to force the frame context of a composition to be a dedicate frame index?
> Yes, see the implementation of the freeze component! Or maybe, you can use the Freeze component itself!

Slashgear:
@JonnyBurger I will work on that this weekend

packages/player-example/src/index.tsx
```diff
@@ -3,6 +3,7 @@ import {createRoot} from 'react-dom/client';
 import App from './App';
 import CarSlideshow from './CarSlideshow';
 import {VideoautoplayDemo} from './VideoAutoplay';
+import {Thumbnail} from "@remotion/player/dist/Thumbnail";
```
JonnyBurger: Should import directly from @remotion/player

packages/player/src/Thumbnail.tsx
```diff
@@ -0,0 +1,63 @@
...
+import type {TComposition, TimelineContextValue} from "remotion";
+import {random, Internals} from "remotion";
+
+export const Thumbnail:React.FC<{
+    composition: TComposition<unknown>;
+    targetHeight: number;
+    targetWidth: number;
+    frameToDisplay: number;
+    style?: CSSProperties
+}> = ({
```
JonnyBurger: Ideally the API is consistent with the one of the `<Player>` (props-wise).
The redundancy that is in this component is totally fine though, no need to create crazy abstractions 👍

Slashgear:
@JonnyBurger I try to render the composition from the component inside the thumbnail but I can't figure out why nothing renders. Maybe you have a lead on that ?

JonnyBurger:
@Slashgear The contexts must be setup very carefully. The `<Player>` component might help as a reference since it renders all the React context providers.
There was an error in the browser console that hinted me towards the contexts missing. I've pushed a commit that makes the component render now, although way too big, so `targetWidth` does not yet seem to do anything 😄

JonnyBurger (approved):
This is awesome 🥳 Very useful, and in fact I only just recently got asked about this feature! Glad I can tell them it's now there!

Appreciate the contribution so much 🙌 💙

How long do you think you spent on this issue? (and the other one?)

Slashgear:
This one, at least 3 hours @JonnyBurger

PR #1412, opened 2022-10-18
Move version checks for node and ffmpeg to the renderer package

### LINKED ISSUE #1359: Old Node/FFMPEG version should trigger an error even when rendering via Node.JS APIs
[Issuehunt badges](https://issuehunt.io/r/remotion-dev/remotion/issues/1359)

🕛 This issue is currently taken by @JRavi2!

---

## Hacktoberfest Acceptance Criteria

- Currently in the CLI we check for the FFMPEG version: https://github.com/remotion-dev/remotion/blob/33668831b0335fcf1ff5580e53046dad97ff6273/packages/cli/src/validate-ffmpeg-version.ts#L8
- And for the Node.JS version: https://github.com/remotion-dev/remotion/blob/33668831b0335fcf1ff5580e53046dad97ff6273/packages/cli/src/check-version.ts#L16
- These checks should be moved into the `renderer` package so the warnings also appear when using the Node.JS APIs.
- The Node.JS version should be immediately checked once the `@remotion/renderer` package is loaded.
- The FFMPEG version should be checked before `renderMedia()`, `renderFrames()`, `getCompositions()`, `stitchFramesToVideo()` gets called.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $112.00)

- [jonnyburger jonnyburger](https://issuehunt.io/u/jonnyburger) ($112.00)

### Submitted pull Requests
- [#1412 Move version checks for node and ffmpeg to the renderer package](https://issuehunt.io/r/remotion-dev/remotion/pull/1412)
---

### Tips

</details>

JRavi2:
Closes #1359 

I've added the checks for the versions to the `renderer` package, but got a few questions:
- Since there already existed a [validateFfmpeg()](https://github.com/remotion-dev/remotion/blob/main/packages/renderer/src/validate-ffmpeg.ts#L42) function in the `renderer` package, I've added the version check within that itself as that seemed like the right place for it. Is that the right approach or should it be separated out into a new file?
- The `renderMedia()` and `renderFrames()` functions are synchronous, so we can't await the call to `validateFfmpeg()`. Could that potentially cause any issues?

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#1359: Old Node/FFMPEG version should trigger an error even when rendering via Node.JS APIs](https://issuehunt.io/repos/274495425/issues/1359)
---
</details>


JonnyBurger (approved):
Thanks for the PR and great questions!

Although renderMedia() is asynchronous and we could add it to the promise chain, I figured we can afford to make validateFfmpeg synchronous.

I moved the check entirely into the renderer package, although the error might trigger later, it is simpler that way.

Appreciate the PR! 🙌

JRavi2:
@JonnyBurger Thanks!
> How long do you estimate you spent on it?

Approximately 1-2 hrs

PR #1417, opened 2022-10-19
WIP: improve player flickering

JonnyBurger:
Try to avoid player flickering while the canvassize gets calculated.

TODO:

- Try `aspectRatio` property


JonnyBurger:
🥳

PR #1434, opened 2022-10-22
Rename MotionBlur to Trail and add CameraMotionBlur component

marcusstenbeck:
~~MotionBlur works by making `n` copies of `children` and assigning each child an opacity to give an experienced trailing effect similar to motion blur.~~

~~This PR removes the `{children}` adjacent to those "onion skin" copies. The component should render only the "onion skin" layers, and not an extra original on top of everything.~~

This PR is waiting on the outcome of the discussion below to determine its future.

### DIFF
--- a/packages/motion-blur/src/MotionBlur.tsx
+++ b/packages/motion-blur/src/MotionBlur.tsx
@@ -72,7 +72,6 @@ export const MotionBlur: React.FC<MotionBlurProps> = ({
 					</AbsoluteFill>
 				);
 			})}
-			{children}
 		</AbsoluteFill>
 	);
 };


JonnyBurger:
I see why this is an useful option, but why do you suggest it as default? Wouldn't the user have to render their markup twice then, which might be cumbersome?

marcusstenbeck:
I dove deeper into this, and create a CodeSandbox with my findings.

https://codesandbox.io/s/remotion-blurring-techniques-5k5r4r?file=/src/MyComposition.tsx

image


### From left to right
**Original content**
Simply the original without any effect.

**The original component with a trail**
This component is equal to the `<MotionBlur/>` component from @remotion/motion-blur@3.2.35.

**Only trail**
Notice that only the end of the trail blends naturally with the background.

**Mimic natural camera motion blur** (similar to After Effects)
This is the closest to natural motion blur that occurs in a camera. Unfortunately the technique will degrade the visuals the more iterations are calculated. So... higher quality motion blur means throwing away color information.

image

The image will degrade as more iterations are added.
- 10 iterations: not very noticeable
- 20 iterations: image starts to degrade
- 50 iterations: massive amounts of bit depth is lost

This is because the color operations end up happening in 8-bit color space, and as we have more iterations we're more sensitive to precision errors.

> Note: Analog film cameras have a spinning device that prevents light from exposing analog film while the camera moves the film reel from a previous frame to the next frame. See: https://en.wikipedia.org/wiki/Rotary_disc_shutter

----------------

So what do I think we should do?

1. Rename the official `<MotionBlur />` component to `<Trail />` (only because that's what it actually does).
2. Add a component called `<CameraMotionBlur />` with the (sadly color-destructive) natural camera motion blur code.

marcusstenbeck:
Close in favor of #1443

PR #1437, opened 2022-10-24
simplex-noise@4.0.1

satelllte:
Small fix of `simplex-noise`'s TSDoc and return type of `createNoise4D` function, so consumers won't get confused in case of looking at `node_modules` for example.

Diff between 4.0.0 and 4.0.1: https://github.com/jwagner/simplex-noise.js/compare/4.0.0...4.0.1


JonnyBurger (approved):
Thanks a lot for keeping it tidy! 🙌

Issue #1377, opened 2022-10-06
If user does not pass a composition in the command line, show a list of compositions


JonnyBurger:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%240%20Funded-%2300A156.svg" />](https://issuehunt.io/r/remotion-dev/remotion/issues/1377)

**🎃 This issue is part of our Hacktoberfest campaign!**
📚 [Read more about Hacktoberfest here](https://remotion.dev/hacktoberfest)
✅ This issue is currently taken by @Uzunov-Stefan! 
💰 Thanks to [CodeChem](https://codechem.com) for sponsoring this issue!

There are several commands that take a composition:

```bash
npx remotion render src/index.tsx [composition-name]
npx remotion still src/index.tsx [composition-name]
npx remotion benchmark src/index.tsx [composition-name]
```

The goal is that one can omit the composition name:

```bash
npx remotion render src/index.tsx
npx remotion still src/index.tsx
npx remotion benchmark src/index.tsx
```

## Acceptance criteria

- Create a generic solution that asks the user what compositions they want to render when they don't specify the composition name
- The logic from here can be copy-pasted and the `prompts` module or another lightweight module may be used: https://github.com/uragirii/remotion/blob/6323e86f076f6b7979a1beb8c1bcf92964e583e8/packages/create-video/src/prompts.ts#L3
- Don't ask and fail if running on a CI `process.env.CI`
- Use getCompositions() to get the list of compositions (afaik this is already done)
- If there is only 1 compositions, choose that one
- Edge case handling: Normally you can specify an output name `npx remotion render src/index.tsx [composition-name] [out-name]`. If you don't include the composition name, you cannot set the out name either and the default will be used.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/remotion-dev/remotion/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

clearlysid:
oh yes please, this feature will be quite a timesaver for me. @JonnyBurger I'm happy to help fund this too, if you'd like.

JonnyBurger:
@clearlysid Thanks a lot! No need, we got you 😁

Uzunov-Stefan:
Hi @JonnyBurger 
I would like to work on this issue

[...]

Issue #1433, opened 2022-10-22
Write article about how to create a series of videos from a dataset


JonnyBurger:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%24195%20Rewarded-%237E24E3.svg" />](https://issuehunt.io/r/remotion-dev/remotion/issues/1433)

**🎃 This issue is part of our Hacktoberfest campaign!**
📚 [Read more about Hacktoberfest here](https://remotion.dev/hacktoberfest)
⏲️ This issue is currently assigned to @ThePerfectSystem!
💰 Thanks to [CodeChem](https://codechem.com) for sponsoring this issue!

A common problem that users want to solve with Remotion is that they have a dataset (let's assume JSON for simplicity) and create a video for each one of the entries in their set.

How I would recommend to do it is to create a TypeScript script, and use `renderMedia()` to render a video for each data and parametrize the video based on [input props](https://www.remotion.dev/docs/parametrized-rendering). There is already https://remotion.dev/docs/ssr which is similar but does not specifically explain how to approach the problem if you have a dataset.

## Acceptance criteria

- Write an article and add it to the website (take for example the Figma article as a reference: #1378)
- In that article, assume that you have a simple example video (it can for example be the example from https://www.remotion.dev/docs/player, source can be found at the bottom of the page) with few parameters (e.g. color and name)
- Explain step for step, how one would write and run a script that renders a video for each person in a dataset. Also create a small sample dataset, for example of 10 people who each have a name and favorite color.
- Include code snippets and screenshots where it makes sense.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $195.00)

- [<img src='https://avatars.githubusercontent.com/u/85344006?v=4' alt='remotion-dev' width=24 height=24> remotion-dev](https://issuehunt.io/u/remotion-dev) ($195.00)

### Submitted pull Requests
- [#1503 add documentation for rendering from dataset](https://issuehunt.io/r/remotion-dev/remotion/pull/1503)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/remotion-dev/remotion/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

flynnzler:
Hello! Can I work on accomplishing this?

JonnyBurger:
[...]

PR #1438, opened 2022-10-24
Issue 1377 show list of compositions if none are passed

Uzunov-Stefan:
Issue #1377  

Added composition selection as well as multiple selections when no composition is passed by the user, for the render, still, and benchmark methods. 

These changes make it easier to choose compositions to render, benchmark, or use the still method.

I created a custom method that is executed in the `getComposiotionId()` for render and still method if there is no `<composition-id>` passed, and is directly called in the benchmark method. By default, the method opens a prompt that is of type `select`, but in benchmark, I pass a boolean value to the method which builds a prompt that is of type `multi-select`.

Render:
image

Still:
image

Benchmark:
image


JonnyBurger (approved):
Thanks, this is a huge improvement for Remotion! 😁 And it works very well, I refactored it just a bit and added support for Lambda!

Huge thanks! How long do you estimate that you spent on this issue?

JonnyBurger:
And also thanks for waiving the bounty, I have reassigned it to https://github.com/remotion-dev/remotion/issues/1433 which will also be very cool :)

Uzunov-Stefan:
> Huge thanks! How long do you estimate that you spent on this issue?

I was working on it for around 1day. I made most of the changes on Saturday, and then slight changes on Sunday, but all in all around 1day

Issue #1426, opened 2022-10-21
Cannot navigate through menus with keyboard anymore


JonnyBurger:
We have a regression where the Enter key cannot select any menus anymore

PR #1447, opened 2022-10-27
feat(editor): add keydown handler to menus

arthurdenner:
A small improvement to the editor. Relates to #1426.


JonnyBurger (approved):
Thanks a lot for the PR, which made it very easy for me to dig into the problem! 😃

Unfortunately I had to get rid of most of the changes, because the Enter was already supposed to work naturally by making everything a button and putting a `tabIndex` attribute on it. I found the bug now, we were listening on the <kbd>Enter</kbd> key as an alternative to pause the video, and that was preventing the Enter key being used to trigger a button click, hence why the explicit keydown listener was not necessary.

Thanks nonetheless, I would not have fixed it so fast if it was not for your PR 😃

PR #1448, opened 2022-10-27
If entry point isn't specified, try to automatically detect it

### LINKED ISSUE #1391: Make entry point optional
[Issuehunt badges](https://issuehunt.io/r/remotion-dev/remotion/issues/1391)

🕛 This issue is currently assigned to @Pompette
💰 Thanks to [CodeChem](https://codechem.com) for sponsoring this issue!
⚠️ **This issue can only be taken once #1377 is finished!**

Each Remotion command starts with `npx remotion [command] [entry-point]`:

```
npx remotion preview src/index.tsx
npx remotion render src/index.tsx
npx remotion benchmark src/index.tsx
```

The goal is to make it that if you omit the entry point, Remotion will detect it itself:

Rules:
- First priority: If the entry point is specified, use it.
- Second priority: Add a new rule in the config file: `Config.Preview.setEntryPoint('src/index.tsx')` (subject to change) and check if it is set. If so, use it
- Third priority: Look for common paths like `src/index.tsx`, `src/index.ts`, `src/index.js`, `remotion/index.js`. If that is the case, use `Log.verbose` to print out the path that Remotion has decided on
- Fourth priority: Fail because entry point could not be determined.

## Acceptance criteria
- Implement logic according to rules above
- Update the CLI section of the documentation to reflect that you don't have to pass the entry point anymore and document the rules of order.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $167.00)

- [remotion-dev remotion-dev](https://issuehunt.io/u/remotion-dev) ($167.00)

### Submitted pull Requests
- [#1448 If entry point isn't specified, try to automatically detect it](https://issuehunt.io/r/remotion-dev/remotion/pull/1448)
---

### Tips

</details>

JRavi2:
Closes #1391

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#1391: Make entry point optional](https://issuehunt.io/repos/274495425/issues/1391)
---
</details>


packages/cli/src/entry-point.ts
```diff
@@ -0,0 +1,24 @@
...
+import { Log } from "./log";
+
+export const getEntryPoint = () => {
+	let file: string | null = parsedCli._[1];
+  if (file) return file;
+
+  file = ConfigInternals.getEntryPoint();
+  if (file) return file;
+
+  if (existsSync("./src/index.tsx")) file = "src/index.tsx";
```
arthurdenner: Correct me if I'm wrong but using the forward slash here could break on Windows.
Perhaps you need to create each path with something like:
`path.join(process.cwd(), 'src', 'index.tsx')`

arthurdenner (commented):
I was watching the related issue and saw the PR - thanks for working on it.
I just have one comment that may be relevant.

packages/cli/src/entry-point.ts
JRavi2: Ah, nice catch! Though I'm pretty sure that under the hood, `existsSync` at some point calls the `resolve()` function which should just take care of the platform specific paths?
Ref: https://github.com/nodejs/node/blob/main/lib/fs.js#L299

JonnyBurger:
Thanks a lot! It works well 🥳

I did some rearchitecture myself that made sense to do at the same time and managed to get some improvements out of it, so that you can omit the entry point but pass a composition name, e.g. `npx remotion render my-comp`.

Gonna do some smoke testing to make sure nothing did break, then gonna merge it!

packages/cli/src/entry-point.ts
JonnyBurger: Good point! I've replaced it with `path.join("src", "index.tsx")` which should give us a guarantee that it's going to work :)

JonnyBurger:
Beautiful! 🤩
Thanks a lot again! 💙🙌🏼

For statistics: How long do you think you spent on the issue?

JRavi2:
Thanks @JonnyBurger! I spent about an hour or two on this.

PR #1453, opened 2022-11-01
fix: `hot reload` instead of `fast refresh`

Just-Moh-it:
Fast refresh is a vague term. It's officially called hot reloading 🙃


JonnyBurger:
Seems legit 👍 Thanks!

PR #1458, opened 2022-11-02
lottie add frame overload value

Fenki404:
This PR add a optional value _**animationFrame**_ in the props of Lottie to override the frame calculation.
So the playback can be maipulated from the outside of Lottie.

### DIFF
--- a/packages/lottie/src/Lottie.tsx
+++ b/packages/lottie/src/Lottie.tsx
@@ -13,6 +13,7 @@ export const Lottie = ({
 	direction,
 	loop,
 	playbackRate,
+	animationFrame,
 	style,
 	onAnimationLoaded,
 }: LottieProps) => {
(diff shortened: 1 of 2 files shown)


JonnyBurger:
Hi Christian,

thanks a lot for the PR, appreciate it a lot! 
In that case, I'm thinking that it is not required to add this prop, because you can already achieve the same by wrapping the component in `<Freeze>`:

```tsx
import {Freeze} from "remotion";
import {Lottie} from "@remotion/lottie"

const MyComp: React.FC = () => {
	return (
		<Freeze frame={50}>
			<Lottie />
		</Freeze>
	);
}
```

By adding this prop, we would offer two ways of doing one thing, which is not bad in itself, but will make teaching Remotion a bit harder. So I am sorry to decline the PR, and hope that you get to solve your problem nonetheless!

PR #1466, opened 2022-11-07
docs: remove misleading info

marcusstenbeck:
Since *any* global styles from the enclosing webpage will be applied it's better to not mention any CSS frameworks. Anyone trying to embed the `<Player />` component into a website will experience "style leaking". Whether this is an issue or not is up to the person in question.


JonnyBurger (approved):
I agree! Thanks 🙌

PR #1481, opened 2022-11-12
[FIX][Expert page] Social link occupies full width

satelllte:
Social link occupies full-width, so users can click it even if it's hovered way far from the right side

image

This change fixes it ⚙️


JonnyBurger:
Thanks 🙌🏼

PR #1488, opened 2022-11-16
Handle downloads getting stuck by throwing a timeout error

JonnyBurger:

### DIFF
--- a/packages/renderer/src/assets/download-file.ts
+++ b/packages/renderer/src/assets/download-file.ts
@@ -2,6 +2,8 @@ import {createWriteStream} from 'fs';
 import {ensureOutputDirectory} from '../ensure-output-directory';
 import {readFile} from './read-file';
 
+type Response = {sizeInBytes: number; to: string};
+
 export const downloadFile = ({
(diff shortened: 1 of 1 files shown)


packages/renderer/src/assets/download-file.ts
```diff
@@ -17,13 +19,48 @@ export const downloadFile = ({
...
+				if (resolved) {
+					return;
+				}
+
+				rejectAndFlag(
+					new Error(
+						`Tried to download file ${url}, but no progress was made for 20 seconds`
+					)
+				);
+			}, 20000);
```
Sterv, line 59: @JonnyBurger  I'm running `remotion render` from the CLI and using a Composition that loads a really large video file (2.54GB). This timeout error occurs every time. I had to revert back to a version before this was in place in order to render the video locally. Can you enable us to extend the timeout to more than 20 seconds?

JonnyBurger: Thanks for reporting, turned it into an issue: #1570 Hoping to get this into the next version!

PR #1497, opened 2022-11-17
Remotion learn temporary blob url

patsalv:


JonnyBurger (approved):
Thanks! Very good, I made a few changes that I think make sense:

- Don't assume that S3 is being used
- Tried to simplify by removing obvious or redundant wording
- Not rendering the Player if there is no video, because it would just be empty

But the example makes a lot of sense and glad that we document it 🙌

PR #1505, opened 2022-11-22
doc: add entrypoint of lambda function in the faq

Grummfy:
Documentation about the lambda function entryoint

Feel free to adapt if required


JonnyBurger:
Thanks! We don't want to create confusion with the Remotion entry point (https://www.remotion.dev/docs/terminology#entry-point) which is something entirely different, so I will tweak it a bit.

Grummfy:
Indeed. That's a good point, thanks for upgrading it \o/

Issue #1475, opened 2022-11-10
Allow to style/add a ref to `<Series.Sequence>` and `<Loop>`


JonnyBurger:
Like a `<Sequence>`!

JonnyBurger:
And a `<Loop>`!

JonnyBurger:
Taken by @afzsalsayed96!

afzalsayed96:
Facing issues setting up Algora for UAE. Let me know if wire transfer or [wise.com](https://wise.com) would be feasible for you.
<img width="517" alt="Screenshot 2022-11-23 at 9 52 30 PM" src="https://user-images.githubusercontent.com/14029371/203598023-2cf10c09-118c-4f5d-8ec2-76146f02f271.png">

MehmetAdemi:
Hi Afzal, just checked with Algora and they said it should work now. Can you try it again?

afzalsayed96:
@MehmetAdemi Ran into another issue. Contacted folks from Algora again.
Meanwhile rechecking feasibility of direct transfer.

PR #1507, opened 2022-11-22
Expose `style` and `ref` in `Series.Sequence`

afzalsayed96:
Adresses #1475 

- [x] Similar to how you can now add a style to a <Sequence>: https://github.com/remotion-dev/remotion/blob/main/packages/core/src/Sequence.tsx, the same should be possible for <Series.Sequence>
- [ ]  and `<Loop>`

- [x] You should be able to add a ref to <Series.Sequence> using forwardRef, also analogue to how it is possible with <Sequence>

- [ ] Documentation should be added for both, mentioning the possibility in the same style as the documentation for <Sequence>

### DIFF
--- a/packages/core/src/Sequence.tsx
+++ b/packages/core/src/Sequence.tsx
@@ -24,7 +24,7 @@ export type SequenceContextType = {
 
 export const SequenceContext = createContext<SequenceContextType | null>(null);
 
-type LayoutAndStyle =
+export type LayoutAndStyle =
 	| {
 			layout: 'none';
 	  }
--- a/packages/core/src/series/index.tsx
+++ b/packages/core/src/series/index.tsx
@@ -1,6 +1,6 @@
 import type {FC, PropsWithChildren} from 'react';
-import {Children, useMemo} from 'react';
-import type {SequenceProps} from '../Sequence';
+import {Children, forwardRef, useMemo} from 'react';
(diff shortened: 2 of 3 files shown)


JonnyBurger:
Nice! Thank you 🙌🏼 Will review tomorrow if the remaining points are addressed!

packages/core/src/loop/index.tsx
```diff
@@ -53,10 +53,11 @@ export const Loop: React.FC<LoopProps> = ({
 						key={`loop-${i}`}
 						durationInFrames={durationInFrames}
 						from={i * durationInFrames}
-						layout={layout}
 						name={name}
 						showLoopTimesInTimeline={actualTimes}
 						showInTimeline={i === 0}
+						layout={props.layout}
+						{...(props.layout==='absolute-fill' && {style: props.style})}
```
afzalsayed96: while this is more explicit it can be changed to as follows based on your preference
```suggestion
						{...props}
```

JonnyBurger: Thanks! I like it but we also need to handle the case where layout is undefined (the default case where it defaults to `absolute-fill`). Fixed that now!

JonnyBurger:
Quality work! Thanks a lot, much appreciated! 🙌 LGTM!

We are trying out Algora, a new bounty platform, please claim our bounty over there and let us know if it works well!

PR #1517, opened 2022-11-28
fix non ascii characters in staticFile

### LINKED ISSUE #1514: staticfile & non-ascii characters
# Bug Report 🐛

When I use some path using non-ascii charactere like 'é' the static path is correctly generated, but not loaded.

Example
http://localhost:3000/static-778329f84875/remotion-lottie-resume/2-Journ%C3%A9e/02_Affiche_Match_Score.json
/var/www/src/resources/remotion-lottie-resume/2-Journ%C3%A9e/02_Affiche_Match_Score.json does not exist

so the conversion is properly made in the url (urlencoded), but the filepath seems not

JonnyBurger:
Fixes #1514


Grummfy:
thanks

Issue #1520, opened 2022-11-29
Use the "packageManager" field after scaffolding a template


JonnyBurger:
Scenario:

If you create a project based on a template with Yarn (as an example, but also with npm or pnpm), it should detect the current yarn version (`yarn -v`, let's say it yields `1.22.19`) and add `"packageManager": "yarn@1.22.19"` to the package.json.

This will prevent someone from accidentally using pnpm or npm in that project and also force everybody to use the same version, so hopefully fewer problems occur. (This only applies if the user has corepack enabled though, at least for now)

It can be implemented here: https://github.com/remotion-dev/remotion/blob/main/packages/create-video/src/patch-package-json.ts

alexfernandez803:
@MehmetAdemi can I take this?

alexfernandez803:
oh, all good @MehmetAdemi

RamiroMonte:
Hi, if someone can give me a hint on how this can be done, maybe I can get

MehmetAdemi:
Hi Ramiro, thanks for your interest! During Hacktoberfest, we try to give guidance and encourage new contributors whenever possible, but it's hard to give this guidance throughout the year. We assume that Jonny's description should be enough to solve it. Hope you don't mind. :)

mykeels:
Happy to take this if possible

JonnyBurger:
Good luck @mykeels! :)

PR #1530, opened 2022-12-01
feat: add packageManager to app scaffold #1520

mykeels:
#1520

# Implementation Strategy

The [selectPackageManager](https://github.com/remotion-dev/remotion/blob/main/packages/create-video/src/pkg-managers.ts#L30-L40) function already does a good job of telling us which package manager is responsible for running `npm init video`, so I had to 

1. Get the version of the package manager, which `child_process.exec` is great for.
2. Pass the package manager version to the `patchPackageJson` function, which passes it to the rendered package json.

## Tests

I noticed the create-video package does not have a `test` script in its package.json, but I wanted to add tests anyway for the `patchPackageJson` function, so I took the liberty of adding an abstraction for its `fs.readFileSync` and `fs.writeFileSync` dependencies. 

If this is not acceptable, I am happy to revert this.

## Results

When I run `create-video` locally, with the `Hello World` template, I get:

```json
{
  "name": "my-video",
  "version": "1.0.0",
  "description": "My Remotion video",
  "scripts": {
    // ...
  },
  "repository": {},
  "license": "UNLICENSED",
  "devDependencies": {
    // ...
  },
  "dependencies": {
    // ...
  },
  "packageManager": "npm@8.1.2"
}
```

If I have missed something, please let me know and I'll get to it.

### DIFF
--- a/packages/create-video/package.json
+++ b/packages/create-video/package.json
@@ -12,7 +12,8 @@
 	"scripts": {
 		"build": "tsc -d",
 		"watch": "tsc -w",
-		"lint": "eslint src --ext ts,tsx"
+		"lint": "eslint src --ext ts,tsx",
+		"test": "vitest --run"
 	},
 	"author": "",
 	"license": "SEE LICENSE IN LICENSE.md",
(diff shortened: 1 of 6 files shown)


JonnyBurger:
This is excellent! I love the typesafety and the added tests! Thank you so much! 🙌

Added a safeguard to still continue the installation just in case something goes wrong. If you run it in a repository which already has corepack enabled, the command might not be available 😂

mykeels:
Thanks for the quick review. 🙌

PR #1533, opened 2022-12-04
add expert

MehmetAdemi:


JonnyBurger:
Thanks! FYI: When adding new experts, you can now run `node render-cards.mjs` and Preview images will be generated that show up when sharing the link on social media.

PR #1542, opened 2022-12-05
Support getting credentials via AWS profile

alexey-pelykh:
Introduce support for `REMOTION_AWS_PROFILE` and `AWS_PROFILE` environment variables.


alexey-pelykh:
@JonnyBurger this PR addresses more complex use-cases when just having IAM user is not ok (e.g. security policies) and when SAML/Federated credentials are used. In this case, all permissions that should be directly attached to IAM user should go to role that can be further assumed by some subset of federated users.

JonnyBurger:
Many thanks! 🙌💙

PR #1544, opened 2022-12-06
iam:GetUser is not required

alexey-pelykh:
It's possible to use an equivalent of `aws sts get-caller-identity` to achieve the same goals.


alexey-pelykh:
@JonnyBurger would it be possible to share why Vercel is complaining? The logs are not public it seems

alexey-pelykh:
Thanks for the insight into why it failed! I took the liberty to squash the commits for easier management

JonnyBurger:
🔥 Super awesome! I did not know that, less permissions is always better! Thanks a lot for sending this!

PR #1552, opened 2022-12-08
preload asset as more generic preload method

jmaguirrei:
This PR is intended to be a more generic way to load assets (video, image, audio and font) because in my use case I have images that the preload feature does not allow.

If accepted, further changes may include:
* Refactor `preloadVideo` and `preloadAudio` methods to call `preloadAsset` internally instead.
* Automatically detect element type (by looking at it's extension) so `elemType` parameter could be optional.


jmaguirrei:
Hello @JonnyBurger I made this PR, but it seems to have errors. 
Sorry I am not much experienced with git, but let me know if this PR can be merged

JonnyBurger:
Good idea! Thanks a lot!

To stay coherent, I added `preloadFont()` and `preloadImage`, but all 4 now call your function under the hood.

PR #1553, opened 2022-12-08
renderStill() should also be able to return a Buffer

### LINKED ISSUE #1525: renderStill() should also be able to return a Buffer
As requested by @FelippeChemello 

One unfortunate design decision we made in renderMedia() was that we returned the raw buffer, not an object. This led to us having to implement the "slowest frames" API in an unfortunate way. Let's not make that mistake again

afzalsayed96:
Fixes #1525 

I'm trying to add a test case in `render-still.test.ts` as follows:

```js
test('Returns buffer in promise result', () => {
	return expect(
		renderStill({
			composition: {
				width: 1000,
				height: 1000,
				fps: 30,
				durationInFrames: 30,
				id: 'hithere',
				defaultProps: undefined,
			},
			frame: 0,
			serveUrl: 'https://gleaming-wisp-de5d2a.netlify.app/',
			downloadMap: makeDownloadMap(),
		})
	).resolves.toHaveProperty('buffer');
});
```

But I get this error:

```
Error: promise rejected "Error: Found no composition with the name hithere" instead of resolving
```

Any pointers how this could be solved?


afzalsayed96:
Update: Added a test case

JonnyBurger:
Perfect! Thanks a lot! 😊 🙏

PR #1554, opened 2022-12-09
Drop S3 Website option

alexey-pelykh:
The S3 Website option is not needed and is not really used since the default S3 endpoint (not the S3 Website endpoint) is being used everywhere. Also, the S3 Website endpoint does not support HTTPS protocol which is now required for the proxy to work on lambda due to the [Private Network Access](https://developer.chrome.com/blog/private-network-access-update/).


JonnyBurger:
Thanks a lot! This is very embarassing, I simply was not aware that we are not using the S3 website.
We also never printed the right endpoint, so I think we are safe in just removing this feature.

I also had no idea that the website feature was this cursed that endpoint is sometimes with a `-` and a `.`. And that it supports only HTTP 😅 which I think was the curlprit in the end.

Issue #1561, opened 2022-12-12
Allow to open project in VS Code or other editor from the preview


JonnyBurger:
Workflow:

I `cd` into a project, run `npm start`.
Then I realize I forgot to run `code .`, and I need to Ctrl+C, run it and then start the preview again.

Somewhere around here we could put a button "Open in VS Code", but of course it should be editor agnostic. The logic from the error overlay should be reused:

![Screenshot 2022-12-12 at 09 17 44](https://user-images.githubusercontent.com/1629785/206995442-da3c6dca-50bd-4513-a123-e7623ec522d6.png)

JonnyBurger:
Better idea: Put it as another item in the `File` menu, then it can also be invoked using `Cmd+K`!

mykeels:
How can an "Open in VS Code" button be editor-agnostic? Does it let you open in any editor of your choice?  Perhaps, a link to the error overlay logic will help?

mykeels:
Sounds great!

RohovDmytro:
I made a shallow search. 
- The editor package should be updated
- use-menu-structure.tsx contains buttons handlers
- launch editor inspiration from react team: 
https://github.com/facebook/create-react-app/blob/19fa58d527ae74f2b6baa0867463eea1d290f9a5/packages/react-dev-utils/launchEditor.js

I'm not planning to proceed further. It should be easy to open a specific file, not sure about opening a specific folder.

orimdominic:
I've started working on this

MehmetAdemi:
@RohovDmytro Thanks for the input!
@orimdominic Let's go! 🚀

uditkumar01:
Instead of showing a button, we can ask user to open the project in their preferred code editor, such as Visual Studio Code when the `npm start` command runs. This approach is much simpler and have a similar implementation [here](https://github.com/remotion-dev/remotion/blob/main/packages/create-video/src/open-in-editor.ts).

what do you think 🤔 ?

orimdominic:
I think that @uditkumar01 has a point.
If the button is on the UI, how many times would the user need to use it? Once the editor is opened, they will have no need to click on the button again.

JonnyBurger:
We already ask the user after they have initialized the project if they want to open it in their editor.
So this flow is intended for the all the times after the initialization of the project.

Allowing to do it via CLI is also a good idea, but it should never block the start of the preview server.
[...]

PR #1569, opened 2022-12-14
Open project in IDE

orimdominic:
This PR affects issue #1561 

I've been able to resolve this with a few caveats.

If no code editor is open, it will try to open it in VS Code by default. This happens because [guessEditor](https://github.com/remotion-dev/remotion/blob/a255b47c0daa632e4fbb49675853dff618a42788/packages/cli/src/preview-server/error-overlay/react-overlay/utils/open-in-editor.ts#L346) returns the currently open code editor and when there is none, it returns undefined

PS: I noticed that when Sublime Text is open on my computer and it should be the editor that opens up the project, I get the error in the screenshot below when I try to open the project.

Screenshot from 2022-12-14 11-47-14


packages/cli/src/editor/helpers/use-menu-structure.tsx
```diff
@@ -151,6 +151,34 @@ export const useMenuStructure = (closeMenu: () => void) => {
 						subMenu: null,
 						quickSwitcherLabel: 'New still...',
 					},
+					{
```
JonnyBurger: The action should only be displayed if `window.remotion_editorName` is not null

packages/cli/src/editor/helpers/use-menu-structure.tsx
```diff
@@ -151,6 +151,34 @@ export const useMenuStructure = (closeMenu: () => void) => {
...
+						id: 'open-in-editor',
+						value: 'open-in-editor',
+						label: 'Open in code editor...',
```
JonnyBurger: Can we rename it to `Open in ${window.remotion_editorName}`?
The `...` signals that a modal is opened in the Preview, so I would remove the `...`.

packages/cli/src/editor/helpers/use-menu-structure.tsx
```diff
@@ -151,6 +151,34 @@ export const useMenuStructure = (closeMenu: () => void) => {
...
+						onClick: async () => {
+							await fetch(`/api/open-in-editor`, {
```
JonnyBurger: Now duplicated with the code in `OpenInEditor`, can we create a helper function to be used in both places?

packages/cli/src/editor/helpers/use-menu-structure.tsx
```diff
@@ -151,6 +151,34 @@ export const useMenuStructure = (closeMenu: () => void) => {
...
+								body: JSON.stringify({
+									stack: {
+										originalFileName: `${window.remotion_cwd}`,
+										originalLineNumber: 1,
+										originalColumnNumber: 1,
+									},
+								}),
+							})
+							.catch(err => {
+								console.error(err)
```
JonnyBurger: You can handle the error with `notificationCenter.current?.addNotification()`.

packages/cli/src/preview-server/routes.ts
```diff
@@ -148,9 +148,10 @@ const handleOpenInEditor = async (
 		const stack = body.stack as SymbolicatedStackFrame;
 
 		const guess = await editorGuess;
+		const defaultCodeEditor = { process: 'code', command: 'code' }
 		const didOpen = await launchEditor({
 			colNumber: stack.originalColumnNumber as number,
-			editor: guess[0],
```
JonnyBurger: Let's not default to VS Code, instead if the editor was not guessed, let's not display the menu at all.

JonnyBurger (changes requested):
Thanks a lot! 🙌
It works, some polishing needed.

JonnyBurger:
How do you launch Sublime normally? Should we rename `sublime_text` to `sublime`?

JonnyBurger:
Alright, thanks!
And sorry for pushing a commit to this branch, was accidentially. Will be in main soon so the diff will be clean afterwards.

JonnyBurger:
Awesome! Looks great, thanks a lot! 🙌

@MehmetAdemi Can you reward the issue?¨

MehmetAdemi:
For sure! 
Thanks @orimdominic! Can you go to issue #1561 and sign up on Algora so I can make the payment?

JonnyBurger (approved):

orimdominic:
> For sure @JonnyBurger ! Thanks @orimdominic! Can you go to issue #1561 and sign up on Algora so I can make the payment?

I will soonest!

orimdominic:
Felt great contributing to this!

JonnyBurger:
Awesome! Your contribution is much appreciated 🙌 💙

PR #1574, opened 2022-12-15
fix typo

zavalit:
There is a typo in https://www.remotion.dev/docs/webpack#enable-support-for-glsl-imports

awesome name for a package `raw-roader` ;) but it should be `raw-loader` in this case


JonnyBurger:
Lol! 😂 Thanks for that! 🙌

PR #1582, opened 2022-12-15
Use URL time fragments for displaying the time in the video

### LINKED ISSUE #1572: Use URL time fragments for displaying the time in the video
See here: https://stackoverflow.com/questions/56428378/no-way-to-preload-first-video-frame-on-ios-safari
And the spec: https://www.w3.org/TR/media-frags/

We can append a fragment to the end of the video and it will get preloaded. `#t=0.5`

Moreso, we can signal when the video ends. We know the composition has a fixed duration, and therefore we can save resources by signaling to the browser it does not have to fetch the end of the video:

http://www.example.org/video.ogv#t=60,100

## Acceptance criteria

- Update the `VideoForDevelopment` component to, if there is no URL fragment specified, append a URL fragment to the URL to signal the expected portion of the video.
- The time must be specified in seconds and both start and end should be correctly calculated
- It should work for both raw URLs and URLs that have been prefetched().
- If the URL already has a fragment, nothing should be appended to the URL.
- Nothing should be done for Base64 URLs.

faisalsayed10:
Fixes #1572 
- [x] Update the VideoForDevelopment component to, if there is no URL fragment specified, append a URL fragment to the URL to signal the expected portion of the video.
- [x] The time must be specified in seconds and both start and end should be correctly calculated
- [x] It should work for both raw URLs and URLs that have been prefetched().
- [x] If the URL already has a fragment, nothing should be appended to the URL.
- [x] Nothing should be done for Base64 URLs.

### DIFF
--- a/packages/core/src/video/VideoForDevelopment.tsx
+++ b/packages/core/src/video/VideoForDevelopment.tsx
@@ -65,7 +65,7 @@ const VideoForDevelopmentRefForwardingFunction: React.ForwardRefRenderFunction<
 		mediaRef: videoRef,
 	});
 
-	const {shouldBeTime} = useMediaPlayback({
+	useMediaPlayback({
 		mediaRef: videoRef,
 		src,
 		mediaType: 'video',
@@ -74,6 +74,9 @@ const VideoForDevelopmentRefForwardingFunction: React.ForwardRefRenderFunction<
 		acceptableTimeshift: acceptableTimeshift ?? DEFAULT_ACCEPTABLE_TIMESHIFT,
 	});
 
+	const actualFrom = parentSequence
+		? parentSequence.relativeFrom + parentSequence.cumulatedFrom
+		: 0;
 	const duration = parentSequence
 		? Math.min(parentSequence.durationInFrames, durationInFrames)
 		: durationInFrames;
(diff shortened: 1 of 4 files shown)


JonnyBurger (changes requested):
Thanks for the super quick implementation!

I think the current code does not yet factor in all the cases. For example, this is a way to start a video only from 2 seconds on:

```
<Sequence from={-10}>
  <Video />
</Sequence>
```

The `startFrom` and `endAt` props actually also just convert into sequences. So the calculation needs to happen from the sequence contact. For reference, see the use-media-playback file: https://github.com/remotion-dev/remotion/blob/main/packages/core/src/use-media-playback.ts

faisalsayed10:
Thanks @JonnyBurger for the hint. How does this look?

packages/core/src/video/VideoForDevelopment.tsx
```diff
@@ -69,7 +74,22 @@ const VideoForDevelopmentRefForwardingFunction: React.ForwardRefRenderFunction<
...
-	const actualSrc = usePreload(src as string);
+	const duration = parentSequence
+		? Math.min(parentSequence.durationInFrames, durationInFrames)
+		: durationInFrames;
+
+	let actualSrc = usePreload(src as string);
+
+	const shouldAppendFragment = !actualSrc.startsWith('data:') && !actualSrc.includes('#t=')
+	if (shouldAppendFragment) {
+		if (typeof shouldBeTime === 'number' && Number.isFinite(shouldBeTime)) {
```
JonnyBurger: Hmm, now it looks like:

- instead of the start point of the video, it takes the current time
- instead of the end point of the video, it takes the duration of the sequence

I recommend taking a look at how the video layer gets displayed in the timeline: https://github.com/remotion-dev/remotion/blob/main/packages/core/src/use-media-in-timeline.ts

You can also start the preview in the `example` folder and open the DevTools to verify that the right fragment is calculated.

JonnyBurger (changes requested):

faisalsayed10:
Thanks @JonnyBurger,

I've updated the correct starting point calculation for the url fragment.

I'm not sure if I understand the requirements correctly for the second part. I'm noting few of my observations:

```
<Video src={framer} startFrom={30} endAt={60} />
```

The above code generates `#t=1,2`

and

```
<Video src={framer} endAt={60} />
```

This generated `#t=0,2`

Could you provide me a test-case and the desired fragment string that is expected?

JonnyBurger:
New code looks correct to me! Thanks a lot! 🙌 💙

JonnyBurger (approved):

faisalsayed10:
Thanks a lot for bearing with me on this 🙌

PR #1625, opened 2022-12-28
Update api.md

UmungoBungo:
I luckily stumbled across the helpful info around creating an external component to display a timeline, even though I had read these API docs multiple times.


JonnyBurger:
Thanks a lot for helping to connect the dots!

PR #1632, opened 2023-01-02
docs: add `deploySite` webpack override example

Just-Moh-it:
Added 'webpackOverride' example for `deploySite`. 

People might have to figure this out themselves, and since the title says 

> When using `bundle()` and `deploySite()`

I believe an example for both would be expected :)


JonnyBurger:
Nice addition! Thanks a lot Mohit 💙

PR #1654, opened 2023-01-09
Fix some missed `await`s

akx:
These were found by PyCharm's code inspections.


akx:
Pretty sure these test failures are also flakes...

PR #1656, opened 2023-01-09
Fix "Changing the temporary directory" docs to actually have an effect on macOS

akx:
macOS sets `TMPDIR` by default, so setting `TEMP` has no effect.

Setting `TMPDIR` on non-Windows platforms is the priority, as evident from the code linked in the doc.


JonnyBurger:
Super important to know!
Thanks, I did not know that, I just looked how Node.JS handles it internally and did not think about it 🫥 Thanks for stepping in! 🫡

PR #1668, opened 2023-01-10
Extend concurrency property to accept percentage

### LINKED ISSUE #1652: Concurrency in percent
I think extending the API of the `concurrency`property in the `renderMedia()` function by a percentage value would be a great addition, for example `100%` using all CPU Threads avaibale on the machine.

JRavi2:
Closes #1652

### DIFF
--- a/packages/docs/docs/renderer/render-frames.md
+++ b/packages/docs/docs/renderer/render-frames.md
@@ -87,7 +87,7 @@ A `string` which must be either `png`, `jpeg` or `none`.
 
 _optional_
 
-A `number` specifying how many render processes should be started in parallel, a `string` specifying the percentage of the cpu threads to use, or `null` to let Remotion decide based on the CPU of the host machine. Default is half of the CPU threads available.
+A `number` specifying how many render processes should be started in parallel, a `string` specifying the percentage of the CPU threads to use, or `null` to let Remotion decide based on the CPU of the host machine. Default is half of the CPU threads available.
 
 ### ~~`parallelism?`~~
 
--- a/packages/docs/docs/renderer/render-frames.md
+++ b/packages/docs/docs/renderer/render-frames.md
@@ -87,7 +87,7 @@ A `string` which must be either `png`, `jpeg` or `none`.
 
 _optional_
 
-A `number` specifying how many render processes should be started in parallel or `null` to let Remotion decide based on the CPU of the host machine. Default is half of the CPU threads available.
+A `number` specifying how many render processes should be started in parallel, a `string` specifying the percentage of the cpu threads to use, or `null` to let Remotion decide based on the CPU of the host machine. Default is half of the CPU threads available.
 
 ### ~~`parallelism?`~~
 
--- a/packages/docs/docs/renderer/render-media.md
+++ b/packages/docs/docs/renderer/render-media.md
@@ -74,7 +74,7 @@ An object of arbitrary shape that will be passed as [props to your composition](
 
 _optional_
 
-A `number` specifying how many render processes should be started in parallel, a `string` specifying the percentage of the cpu threads to use, or `null` to let Remotion decide based on the CPU of the host machine. Default is half of the CPU threads available.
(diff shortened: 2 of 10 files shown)


packages/docs/docs/renderer/render-frames.md
```diff
@@ -87,7 +87,7 @@ A `string` which must be either `png`, `jpeg` or `none`.
 
 _optional_
 
-A `number` specifying how many render processes should be started in parallel or `null` to let Remotion decide based on the CPU of the host machine. Default is half of the CPU threads available.
+A `number` specifying how many render processes should be started in parallel, a `string` specifying the percentage of the cpu threads to use, or `null` to let Remotion decide based on the CPU of the host machine. Default is half of the CPU threads available.
```
akx:
```suggestion
A `number` specifying how many render processes should be started in parallel, a `string` specifying the percentage of the CPU threads to use, or `null` to let Remotion decide based on the CPU of the host machine. Default is half of the CPU threads available.
```

packages/docs/docs/renderer/render-media.md
```diff
@@ -74,7 +74,7 @@ An object of arbitrary shape that will be passed as [props to your composition](
 
 _optional_
 
-A `number` specifying how many render processes should be started in parallel or `null` to let Remotion decide based on the CPU of the host machine. Default is half of the CPU threads available.
+A `number` specifying how many render processes should be started in parallel, a `string` specifying the percentage of the cpu threads to use, or `null` to let Remotion decide based on the CPU of the host machine. Default is half of the CPU threads available.
```
akx:
```suggestion
A `number` specifying how many render processes should be started in parallel, a `string` specifying the percentage of the CPU threads to use, or `null` to let Remotion decide based on the CPU of the host machine. Default is half of the CPU threads available.
```

packages/renderer/src/validate-concurrency.ts
```diff
@@ -7,23 +7,27 @@ export const validateConcurrency = (value: unknown, setting: string) => {
...
-		);
+		if (value > require('os').cpus().length) {
+			throw new Error(
+				`${setting} is set higher than the amount of CPU cores available. Available CPU cores: ${
+					require('os').cpus().length
+				}, value set: ${value}`
+			);
+		}
+	} else if(!(/^\d+(\.\d+)?%$/.test(value))) {
+		throw new Error(setting + ' must be a number or percentage, but is ' + value);
```
akx: This should probably use a template string too.

packages/renderer/src/validate-concurrency.ts
```diff
@@ -7,23 +7,27 @@ export const validateConcurrency = (value: unknown, setting: string) => {
...
-			}, value set: ${value}`
-		);
+		if (value > require('os').cpus().length) {
+			throw new Error(
+				`${setting} is set higher than the amount of CPU cores available. Available CPU cores: ${
+					require('os').cpus().length
+				}, value set: ${value}`
+			);
+		}
+	} else if(!(/^\d+(\.\d+)?%$/.test(value))) {
```
akx: This allows a floating point number, but you're using `parseInt` elsewhere. Will that work?

akx (changes requested):

packages/renderer/src/validate-concurrency.ts
JonnyBurger: Seems like it will round down the number then. I'm fine with that behavior

JonnyBurger:
And thanks for the contribution 🙌  will pay out the bounty today.

JRavi2:
@JonnyBurger in `getActualConcurrency` the calculated value is [floored](https://github.com/remotion-dev/remotion/pull/1668/files#diff-499d509d3834565a806420e611c780d6e649a227ea458bf1ba22c59484baeb9fR16), hence the failing test? Should it be rounded instead?

JonnyBurger:
@JRavi2 floor seems better! I changed the tests

JRavi2:
Changed a test to rounded for the `null` case, since that was how it was being handled earlier

JonnyBurger:
Awesome, thanks a lot!

Issue #1699, opened 2023-01-18
When renderMediaOnLambda from lambda function we still need have AWS Access Id and Secret


alexfernandez803:
When renderMediaOnLambda from lambda function we still need have AWS Access Id and Secret.  I propose to fix it by providing the lambda role access to invoke 'renderMediaOnLambda', by obtaining temporary credentials(https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_request.html). Aside from this, I'll be creating a writeup for the step by step process.

PR #1702, opened 2023-01-18
Fix issue UnrecognizedClientException 

alexfernandez803:
Part 1 fix for issue, https://github.com/remotion-dev/remotion/issues/1699.
When using renderMediaOnLambda with appropriate "permission" required by remotion, still throws 'UnrecognizedClientException'. The fix adds the session token that is paired with AWS temporary client and secret.
https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_request.html

With this fix, we can assign the 'remotion policy statements' directly to lambda.


JonnyBurger:
Thanks! Needs a documentation update as well.

PR #1710, opened 2023-01-20
fix: add back word lambda for command

Carson12345:
or else it will throw "command sites not found" error


JonnyBurger:
Whoops.. thanks for catching that! 👍🏼

PR #1726, opened 2023-01-23
Shapes star

alexfernandez803:
Just the basic star, I still need to understand how the star should fit in the instructions implementation.


JonnyBurger:
Super awesome! I also integrated rounding the edges and it can create some pretty cool and unexpected shapes!

Thanks a lot for this contribution 🙌

PR #1727, opened 2023-01-23
documentation fix: webhook signature sha1 -> sha512

Grummfy:
simple fix


JonnyBurger:
Makes sense! Thanks a lot!

Issue #1681, opened 2023-01-14
Ease Player controls gradient


JonnyBurger:
Like https://twitter.com/jaffathecake/status/1612766262064455681?s=46&t=d92JEcELcBvrKmSiu_prJw

CruuzAzul:
Hi @JonnyBurger I also saw this little problem on the player, I can take a look to do like the example if you want ? 😉

CruuzAzul:
Would something like this be good? I've taken some of what was in the example and adapted some values 😊

<img width="1192" alt="image" src="https://user-images.githubusercontent.com/60877626/212567379-e6a35b13-e4c0-48a3-94fd-f1a402b4a334.png">

JonnyBurger:
@CruuzAzul Awesome, thanks for taking this on!
I think the gradient is now smoother, nice! But it is very dark, how about making it a bit lighter?

CruuzAzul:
@JonnyBurger is it better this way ? 😊 If it's good I can make a pull request 🎫

<img width="1201" alt="image" src="https://user-images.githubusercontent.com/60877626/214242207-2a6a99cd-acfd-4f13-a1c8-a6b0de1519d3.png">

JonnyBurger:
Yes much better, still quite the difference in darkness! Is it a legibility concern?

Feel free to open a PR 😇

CruuzAzul:
I will open the PR to look at the final adjustments

JonnyBurger:
Fixed by #1742

PR #1742, opened 2023-01-25
fix(player): ease player controls gradient

CruuzAzul:
## 🤔 Why ?

Related to #1681

## 💻 How ?

Modification of the background on the PlayerControl. Don't know if there is a better solution to make it cleaner, don't hesitate to tell me 😊

### DIFF
--- a/packages/player/src/PlayerControls.tsx
+++ b/packages/player/src/PlayerControls.tsx
@@ -16,6 +16,18 @@ export type RenderFullscreenButton = (props: {
 export const X_SPACER = 10;
 export const X_PADDING = 12;
 
+const gradientSteps = [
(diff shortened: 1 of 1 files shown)


JonnyBurger:
Thanks a lot! Comparing to the previous one and looking at it closely, it indeed looks a lot smoother!

This PR also darkened the gradient a lot which I was a bit concerned about since we shouldn't make too many unexpected changes in minor versions, even if they are just visual.

Therefore I reduced the opaqueness by about 30% to make it closer to before. Hope that's alright for you! ✌️ 

Thanks a lot for the contribution and we are ordering some samples for hoodies soon, stay tuned!

PR #1765, opened 2023-01-30
update experts' network

MehmetAdemi:


packages/docs/src/data/experts.tsx
```diff
@@ -201,4 +201,52 @@ export const experts: Expert[] = [
...
+    email: "ray@blocklab.ch",
+    since: new Date("2023-01-30").getTime(),
+    description: (
+      <div>
+        I made:{" "}
+        <a target={"_blank"} href="https://www.romrif.com/">
+          Romrif.com
+        </a>
+        ! <br />
+        I am an experienced integration developer using Mulesoft, a backend
```
JonnyBurger: Isn't that the description of Alex?

JonnyBurger (changes requested):
Is that the description of Alex? Looks good otherwise!

packages/docs/src/data/experts.tsx
MehmetAdemi: Sorry, this is a mistake!

JonnyBurger (approved):

Issue #1764, opened 2023-01-30
Flat config file


JonnyBurger:
Currently, the config file works like this:

```ts
Config.Puppeteer.setChromiumOpenGlRenderer('angle');
```

I and others find it cumbersome to discover all the options by having to trigger TS autocompletion for every category.
It would be better if you can just do:

```ts
Config.setChromiumOpenGlRenderer('angle');
```

## Notes

- The old behavior should also be possible for now for backwards compatibility
- The documentation needs to be updated
- Note the config type is in `packages/core/src/config.ts` but the implementation is in `packages/cli/src/config/index.ts`.

## Contributors guide

https://remotion.dev/docs/contributing

alexfernandez803:
can I take this one?

JonnyBurger:
@alexfernandez803 Yes! If you want the SQS one also became available, you can choose one at a time.

PR #1769, opened 2023-02-01
New flat config file

alexfernandez803:
Fixes issue for https://github.com/remotion-dev/remotion/issues/1764


JonnyBurger:
Hey Alex, thanks a lot for the PR!

Sorry for the misunderstanding, but by this issue I meant to create a flat config file, and the OpenGL renderer option was just one example of that.

So the scope of the issue was meant to be to do this procedure to all options. Are you willing to do this to all options as well or did you underestimate the scope? You can also DM for re: bounty or I can help finish it if this is not what you expected.

JonnyBurger:
Awesome! Thanks a lot 🎉

Issue #1759, opened 2023-01-29
Bad legibility in dark mode


0xflotus:
At https://www.remotion.dev/docs/timeline#toggling-transparency-mode and https://www.remotion.dev/docs/timeline#in--out-markers the sections have partially bad legibility in dark mode.

Dark Mode:

<img width="999" alt="Bildschirmfoto 2023-01-29 um 23 40 04" src="https://user-images.githubusercontent.com/26602940/215359926-936977db-ed00-44c8-be89-7b8037e9a299.png">

Light Mode:

<img width="995" alt="Bildschirmfoto 2023-01-29 um 23 39 57" src="https://user-images.githubusercontent.com/26602940/215359931-da4b74aa-075a-48d4-b73c-7a1685450fe4.png">

CruuzAzul:
@JonnyBurger Do we need to also update the svg color to `currentcolor` in the **[Toggling transparency mode](https://www.remotion.dev/docs/timeline#toggling-transparency-mode)** section of the documentation ?

0xflotus:
I already did that. in #1777

PR #1777, opened 2023-02-02
fix: part of #1759

0xflotus:
fixes the missing part of #1759 

See: https://www.remotion.dev/docs/timeline#toggling-transparency-mode in dark mode


JonnyBurger:
Whoops, missed that! Thanks a lot 😁

PR #1785, opened 2023-02-04
updated tailwind import in config

jahamed:


JonnyBurger:
Thanks a lot for the PR!

Actually, the shorter format is the new one and you could get it if you upgraded Remotion.
But it is indeed confusing since that change was very recent, so I will for the moment add the old way too.

PR #1809, opened 2023-02-13
"Available from" badges

JonnyBurger:
Works!

image


JonnyBurger:
not yet everywhere, will adopt incrementally

Issue #1779, opened 2023-02-02
Confusing statement in docs


JonnyBurger:
![image](https://user-images.githubusercontent.com/1629785/216444810-17247a9b-3080-4834-92b6-a3fdd2255546.png)

CruuzAzul:
@JonnyBurger You just want to change the text that describes the code to make it clearer ? If so, I can change that 😊

JonnyBurger:
@CruuzAzul Yes, it is actually wrong because a `<Sequence>` by default is already positioned absolutely, so the layout would not shift anyway

JonnyBurger:
If you take this on, it is appreciated! 😊

CruuzAzul:
I take it 😉

JonnyBurger:
Closed by #1838

PR #1838, opened 2023-02-19
docs: change confusing statement in sequence documentation

CruuzAzul:
## 🤔 Why ?

Related to #1779

## 💻 How ?

Just a small correction in the documentation on a part that is a bit confusing, on the positioning of sequences what was said is actually wrong because a `<Sequence>` by default is already in `position: absolute`, so the layout would not shift anyway, I think that this part is no longer useful in the doc and that just with this we understand 😊

### DIFF
--- a/packages/docs/docs/sequences.md
+++ b/packages/docs/docs/sequences.md
@@ -70,8 +70,7 @@ export const MyVideo = () => {
 };
 ```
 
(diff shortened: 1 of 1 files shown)


JonnyBurger:
Thanks a lot! 🙌 
It happens often that people are surprised by the absolute positioning, so I added the opposite statement to hopefully reduce the amount of surprise.

Issue #1853, opened 2023-02-25
Disable zoom in remotion's previewer (possibly using a CLI flag)


Just-Moh-it:
# Feature Request 🛍️

Currently, pinching/scrolling into the video preview (that appears when you run `remotion preview ./index.tsx`), triggers a zoom. This was recently introduced. However, I've had my setup in such a way that I use `@react-three/drei`'s `<OrthographicCamera />` component, which adds pan/zoom controls to the react-three fiber scene. 

However, the `<OrthographicCamera />`'s gestures conflict with those of Remotion's previewer. So an option to disable zooming/panning on the Remotion previewer would be super useful.

## Use Case

As explained above

## Possible Solution

Have a CLI argument `remotion preview ./index.tsx --disable-zoom` or something similar

Just-Moh-it:
cc @JonnyBurger for visibility

JonnyBurger:
Sounds good! The whole thing can be disabled in https://github.com/remotion-dev/remotion/blob/4e838d228cf9e6ae42e5ebe9376543d805d9f334/packages/cli/src/editor/components/Canvas.tsx#L87

This can become an entry in the menu like for example the transparency toggle: https://github.com/remotion-dev/remotion/blob/4e838d228cf9e6ae42e5ebe9376543d805d9f334/packages/cli/src/editor/helpers/use-menu-structure.tsx#L310

Adding a bounty if someone wants to grab it!

/bounty 125

kaf-lamed-beyt:
I can take this, if you'd love me to @JonnyBurger

Just-Moh-it:
Hey @JonnyBurger, thanks for the prompt response! 

Was mostly free this weekend. How about I take this bounty? Would love to finally contribute to the source code of Remotion and not just the docs, haha 🙃

kaf-lamed-beyt:
> I can take this if you'd love me to @JonnyBurger

I think it'll be great if @Just-Moh-it takes this issue, @JonnyBurger. Since they opened it initially (and is interested in working on it), and there'd be a huge understanding of how to fix it.

JonnyBurger:
Should I throw a coon 🤔

kaf-lamed-beyt:
What's a coon? 😃

JonnyBurger:
A coin 😄

kaf-lamed-beyt:
[...]

PR #1860, opened 2023-02-26
feat: change triggers for menu in editor

Just-Moh-it:
Context: 
- https://discord.com/channels/809501355504959528/817306414069710848/1079370595709489184
- https://github.com/remotion-dev/remotion/issues/1853#issuecomment-1445308236


JonnyBurger:
Thanks a lot! 🎉 

I added some changes, because you can also use the tab key and then press enter, and that triggers a Click event.
Another thing I noticed is that if you point down and move it up outside a menu item, the menu disappears on macOS.
Lastly, I also added it to the ComboBox!

PR #1868, opened 2023-02-28
Add layers export

bndkt:
Hi Jonny,

First, I'm experimenting with Remotion and love it so far! I'm deploying it on Lambda, but as all my infra is defined in CDK, I don't want to use the Remotion CLI function to deploy/update the Lambda functions. I have successfully set up my project in a way that deploys the Lambda function as part of my CDK setup and am working on a blog post to document the process. I did, however, have to patch Remotion to add an export for the lambda layers definition. Would you consider adding this export to your package.json? This little detail was the only thing I needed to patch to make it work, so I think it could be a great alternative approach for running Remotion on Lambda!

Benedikt

### DIFF
--- a/packages/lambda/package.json
+++ b/packages/lambda/package.json
@@ -68,7 +68,8 @@
 		"./defaults": "./dist/defaults.js",
 		"./regions": "./dist/regions.js",
 		"./policies": "./dist/api/iam-validation/suggested-policy.js",
-		"./client": "./dist/client.js"
+		"./client": "./dist/client.js",
+		"./layers": "./dist/shared/hosted-layers.js"
 	},
 	"typesVersions": {
 		">=1.0": {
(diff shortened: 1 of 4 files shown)


bndkt:
Closing this myself and will open another one. Just realized this way to add the export was the easiest for my patch but not the cleanest when integrated in Remotion.

JonnyBurger:
Thanks for the PR!

As discussed, let's only keep the arm64 layers. The structure of the symbol has been changed, you can see in the changes the updated structure.

Also I wanted to let you know that if you use this version exactly, that it should work indefinitely, however we will make an update in the future where we will remove FFmpeg from the layers and bake it into the Remotion packages. So check for that before you make an update!

shellscape:
@bndkt any chance you should share the stacks you created to deploy this with CDK?

PR #1872, opened 2023-03-01
Clean up all things in the temporary directory if render fails

JonnyBurger:

### DIFF
--- a/packages/bundler/src/bundle.ts
+++ b/packages/bundler/src/bundle.ts
@@ -18,7 +18,9 @@ const prepareOutDir = async (specified: string | null) => {
 		return specified;
 	}
 
-	return fs.promises.mkdtemp(path.join(os.tmpdir(), 'react-motion-graphics'));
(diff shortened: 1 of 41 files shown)


fwextensions:
On Windows, the contents of the temp folders (with prod, not this PR) get cleaned up, but the folders themselves don't.  The `*remotion*` folders in temp are empty (though the puppeteer ones aren't):

image

I looked at the [deleteDirectory](https://github.com/remotion-dev/remotion/blob/main/packages/renderer/src/delete-directory.ts) code, and it's using shell commands instead of the node API to delete the folders.  I'm guessing the `rmdir` command is always failing, so the directory isn't deleted.  Maybe because node is still holding on to that path?  Why not use the node APIs?

JonnyBurger:
Thanks for all the reports! I have now added a lot of cleanup logic and running the whole test suite which also contains failures does not leave any extra artifacts anymore. Sorry for those!

PR #1879, opened 2023-03-02
feat: serveURL CORS tester

kaf-lamed-beyt:
@JonnyBurger, kindly take a look at this. I've just added the `TestCORS` component. But I don't know where to reference it exactly in the code. I know it should go into `Homepage`, but I may be needing an exact location in the code, so I don't end up breaking the layout. 

...and lastly, if it goes into `Homepage`, how do I test it? Would "cd-ing" the /bundler directory and running the command that starts the server &mdash; by checking `package.json` suffice?

### DIFF
--- a/packages/bundler/src/homepage/homepage.tsx
+++ b/packages/bundler/src/homepage/homepage.tsx
@@ -82,36 +82,54 @@ const TestCORS: React.FC = () => {
 	const [serveUrl, setServeUrl] = useState('');
 	const [result, setResult] = useState('');
 
-	const handleServeUrl = (e: React.ChangeEvent<HTMLInputElement>) => {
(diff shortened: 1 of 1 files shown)


kaf-lamed-beyt:
I took a look at `package.json` in the bundler directory, and I noticed that there's a `test` command that uses vite. 

Could that be the command to test the component. In the browser?

JonnyBurger:
Excellent! Simple but powerful :)

Thanks a lot for this! 🙌  I will reference it to people having problem with CORS a lot for sure!

kaf-lamed-beyt:
Ayy! Thank you so much for this feedback @JonnyBurger. 

Can you recommend an issue that you think I'll be able to take on next?

PR #1884, opened 2023-03-02
docs: tailwindcss install missing deps; npm dev only install

Just-Moh-it:
Setting up tailwindcss in existing projects always resulted in errors and I usually brute-forced my way into making it work. But today I tried to track what changes I made to make it work.

Turns out the docs were missing key dependencies, so added those.

Gist of the PR:

- Converted npm install to npm dev dependency installs, just like how they were in the tailwindcss starter template
- Added missing deps which caused errors

### DIFF
--- a/packages/docs/docs/tailwind.md
+++ b/packages/docs/docs/tailwind.md
@@ -60,7 +60,7 @@ values={[
 <TabItem value="npm">
 
 ```bash
-npm i postcss-loader postcss postcss-preset-env tailwindcss autoprefixer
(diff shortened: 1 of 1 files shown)


Just-Moh-it:
cc @JonnyBurger

JonnyBurger:
Thanks! Fair enough, css-style and style-loader are already dependencies of `@remotion/bundler` but if you have e.g. a Next.js app as well you probably want it as well so it's a good idea to explicitly install them.

Appreciate that and the `-D` is obvious too, thanks! 🙌

PR #1894, opened 2023-03-05
Fix/cli/track-warning-wrong-options

Just-Moh-it:
Just a thing I noticed working in a project with multiple compositions 🙃


JonnyBurger:
Thanks! 🙌

PR #1895, opened 2023-03-05
feat: add code push prettier check in GitHub Actions workflow

### LINKED ISSUE #1858: [Repo DX]: Add a GitHub action to check if files are prettified to ensure code cleanliness
### Why? 🤔

Just saw a few pull requests with un-prettified code and messed up indentation. Another repo I worked on that had a prettier check, and it suddenly clicked in my mind.

We could have a GitHub action that could check if all files are prettified or not, and if they aren't, it'd fail the check and ask the user to commit after running a prettier script.

The prettify script would simply be `prettier -w .` and that'd prettify all code in the repo.

Let me know what you think! Feel free to share any wrong assumptions I'm making.

cc @JonnyBurger 🙃

Just-Moh-it:
As discussed in #1858, we decided to add it in Remotion v4!

Feel free to fast-forward this PR and merge this when v4 is ready to be released. Also, mark the 'unit-tests' test as `Required` in the repo settings!

(closes #1858) cc @JonnyBurger


JonnyBurger:
@Just-Moh-it We will soon add this to Remotion right before merging v4 – but it seems like it doesn't work, `pnpm run prettier:check` is not a command from the root. How did you intend this to work?

Just-Moh-it:
@JonnyBurger It's supposed to be a single command, `pnpm run prettier`, ran from the root to prettify the entire project so that in case a PR author make changes in a few different files and forgets to prettify those files, they could just run `pnpm run prettier` and won't have to go into each file and format using their editor.

The GitHub action ensures that all files are prettified using the `prettier:check` script once a PR is created... if they are not, then it makes the workflow fail and the author has to run the prettier command locally and recommit the prettified version for it to pass again... This would ensure that all files are prettified by the authors so even if prettier acts up sometimes and breaks code (rare but happens in edge-cases), the author themselves could make the necessary edits to make the code and prettier work together well.

At least that's how it's supposed to be... I converted this to a draft because the GH action wasn't running, and I didn't know much about GH actions back then. Though now, I have better tooling and an understanding of GH actions, so I'd have a look soon... 

Feel free to make commits to this PR too if you're able to find a fix!

JonnyBurger:
Superseded by #2299

PR #1900, opened 2023-03-06
Render PDFs and WebP

### LINKED ISSUE #1899: renderStill() should support pdf
Besides `.png` and `.jpg`, `renderStill()` should also support `.pdf` export.

## Use Case

`.pdf` exports aren't raster images like `.png` and `.jpg`, instead it is based on vectors which makes it scale without lost in detail. This is useful in case of printing the exported image in high dpi without having to deal with large `.png` or `.jpg` files to keep the same amount of detail.

## Possible Solution

See MR linked to this Issue

## Definition of Done (DoD)
- [x] `renderStill()` renders a valid `.pdf` file if the `imageFormat: "pdf"` is defined
- [x] Documentation is updated
- [x] Tests are working

bennodev19:
https://github.com/remotion-dev/remotion/issues/1899


JonnyBurger:
Thanks a lot for this! This looks excellent, will review and test it tomorrow!

bennodev19:
still wip as I've not yet fixed the tests and updated the docs :) 
also there are some things to think about like the `clip region` or whether only the `StillImageFormat` should have the `pdf` option instead of `ImageFormat`. But it seems like `StillImageFormat` is only used for the most outer function `renderStill()` and `ImageFormat` in all nested internal functions that are used by `renderStill()` but also at other places.

JonnyBurger:
All clear!

I know I said today but it was a bit busy so I hope to be able to look at the PR regardless of where it stands tomorrow as one of the first things. Much appreciated!

bennodev19:
Alright, I suppose it's ready to be refactored by you or other maintainers, as I'm not that familiar with the remotion code base and probably have missed some side cases. I also have not tested the lambda functionality, although I looks like it should work (but thats not good enough :)).
Thanks a lot :) cheers

JonnyBurger:
Thanks a lot, I can take it from here!
As you've maybe noticed, the still image format type is a bit weird.

We are fixing that in v4 by splitting it up into `StillImageFormat` and `VideoImageFormat`. I already opened an issue for that and I'm going to implement it in this PR and then stage it as a new feature for v4.

PR #1904, opened 2023-03-08
<Rive /> 

kaf-lamed-beyt:
Hi @JonnyBurger please take a look at what I've been able to add. 

I decided to go with the `useRive` hook from `"@rive-app/react-canvas"` since it removes the stress of configuring a lot of things like the source of the video via fetch or something related. 

But, I think I may still be having an issue with the `advance` API since most of the time I tried using it like so below; 

```tsx
const { RiveComponent, rive } = useRive({
     src,
    autoplay: true,
    stateMachines: ["default"],
    artboard: "artboard 1",
    
    if (rive) {
       const animation = rive.load(src)
       animation.advance((currentFrame/fps) * 30)
    }
})
```
It is not part of the `rive` type like, the method itself is not exported from the hook. I found a workaround though. But, I doubt if it'd work because I've not been able to test it properly. 

This is because when i tried running the play ground in the examples folder, with the command below;

```bash
pnpm start
```

I got this error 

image


JonnyBurger:
That's amazing!! Looks really good! Let me make some quick build changes to get you unstuck

kaf-lamed-beyt:
> That's amazing!! Looks really good! Let me make some quick build changes to get you unstuck

Oh! Great!

Thank you so much for this feedback.

JonnyBurger:
It took me a while to get this right 😅 

We need to use the advanced canvas instead and now I built it out so that it renders to a canvas element and we can use the `advance()` API!

The syncing currently has a bug where if you go backwards, it still goes forwards. Also, there is a build system bug, after every change you make to `packages/rive`, you have to run `pnpm build` inside `packages/rive`. But other than that, the proof of concept stands!

JonnyBurger:
What we need next is some docs and making the component as unopinionated as possible (remove all hardcoded parameters like `defaultArtboard`)

kaf-lamed-beyt:
> Also, there is a build system bug, after every change you make to packages/rive, you have to run pnpm build inside packages/rive. But other than that, the proof of concept stands!

Ah! Thanks for pointing this out. And for going building the canvas from scratch! I did see the `advanced-canvas` lib, but I don't know if I'm to use it or not. Their docs do not specify explicitly which package to use or not. 

...and thanks for pointing out the build issue too. I'll pull and check out the example.

kaf-lamed-beyt:
Here you said;

> (remove all hardcoded parameters like `defaultArtboard`)

I'm not sure what you mean by that. But, I just pushed a very minimal doc for the component. Please let me know if it makes any sense.

JonnyBurger:
This is so cool!
Thank you so much for taking this initiative 🙌

/tip 150

kaf-lamed-beyt:
> This is so cool! Thank you so much for taking this initiative 🙌
> 
> /tip 150

Oh wow! I didn't even really do much. And I wasn't even expecting this. Thank you soo much @JonnyBurger 😃

PR #1915, opened 2023-03-10
fix(shapes): fix shape radius on triangle and rectangle

### LINKED ISSUE #1766: Bug in `<Triangle>` with corner radius
```ts
import {makeTriangle} from '@remotion/shapes';
import React from 'react';
import {spring, useCurrentFrame, useVideoConfig} from 'remotion';

export const MysteriousTriangle: React.FC<{
	length: number;
}> = ({length}) => {
	const frame = useCurrentFrame();

	const rotation = new Array(10)
		.fill(true)
		.map((_, i) => {
			return spring({
				fps,
				frame: frame - i * 30,
				config: {
					damping: 200,
				},
			});
		})
		.reduce((acc, curr) => acc + curr / 3, 0);

	const {height, path, transformOrigin, width} = makeTriangle({
		length,
		direction: 'right',
		cornerRadius: 60,
	});

	return (
		<svg
			viewBox={`0 0 ${width} ${height}`}
			style={{
				width,
				height,
				overflow: 'visible',
			}}
		>
			<path
				style={{
					transform: `rotate(${rotation * 360}deg) `,
					transformOrigin: transformOrigin
						.split(' ')
						.map((s) => s + 'px')
						.join(' '),
				}}
				d={path}
				stroke="#20e8c2"
				strokeWidth={10}
			/>
		</svg>
	);
};

```

Slashgear:
Fix #1766 issue.

## Why

Border radius on shapes does not seem to work nicely with triangle and rectangle
I tried to add some unit test to make it more stable.

## How

- fixup typo in joinPoints function
- fixup Rectangle radius shapes on initial point
- fixup Triangle on right orientation.

⚠️ Triangle are not completely fixed, I may need your help on that @JonnyBurger because I don't really see what I miss to find any orientation fix for triangle.

### DIFF
--- a/packages/shapes/src/utils/join-points.ts
+++ b/packages/shapes/src/utils/join-points.ts
@@ -52,11 +52,12 @@ export const joinPoints = (
 					];
 				}
 
+				const shorten = shortenVector(nextPoint, cornerRadius)
 				return [
 					{
 						type: 'M',
-						x,
-						y,
+						x: shorten[0],
+						y: shorten[1],
 					},
 				];
 			}
@@ -82,7 +83,7 @@ export const joinPoints = (
 				}
 
 				const prevVectorMinusRadius = shortenVector(prevVector, cornerRadius);
-				const prevVectorLenght = scaleVectorToLength(prevVector, cornerRadius);
+				const prevVectorLength = scaleVectorToLength(prevVector, cornerRadius);
 				const nextVectorMinusRadius = scaleVectorToLength(
 					nextVector,
 					cornerRadius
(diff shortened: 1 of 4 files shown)


packages/shapes/src/utils/join-points.ts
```diff
@@ -52,11 +52,12 @@ export const joinPoints = (
...
 				}
 
+				const shorten = shortenVector(nextPoint, cornerRadius)
 				return [
 					{
 						type: 'M',
-						x,
-						y,
+						x: shorten[0],
+						y: shorten[1],
```
JonnyBurger: I think this is erroneous, `x` and `y` are the coordinates of the current point, while `shorten` is being calculated based off the next point 🤔

JonnyBurger:
This is not quite it, we now got more failing tests and also seemingly a rotation issue. Thanks nonetheless for kicking this off and making a testbed alongside some discoveries!

Slashgear:
@JonnyBurger Maybe I could take a look this weekend to fix it.
I will add more test on triangle too, to be sure this works nicely

JonnyBurger:
Sounds good! And no sweat, if I end up needing a fix, I'll work on it too

Slashgear:
@JonnyBurger It should be good now 😉

JonnyBurger:
Wow, impressive! 👏  Well done and thanks a lot for sticking to solving this hard problem! 🙌

/tip 100

PR #1918, opened 2023-03-12
Rendering datasets bundle needs webpack override

nitinthewiz:
Looking over the dataset example by Alex Fernandez, at the issues around Lambda based rendering, and from my own experience just now, it seems that when we're trying to render programmatically, the bundler needs access to the webpack override. If we don't provide this, the video output is blank/black.

Simple documentation change.


nitinthewiz:
Thanks for the edits! Those definitely make things clearer!

PR #1937, opened 2023-03-15
Update how the package.json should look like in the Lambda setup

Just-Moh-it:
## Why remove `npx` from `npx remotion`

Since we already add remotion to the dev dependencies in the package.json in step 1, using just `remotion` instead of `npx remotion` would ensure that the remotion cli is on the same version as the remotion installation.

This could be particularly helpful for the lambda policies command, where recently amazon lambda broke all remotion renderers because of their sudden change, and after that, I had to re-assign the roles since the `npx remotion lambda roles` command had different roles now. However, using npx remotion in this case instead of just remotion might lead the users to use the older version of the remotion cli which is cached on their PCs.

And so, removing npx from remotion cli command would ensure users are on a consistent CLI version! Hope my assumptions aren't wrong

(also a few minor beginner-friendly and expert-friendly changes :)

### DIFF
--- a/packages/docs/docs/lambda/setup.md
+++ b/packages/docs/docs/lambda/setup.md
@@ -53,7 +53,7 @@ Make sure no package version number has a `^` character in front of it as it can
 Your package.json should look like the following:
 
 ```json
-  "@remotion/cli": "3.0.0", // Replace 3.0.0 with the current version; install as a dev dependeny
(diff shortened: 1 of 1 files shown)


Just-Moh-it:
Oh, right... I had `@remotion/cli` globally installed. Didn't know remotion without npx isn't possible unless you have it globally installed.

Reverted those changes 😄

JonnyBurger:
Perfect, thanks a lot!

PR #1939, opened 2023-03-15
docs: add font awesome and other sources in contributing guide

Just-Moh-it:
While implementing the toggle for disabling panning, I needed to add a disable-panning icon in the toolbar in the remotion previewer. However, the docs didn't indicate where the icons would be sourced from. Even though the icon didn't make it to the final PR, adding info about 3p sources remotion uses would be helpful for contributors and would save just a few messages on discord 😊


JonnyBurger:
Brilliant! We need details like this, thanks a lot!

Just-Moh-it:
Thanks 😊! Just a note, I saw you enabled auto-merge, but there's still something in the file change that needs your attention 🙂 @JonnyBurger

JonnyBurger:
Whoops - fixed up now, thanks for reacting fast enough!

PR #1946, opened 2023-03-16
fix: clip looped tracks that are too wide

marcusstenbeck:
Using Remotion 3.3.77

Using the [technique for looping an offthread video](https://www.remotion.dev/docs/offthreadvideo#looping-a-video) produces a slightly annoying bug as seen in the video.

https://user-images.githubusercontent.com/616906/225666205-05ac4dc0-66a3-45c5-9cc7-f392005ed901.mp4

The PR contains a fix that solves the problem, but very little work has been done to ensure it doesn't interact in some non-desirable way with other timeline track components.

### DIFF
--- a/packages/cli/src/editor/components/Timeline/TimelineTracks.tsx
+++ b/packages/cli/src/editor/components/Timeline/TimelineTracks.tsx
@@ -32,6 +32,8 @@ export const TimelineTracks: React.FC<{
 	const inner: React.CSSProperties = useMemo(() => {
 		return {
 			height: TIMELINE_LAYER_HEIGHT + TIMELINE_BORDER * 2,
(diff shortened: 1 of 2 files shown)


JonnyBurger:
Hi Marcus,
thanks a lot for the repro!

Do you have any markup that triggers this timeline? It would be good to verify the fix is proper if I could verify it.
Actually adding an overflow is maybe just a hack, we don't intend the layers to be longer than the timeline.

JonnyBurger:
reproduced 👌  thanks!

JonnyBurger:
Managed to fix it by adjusting the math instead! This one is more proper. Thanks for the PR and for the repro!

PR #1951, opened 2023-03-17
chore: handle Minified React error #306

kaf-lamed-beyt:
@JonnyBurger, is there a way I can test this?

I added a little change to the error handling to make it more intuitive though. I copied some RegEx somewhere 😃 that will match the component name by checking if there's any `<` or `/>` in the error message and it extracts the text in between it &mdash; which sorta translates to the component name.

```ts
const componentName = err.message.match(/<\w+>/)?.[0] || ""
const errorMsg = `Error: Failed to render ${componentName}. Please check that it is imported correctly.`
```

Please review and let me know If you'd love me to make any change.

### DIFF
--- a/packages/cli/src/handle-common-errors.ts
+++ b/packages/cli/src/handle-common-errors.ts
@@ -49,4 +49,13 @@ export const handleCommonError = async (err: Error) => {
 			)
 		);
 	}
+
+	if (err.message.includes("Minified React error #306")) {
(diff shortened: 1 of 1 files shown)


JonnyBurger:
Thanks a lot @kaf-lamed-beyt! 🎊 Another one 🎉

kaf-lamed-beyt:
It is a pleasure, my chief! 🙌🏽🙌🏽

PR #1961, opened 2023-03-26
Updated setup links for code sharing

thecmdrunner:
This PR updates the docs page at https://www.remotion.dev/docs/player/integration

The [create-react-app](https://reactjs.org/docs/create-a-new-react-app.html) link mentioned in the above docs now redirects to the new React docs, which recommend using a framework for a new React project, instead of CRA. So Next.js and Remix are mentioned, while still keeping a note for create-react-app users with the updated link.


JonnyBurger:
Thanks for raising this, it is indeed very important that we give good recommendations!
I also added Vite and updated the https://remotion.dev/docs/brownfield page which has a very similar topic.

Issue #1970, opened 2023-03-31
Create Dockerfile for other Linux distribution


JonnyBurger:
Problem: We currently have a page for how to dockerize a Remotion app here: https://www.remotion.dev/docs/docker

However, this Dockerfile broke because Alpine Linux updated its packages and removed the old ones. The new packages don't work and make Chrome crash.

It is not good for Remotion users if it is impossible to pin the versions in the Dockerfile, so we should move away from Alpine Linux and use another Linux Distro.

Adding a bounty for a few days if anyone is up for the task, before I will take a look myself on Monday!

/bounty 250

Acceptance criteria: Update the Dockerfile and the step by step instructions here: https://github.com/remotion-dev/remotion/blob/main/packages/docs/docs/docker.md

JRavi2:
Hey, @JonnyBurger Can I pick this up?

JonnyBurger:
@JRavi2 Happily!

raminious:
my suggestion:

```
FROM debian:bookworm

RUN apt-get update
RUN apt-get install -y wget
RUN wget -q https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
RUN apt-get install -y ./google-chrome-stable_current_amd64.deb
RUN apt-get install -y nodejs npm

ENV PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=true

################
### BUILD HERE ###
################

ENV CHROME_EXECUTABLE_PATH="/usr/bin/google-chrome-stable"

################
### RUN HERE    ###
################
```

JRavi2:
Hey, just an update. Looks like chromium is not able to communicate with dbus properly. Have tried installing dbus and running the daemon explicitly but was getting Connection refused. Will try to work on the problem further by trying with different base image/chromium versions.

JonnyBurger:
Hey @JRavi2! Thanks a lot for providing an update and trying hard! 😃

@raminious (also on Discord, filer of the original issue) has found a Dockerfile in the meanwhile that works. Are you willing to work together and split the bounty? As far as I understand, the Dockerfile is already working but is not yet documented.

I am also okay with any other solutions, it would be super cool if you guys could coordinate on a solution that helps get the Dockerfile updated.

raminious:
@JonnyBurger I didn't post a PR since this issue was assigned to @JRavi2. Nevertheless, I did it now since it seems he is having some difficulties.

JRavi2:
[...]

PR #1975, opened 2023-04-02
Implement Dockerfile based on Debian bookworm

raminious:
/claim #1970


JRavi2:
@raminious Awesome! 
@JonnyBurger I think obviously @raminious should have the bounty as they solved the issue.
Also, just curious, why does the debian base image work and not the alpine one? Is it something related to the how the base image sets up the environment or do they have different versions of chromium/chrome?

JonnyBurger:
Thanks a lot for this!

I found a small caveat which is that the current Debian image is unable to render WebGL content. I added a note for now, but of course I am glad that we are able to replace the non-functioning WebGL Docker image with a working one!

JonnyBurger:
@raminious You are not expected to, it might be impossible on Debian.
But if you stumble upon a solution, of course sharing it is appreciated!

PR #1978, opened 2023-04-03
chore: add absolute and public paths conditions to `createRule()`

kaf-lamed-beyt:


kaf-lamed-beyt:
Hi @JonnyBurger, kindly take a look. 

I left some dummy links to the docs, because I don't know if the links are available or if i'm required to create them

```ts
const AbsoluteStaticFile = [
  "Do not pass an absolute path to staticFile().",
  "See: https://remotion.dev/docs/upcoming-docs-for-absolute-paths",
].join("")

const PublicStaticFile = [
  "Do not pass a public/ path to staticFile().",
  "See: https://remotion.dev/docs/upcoming-docs-for-public-paths",
].join("")
```

JonnyBurger:
Thanks a lot! I just added it to the same documentation page again because I think the errors are all very similar.
Appreciate this super nice contribution, I think now we covered all the ways one can do a mistake :D

kaf-lamed-beyt:
Ah, great! I'm glad I could help. 

...and thanks always, for the opportunity! 🙌🏽🙌🏽

PR #1990, opened 2023-04-05
warning when inputProps are provided

patsalv:


packages/cli/src/editor/components/RenderModal/RenderModalData.tsx
```diff
@@ -39,6 +51,8 @@ export const RenderModalData: React.FC<{
 		return composition.schema.safeParse(inputProps);
 	}, [composition.schema, inputProps]);
 
+	const cliProps = getInputProps();
+	console.log(Object.keys(cliProps).length);
```
JonnyBurger: I thought you would not commit it this time 😛

patsalv: I thought so too👀😂

JonnyBurger:
Thanks! 🐣

PR #2015, opened 2023-04-12
remove "runtime" from @babel/preset-typescript config

JLHwung:
Babel's maintainer here. The `runtime` option is invalid and never supported in `@babel/preset-typescript` ([docs](https://babel.dev/docs/babel-preset-typescript) here). Currently Babel 7 allows unrecognized option name in presets, in Babel 8 we will throw on invalid usage. Hence removed the option here so we can be future-proof.


JonnyBurger:
Thanks a lot for correcting this! 🙌🏼

Issue #2012, opened 2023-04-12
Experts page: Clicking any button does take you to the detail page instead of the profile


JonnyBurger:

harshbisle:
Can I give it a try?

MehmetAdemi:
For sure!

harshbisle:
Thank you!

Slashgear:
@harshbisle this is html issue here, the whole card in a `Link` and they are link within that are "captured" by the parent link.
Social link should be siblings, not children.

i could take this If you want I would try to explain how I fixed it in my Pull Request in order to let you know how this works ?

MehmetAdemi:
@Slashgear Thanks for the hint! That would be much appreciated!
@harshbisle Alternatively, you can tackle the "Find a template" issue, of course, only if you want and have time :)

harshbisle:
@Slashgear Thank you for the hint, working on it. 
@MehmetAdemi Yes, will submit a PR for this by the EOD and will move on to the "Find a template" issue. Is this feasible?

MehmetAdemi:
For sure, appreciate your commitment! Let's do it this way!

harshbisle:
Please have look at the PR. @Slashgear thanks again for the hint.

I have messed up some CSS styles. I would be glad to make the required changes based on your input.
 
Thank You!

Slashgear:
This could be considered closed @JonnyBurger ;)

MehmetAdemi:
@harshbisle Awesome! Thanks for the contribution! 
Thanks for guiding @Slashgear! Very much appreciated. 💙

PR #2024, opened 2023-04-17
fixes: expert page buttons #2012

harshbisle:
Fixes the issue-

Experts page: Clicking any button does take you to the detail page instead of the profile #2012

Also, I think I messed up somewhere in the CSS styles. Ready to improvise based on the review.

Thank You.

### DIFF
--- a/packages/docs/src/pages/experts/index.tsx
+++ b/packages/docs/src/pages/experts/index.tsx
@@ -78,13 +78,15 @@ const Experts: React.FC = () => {
           <br />
           {expertsInRandomOrder.map((e) => {
             return (
-              <Link key={e.name} style={link} href={`/experts/${e.slug}`}>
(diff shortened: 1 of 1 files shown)


Slashgear (commented):
**Praise:** We need to re-adapt the style in order not to break the design, basic idea here nonetheless :)

harshbisle:
Yes, but please elaborate on where exactly I am doing wrong.

Slashgear (approved):
Does the job @JonnyBurger LGTM

harshbisle:
> Thanks for this @harshbisle and for the guidance @Slashgear!
> 
> I readjusted the layout to not be exactly the same, but I think this will do for the moment.

Thank you for fixing the layout! I apologize for the mistake.

PR #2027, opened 2023-04-18
Faster `<OffthreadVideo>`

JonnyBurger:
TODO: 

- Is the VP8 scale thing still relevant?
- Should do a test where frames are in reverse
- Should do a test where same videos are being extracted with two different timestamps
- Should remove videos from memory and do general memory testing


.vscode/settings.json
```diff
@@ -3,5 +3,11 @@
   "editor.codeActionsOnSave": {
     "source.organizeImports": false,
     "source.fixAll": true
+  },
+  "rust-analyzer.cargo.extraEnv": {
+    "REMOTION_FFMPEG_RUST_BINDINGS_OUT_DIR": "/Users/jonathanburger/remotion/packages/compositor-darwin-arm64/ffmpeg",
```
JonnyBurger: How to parametrize this?

packages/example/src/RemoteVideo/index.tsx
```diff
@@ -1,14 +1,7 @@
...
 const RemoteVideo: React.FC = () => {
-	return (
-		<Video
-			volume={(f) =>
-				interpolate(f, [0, 500], [1, 0], {extrapolateRight: 'clamp'})
-			}
-			src="https://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4"
-		/>
-	);
+	return <OffthreadVideo src={staticFile('bigbuckbunny.mp4')} />;
```
JonnyBurger: should revert

PR #2028, opened 2023-04-18
Add support to define duration in frames for composition based on default props and fps

sauldom102:
## Why this change?

Now that in v4 it's possible to define props from the web preview, it makes sense to add support to allow changing the duration of our compositions based on the default props, so that it's not needed to change the code at all to make a custom duration video.


JonnyBurger:
Thanks a lot for raising this!

I agree with this, and I am going to work on this PR and document it as a feature for V4.

JonnyBurger:
Hi @sauldom102!
I am closing this in favor for #2311, where you can also do async data fetching, memoize the data fetching function, change width and height, and there will be a tutorial in the documentation 🙂

Thanks nonetheless for sending the PR, I too am excited for this feature!

PR #2031, opened 2023-04-18
fix(cli): fix design of long timeline sequence name

### LINKED ISSUE #2016: Label for Sequence name is multiline
Improve the layout of this:

image

Probably it is just an `<Audio>` tag with a long `src`

Slashgear:
## Why

Fix #2016 

## How

In order to fix the bug I took the possibility of truncating the title of the sequence with overflow.
Sequence name will be limited by 1 line and won't wrap.

But, as the TimelineListItem is a flexible container with `flex: 200 1 0%` which allow timeline name to overflow from the container, the ellispsis don't really work.
This would mean a big refactor of the left navigation in order to to use a `flex-grow` of 200.

Maybe @JonnyBurger you have an opinion on that.

## How to validate

You can modify `packages/example/src/VideoTesting/index.tsx` sequence name to check that overflow in handled correctly.

### DIFF
--- a/packages/cli/src/editor/components/Splitter/SplitterElement.tsx
+++ b/packages/cli/src/editor/components/Splitter/SplitterElement.tsx
@@ -15,6 +15,7 @@ export const SplitterElement: React.FC<{
 			display: 'flex',
 			position: 'relative',
 			overflow: 'hidden',
+			flexDirection: 'column',
 		};
 	}, [context.flexValue, type]);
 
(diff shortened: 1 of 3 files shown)


JonnyBurger:
Thanks a lot! 🙌 and I even managed to fix the ellipsis! 😁

https://user-images.githubusercontent.com/1629785/233288358-f5d5b2d7-c391-40b9-9ce6-5229c45a19c4.mov

The trick was to add flex-direction: column to the containers that you don't want to expand horizontally, this is what the AI said:

image

Slashgear:
Nice @JonnyBurger TY for fixing the ellipsis !

PR #2126, opened 2023-04-21
Migration image format

patsalv:


JonnyBurger:
@patsalv Can you mention that we are referring to the type? And mention from which package the type is coming from?

JonnyBurger:
🙌

Issue #2029, opened 2023-04-18
Home page: "Find a template" button doesn't appear as button but as link


MehmetAdemi:
Reported by: @harshbisle

harshbisle:
Thanks, @MehmetAdemi was about to comment. Hope I can pick this up now.

MehmetAdemi:
Let me know if you encounter any issues or get stuck!

harshbisle:
For sure!

harshbisle:
Fixed by #2138

MehmetAdemi:
@harshbisle 
Great, thanks!

PR #2138, opened 2023-04-24
fix:  template button behaving as link 

harshbisle:
Fixes:

Home page: "Find a template" button doesn't appear as button but as link #2029


JonnyBurger:
Hmm, I kinda disagree with this 😕

Links are better right? Since you can for example right-click and press "Open new tab".
Or it works without JavaScript.

Am I missing a benefit of doing this?

harshbisle:
Totally! I agree that links are better. But we can also right-click and open a new tab using the button. 

Here the PR is just tackling the issue of the Template button rendering as a link occasionally. 

If not, how would you suggest doing it?

JonnyBurger:
I see the problem now! Thanks, will review it soon.

JonnyBurger:
Ok, I reverted it now to a link to keep it an <a> tag but kept the fix otherwise. Thanks a lot for the polish! 😃🙌🏼

harshbisle:
Awesome!

PR #2145, opened 2023-04-25
feat: add `scalechange` event on player

### LINKED ISSUE #2011: `scalechange` event on `@remotion/player`'s ref
# Feature Request 🛍️

Was wondering if we could add a `scalechange` event on the remotion player ref just like how `timeUpdate` and other events currently exist.

## Use Case

I've been injecting the gesture controls for a wysiwyg editor into the player and since it contains controls/gesture-handles that need to be consistently sized regardless of the view port, I apply the inverse of the editorZoom to the controls, so the scaling of the video doesn't affect the controls and it cancels out.

However, adding more and more functionality is slowing down the application and I think that getting the zoom every few seconds would prove to be a major performance drain as I add more features. So having a `scalechange` event would be a life-saver here!

@JonnyBurger let me know if you want to assign me this issue, happy to add this 😊

## Possible Solution

A `scalechange` event on the playerRef!

Just-Moh-it:
Player demoed by adding console log [here](https://github.com/remotion-dev/remotion/compare/main...Just-Moh-it:remotion:feat/scale-event?expand=1#diff-4b871b7f714b266b5bfeba3d196f8a0eb1d83e0a9207df46e01995109c7cbdfdR146)

https://user-images.githubusercontent.com/48997634/234204284-cc37ef52-1794-47e5-80fa-c91ed11d0f74.mov

## Miscellaneous

Renamed `dispatchFullscreenChangeUpdate` to `dispatchFullscreenChange` for consistency (all internal functions either have the `Change` or `Update` suffix, but this fn had both 🙃)

Fixes #2011


Just-Moh-it:
@JonnyBurger I accidentally pushed compositor, but it's not reverting... if it's alright, we could merge it but if it's not, please revert it on your end 😅

packages/player-example/src/App.tsx
```diff
@@ -440,8 +449,8 @@ const ControlsOnly: React.FC<{
 				.reverse()
 				.slice(0, 10)
 				.reverse()
-				.map((l) => {
-					return <div key={l}>{l}</div>;
+				.map((l, i) => {
+					return <div key={`${l}-${i}`}>{l}</div>;
```
Just-Moh-it, line 453: Scale change when resizing up was alright but resizing down meant that some log content was duplicated (since scales were repeated)... this should fix that 😊

JonnyBurger: Very nice! 💡 I noticed it was buggy!

JonnyBurger (dismissed):

Issue #2034, opened 2023-04-19
Color of inline button is always white in editor


JonnyBurger:

Slashgear:
@JonnyBurger Could you tell more on this issue, I may be able to fix it ;)

JonnyBurger:
Ok, didn't expect contributors for V4 issues, but I'll gladly explain it! 🙂

There is a component, `<InlineIcon>`, whose icons are always white (`#fff`). They are meant to be `LIGHT_TEXT` instead, but in the Remotion Preview, we need to set all styles explicitly with inline CSS.

This is very special, but consider if someone imports a CSS file `* { background: red; font-family: Comic Sans !important; }`, then the Remotion Preview should not be affected by this. The way we do this is pass inline style 😄

But it is not implemented currently. We probably want a hover listener (like this https://github.com/remotion-dev/remotion/blob/4cd40c8a711b1258e6aeccdf341438ccd8e5c1ae/packages/cli/src/editor/components/NewComposition/RemInputTypeColor.tsx#L32) and then pass the color down to the icons explicitly.

Let me know if you want the issue, then we'll assign it to you!

Slashgear:
Can't wait for the V4 so If I could help a bit, that would be great !

I think I see what to do, I will try to fix it.

JonnyBurger:
Awesome! Thanks a lot and good luck!

Slashgear:
OK

JonnyBurger:
Fixed by #2148

PR #2148, opened 2023-04-25
Fix button color

Slashgear:
Related to #2034

Not sure that was what you intended by that, it seems that color and background was already overriden.
Let me know If I am wrong


JonnyBurger:
Thanks for the PR!

It is not what I meant though - you are right that the style `color: LIGHT_TEXT` is already passed, but if you look at the preview, the icon color for every icon is `#ffffff`, so we need to pass down the color for every icon, so that it is gray when unhovered and white when hovered

Slashgear:
I am on it

Slashgear:
@JonnyBurger Normally this should do the trick, let me know if this ok.

JonnyBurger:
Good idea, but Remotion is different 🤣
`currentColor` is always white because Remotion resets the style for every element, so it does not work, we have to explicitly pass the color.

Thanks for the attempt though!

PR #2154, opened 2023-04-26
Add template: Google Text-to-speech (for v4)

thecmdrunner:
1. Add new template for Google Text-To-Speech at https://github.com/thecmdrunner/remotion-gtts-template
2. Label both TTS templates distinctly as 'Azure' and 'Google' to avoid confusion

On line 201, the promo video link mentioned needs to be replaced with the actual `muxId`.


JonnyBurger:
Nice! Excited about this 🙌

thecmdrunner:
Yeah! Can't wait for it to be in the official CLI!

Issue #2129, opened 2023-04-21
Allow fragments for `Series`


JonnyBurger:
Similar to how TransitionSeries added support for fragments:

https://github.com/marcusstenbeck/remotion-transition-series/commit/9a017922ba7a7876f728289b6b5cb73763192a26#diff-b32088b0064b3b7c1e269133ef72e53fdcc70613c13dd71b09636c0880cd3c3d

/bounty 40

kaf-lamed-beyt:
@JonnyBurger, would modifying `childrenValue` to accept `Series.Sequence` and `React.Fragment` suffice in this case?

rjackson:
@JonnyBurger Been looking into this one and, unless I'm missing something, it appears Series already supports fragments?

The Series component accepts `children` (React.ReactNode), flattens all nested children into a one-dimensional array, and then checks to make sure every item in that children array is of type SeriesSequence:

https://github.com/remotion-dev/remotion/blob/784fd282a16d3917651d7356c51b440b6718d2c1/packages/core/src/series/index.tsx#L59-L63

And there's a currently-passing test proving this works with fragments:

https://github.com/remotion-dev/remotion/blob/784fd282a16d3917651d7356c51b440b6718d2c1/packages/core/src/test/series.test.tsx#L75-L98

JonnyBurger:
@rjackson You are totally right!

For context, I opened this issue to not forget about it when I wasn't in a position to check myself at that time.
Thanks for investigating this and if you submit a dummy PR we'd be happy to honor it with half of the bounty amount! 😄

JonnyBurger:
Ok, plan change, Algora made a new feature for us where I can tip you directly! Let's try it out!

/tip $20 @rjackson

PR #2187, opened 2023-05-04
Update series fragment test to cover nested fragments

rjackson:
As mentioned in the comments on the issue, the Series already support React Fragments.

The underlying code flattens all children given to Series, and we had an existing test verifying that behaviour with a single layer of fragments.

This PR expands the test to verify that nested fragments are parsed properly. This feels like a very silly test and use-case, but the code does support nested fragments so we ought to make sure they're tested too.

Solves #2129
/claim #2129


JonnyBurger:
Thanks for the test! 😁

PR #2189, opened 2023-05-05
upgrade rust CI depending on node 12

### LINKED ISSUE #1921: Update `action-rs/toolchain@v1`GH actions node env to > v12 (deprecated)
image

While checking out the progress of some GitHub actions on this repo recently, I noticed a bunch of node v12 deprecation warnings. Since some actions in the GH Actions workflow run directly/indirectly on node v12, and it is supposed to be dropping support by early 2023 (we might just be a few days away from support being dropped), we should migrate away from using node v12... 🙃

JonnyBurger:
Fixes #1921

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -111,8 +111,7 @@ jobs:
         with:
           channel: "1.67.1"
           profile: minimal
-          targets:
-            - x86_64-pc-windows-gnu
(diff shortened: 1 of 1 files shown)


JonnyBurger:
https://github.com/moonrepo/setup-rust/issues/7

PR #2190, opened 2023-05-05
Update sentence for Lambda CLI render.md

UmungoBungo:
It doesn't make sense to 'deploy a project to Remotion'


JonnyBurger:
So true 😅 Thanks for sending a fix!

PR #2203, opened 2023-05-06
Update validate-privacy.ts public-read error

UmungoBungo:
If you set privacy to `public-read`, as the error message says you should, you would end up with the same error 😅


JonnyBurger:
Excellent find! Thanks a lot!

PR #2204, opened 2023-05-08
CLI command `remotion lambda regions` required AWS creds

UmungoBungo:
Fetching the available regions shouldn't need AWS env variables.

I think this had been accounted for, but incorrectly, by placing it under the Policies CLI command along with USER_SUBCOMMAND and ROLE_SUBCOMMAND.

Before:
image

After:
image


JonnyBurger:
Oh, very nice! Thanks a lot! 🙌

PR #2214, opened 2023-05-09
added more info around envVariables input

UmungoBungo:
Saw the confusion in [Discord](https://discord.com/channels/809501355504959528/817306238811111454/1105149863227822231), thought I would add Jonny's comment to the docs.

### DIFF
--- a/packages/docs/docs/cli/compositions.md
+++ b/packages/docs/docs/cli/compositions.md
@@ -35,7 +35,7 @@ Specify a location for the Remotion config file.
 
 ### `--env-file` <AvailableFrom v="2.2.0" />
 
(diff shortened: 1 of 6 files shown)


JonnyBurger:
Thanks a lot! I guess this is confusing for real 🤔 
Here you **can** specify your credentials!

The .env file can contain credentials and the variables also get forwarded to the render (ok to send credentials, since they are only in memory and not part of the site). This also if you specify a custom path for the env file.

The problem was that passing `envVariables` in the Node.JS APIs does not add the variables to `process.env`, instead you you should set them manually using CLI or using the dotenv.

JonnyBurger (approved):

Issue #2210, opened 2023-05-08
Implement a `reverse` prop for spring()


JonnyBurger:
You can already reverse a spring function:

```ts
const spr: Partial<SpringConfig> = {
	damping: 200,
};

const duration = measureSpring({fps, config});

const two = spring({
	fps,
	frame: duration - frame,
	config,
});
```

But (even for me) it requires quite some mental gymnastics to figure that out.

**Bounty**
- Should implement a `reverse` parameter for spring() that takes care of this calculation (default `false`)
	- Important: Take into account the `durationInFrames` option!	
- Should document in spring.md and label as available in the next Remotion version using `<AvailableFrom>`
- Write 1-2 unit test assertions

JonnyBurger:
💎 This issue has a bounty on it!

Read our contributing guidelines: 
- **How to contribute**: https://www.remotion.dev/docs/contributing/
- **Implementing a new feature**: https://www.remotion.dev/docs/contributing/feature
- **Implementing a new option**: https://www.remotion.dev/docs/contributing/option
- **Write documentation**: https://www.remotion.dev/docs/contributing/docs
- **Bounty information**: https://www.remotion.dev/docs/contributing/bounty

/bounty 75

tachyonicbytes:
Hi! I would like to work on this issue! And nice to meet you, and the whole Algora project.

JonnyBurger:
@tachyonicbytes Awesome! Good luck! 🙌

PR #2222, opened 2023-05-09
#2210 Add option for rendering a spring animation

tachyonicbytes:
This PR implements the `reverse` option for the spring animation rendering. I added tests and updated the documentation where necessary. The implementation is very close to the one discussed in the [original issue](https://github.com/remotion-dev/remotion/issues/2210).

I am looking forward to implementing any feedback from the review process to things for which I may have blind spots.

/claim #2210

### DIFF
--- a/packages/core/src/spring/index.ts
+++ b/packages/core/src/spring/index.ts
@@ -28,10 +28,10 @@ export function spring({
 	config = {},
 	from = 0,
 	to = 1,
-	durationInFrames,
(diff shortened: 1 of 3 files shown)


JonnyBurger:
Awesome!
I put in a small performance improvement so that `measureSpring` is not called if it is not needed, just a detail, but awesome work!Thanks a lot for your contribution! 🙌

Issue #2201, opened 2023-05-05
Copy button in info popover in rendermodal polish


patsalv:
no feedback if copying has succeeded or not. Color of icon does not change on hover -> inconsistent to other buttons

PR #2223, opened 2023-05-10
copy to clipboard polished

patsalv:
#2201


packages/cli/src/editor/components/RenderModal/CliCopyButton.tsx
```diff
@@ -17,17 +35,56 @@ export const CliCopyButton: React.FC<{valueToCopy: string}> = ({
...
+			setCopied(false);
+			setHovered(false);
+		};
+
+		const to = setTimeout(() => handleClear(), 2000);
+		return () => clearTimeout(to);
+	}, [copied]);
+
+	// eslint-disable-next-line no-negated-condition
+	return !copied ? (
```
JonnyBurger: You can fix this warning by changing the order: `copied ? <span/> : <button />`...

It's an annoying rule and we might remove it but would be good to align it for now

Excellent otherwise 👍

PR #2231, opened 2023-05-12
remove bucketName input from lambda GetSitesInput type

UmungoBungo:
This input is not used in the code.

Not sure if you want to retain the GetSitesInput or put the type declaration inline?  
Removing the exported GetSitesInput type would mean removing it from `packages/lambda/src/index.ts` as well. I couldn't see that it was used anywhere else, so it would probably be a good thing to clean up - let me know if you want me to do this.

### DIFF
--- a/packages/lambda/src/api/get-sites.ts
+++ b/packages/lambda/src/api/get-sites.ts
@@ -16,7 +16,6 @@ type Site = {
 
 export type GetSitesInput = {
 	region: AwsRegion;
-	bucketName?: string;
 };
 
 export type GetSitesOutput = {


JonnyBurger:
Very strange 🤔 Guess we changed our mind about the API and this is the result of that.
Since it's not used right now and nobody has a problem with it, let's just remove it for now. Thanks!

JonnyBurger:
Very strange 🤔 Guess we changed our mind about the API and this is the result of that.
Since it's not used right now and nobody has a problem with it, let's just remove it for now. Thanks!

PR #2239, opened 2023-05-12
initial framerange for render based on timelineInOutFramePosition

patsalv:
If timelines in and/or out marker are used, their values are taken as initial framerange in the render modal


JonnyBurger:
Very smart, but I wonder what happens if you for example:

Are on composition A which has outFrame = 150
But you render composition B which only has a duration of 100

This would result in a frame range that is out of bounds?

It would be better to only do this if you render the current composition, not any composition.
So maybe instead add it to the render modal initialization payload

PR #2243, opened 2023-05-13
fix zod warnings in v4

### LINKED ISSUE #2232: fix currently neccessary @remotion/zod-types install
Bug: Even if zod is never used in a remotion project, @remotion/zod-types has to be installed, otherwise an error will be thrown in `get-zod-if-possible.js`  (line 31 & 41)

JonnyBurger:
Thanks @patsalv for reporting - fixes #2232


JonnyBurger:
Hmm no this does not seem to do the trick

PR #2247, opened 2023-05-15
lambda region-selection.md improved wording

UmungoBungo:
There was a slight typo, which I fixed, and IMO it is worded a little more clearly now.

I won't be offended if you disagree my rewording 😅 but there is definitely an error with the original;
> not being able to benefit less from already warm functions


JonnyBurger (approved):
Whoops - thanks for the fixes!

PR #2256, opened 2023-05-15
`@remotion/docs`: fix homepage scroll in mobile view

Just-Moh-it:
https://github.com/remotion-dev/remotion/assets/48997634/fc50ae9e-a325-48ce-9162-8841ecd8c61d


Just-Moh-it:
Update–changed to draft because the width was fixed on mobile but broke on desktop

Just-Moh-it:
Mobile shouldn't break because there's enough space already, but it's breaking because link tags have padding and the find-a-template button has non-responsive margin on the left. May not require `width: 100%;` if we remove the padding and use flex with `gap` instead of margin! Will look into it soon

patsalv (approved):

PR #2259, opened 2023-05-16
Loop: Updated to use a single Sequence to fix video gaps

### LINKED ISSUE #2207: `<Loop>` component will re-mount a component due to `key` change
Issue from Discord here:
https://discord.com/channels/809501355504959528/817306414069710848/1098984011617017947

Consider the following markup:

```tsx
<Loop durationInFrames={100}>
	<Video src={staticFile("video.mp4")}>
</Loop>
```

then the video will be re-mounted every 100 frames, leading to a flicker (https://www.remotion.dev/docs/troubleshooting/player-flicker)

I posted as a response a snippet that does not have this behavior, it would be nice to refactor the Loop component so it also does not have this behavior!

deevus:
This took me a few hours to understand what was going on. To be fair I had not looked at the source code for Remotion before.  As noted in https://github.com/remotion-dev/remotion/issues/2207#issuecomment-1539393311 the issue is that `Sequence` unmounts it's children when it isn't in frame. Looping using a collection of `Sequence` components therefore has this gap issue because of this unmounting and remounting of the underlying `Video` component. 

This solution uses a bit of code from [Discord](https://discord.com/channels/809501355504959528/817306414069710848/1098984011617017947) modified such that it will stop after the maximum frame has been reached. 

Fixes #2207 
/claim #2207


JonnyBurger:
Thanks a lot! 🥳 🙌 

Sorry for the inaccurate issue description leading you the wrong way. We'll definitely try to learn from that to improve in the future! Even more impressive that you hung in there! ✊

Also unexpectedly, this led to the Loop component being displayed in a weird way in the timeline - fixed that up now as well!

JonnyBurger (approved):

deevus:
No worries @JonnyBurger thanks for the help along the way. Hopefully I can contribute more in the future.

PR #2272, opened 2023-05-16
feat(v4/cli/render): add `--output`

### LINKED ISSUE #2249: Support for `--output` in `remotion render` command to customize render output file location.
# Feature Request 🛍️

Just like how the lambda cli command the `--out-name` flag for setting the output file name, it would be great to have an option for `--output` in the cli for normal rendering to specify the path of the file.

## Use Case

I'm using the `remotion render` command in a GitHub action, and I have a composition called `base-comp` that is rendered in the workflow. Since it requires you to specify the path of the file to save as downloadable artifacts on GitHub actions run, having an obscure file called `out/base-comp.mp4` is weird and limits the possibility of using automations like Zapier (since they require a specified file-name), but the comps I render are dynamic, and I don't like the idea of having to change the comp names for the sake of uniformity in output file names...

## Possible Solution

Just support the `--output` flag in the CLI, simple 😊

PS: @JonnyBurger you could assign this to me, would love to learn more about Remotion's CLI's inner workings!

Any tips welcome too 😉

Just-Moh-it:
### Flag only
Screenshot 2023-05-17 at 2 07 21 AM

### Arg only
Screenshot 2023-05-17 at 2 07 58 AM

### Both flag and arg
Screenshot 2023-05-17 at 2 08 40 AM

Fixes #2249
/claim #2249


Just-Moh-it:
@JonnyBurger just updated the desc to include scrnshts!

JonnyBurger:
Excellent! 🙌 Reviewing tomorrow

JonnyBurger:
Let's go! 🙌

PR #2285, opened 2023-05-17
v4: Mark React Native Reanimated as optional dependency

JonnyBurger:


JonnyBurger:
We backport this feature to v3

PR #2293, opened 2023-05-17
added derived state -> fixed

### LINKED ISSUE #2288: Fix right sidebar warning glitch
When changing from a component which doesn't define a zod schema to a component which defines one, whilst having the props panel in the right sidebar open, warnings which should only be displayed in the json view get displayed in the props view upon mounting and disappear within e few ms.

patsalv:
#2288
#2053


JonnyBurger (approved):
Perfect! 🧑🏼‍🍳

PR #2310, opened 2023-05-20
Make a @remotion/tailwind package for Webpack override

### LINKED ISSUE #2304: Make a `@remotion/tailwind` package for Webpack override
Similar to the `@remotion/skia` package, this should have an `enable` function that will do the full Webpack override. It should include all dependencies that are needed for it (tailwind, postcss-loader, style-loader etc.)

The steps 1-3 https://www.remotion.dev/docs/tailwind#install-in-existing-project should be replaced and refer to the `@remotion/tailwind` package.

Two new documentation pages should be created:
- Package overview page like https://www.remotion.dev/docs/skia/
- `enableTailwind()` page like https://www.remotion.dev/docs/skia/enable-skia

/bounty 175

rjackson:
This PR adds a new @remotion/tailwind package which exports a `enableTailwind` function, to be used to customise the Webpack configuration as required to install TailwindCSS.

There are two new documentation pages for the `@remotion/tailwind` package (`docs/tailwind/tailwind` and `docs/tailwind/enable`), and the original TailwindCSS page has been updated to instruct users customise their Webpack configuration via `@remotion/tailwind` rather than doing it themselves.

The `enableTailwind` function implements the exact Webpack configuration that was previously in the docs page.

The overall package is a copy of the `@remotion/skia` package, but is a little simpler as there are no React components to export alongside the `enableTailwind` function. I had a choice to make here on whether to follow the `@remotion/skia` pattern of maintaining a separate export for the enable function, but I decided not to do so as the index export would be empty in that case.

I've expanded the example project to use `@remotion/tailwind` and included one video as an example that uses Tailwind classes, which renders as follows:

https://github.com/remotion-dev/remotion/assets/602850/640d2c10-8226-486f-9757-a0e8fac3d112

Most of my time on this PR was lost putting that example together, surprisingly. I was having a lot of "module not found" errors with the webpack override that I was really struggling to track down, which I believe was a bug or oversight in the `@remotion/cli` package: Custom webpack configs are evaluated (eval) by `@remotion/cli`, so any custom packages used in the webpack override or `remotion.config.js` that are not present in `@remotion/cli` cannot be loaded.

https://github.com/remotion-dev/remotion/blob/c533f88eed38a95be6c8fa5ac2c03d41827d4828/packages/cli/src/load-config.ts#L53-L55

I've included a quick fix for that by adding `@remotion/tailwind` and `@remotion/skia` as dependencies of `@remotion/cli`, but that wouldn't help if any users of Remotion were trying to load third-party scripts in their config files. We may want to look into a more resilient solution there separately.

Closes #2304
/claim #2304


packages/tailwind/src/enable.ts
```diff
@@ -0,0 +1,49 @@
...
+          if (rule === "...") {
+            return false;
+          }
+
+          if (rule.test?.toString().includes(".css")) {
+            return false;
+          }
+
+          return true;
+        }),
```
Just-Moh-it:
```suggestion
        ).filter((rule) => rule !== "..." && !rule.test?.toString().includes(".css")),
```

packages/docs/docs/tailwind.md
```diff
@@ -60,110 +60,59 @@ values={[
...
-yarn add -D postcss-loader postcss postcss-preset-env tailwindcss autoprefixer css-loader style-loader
+yarn add -D @remotion/tailwind postcss-loader postcss postcss-preset-env tailwindcss autoprefixer css-loader style-loader
 ```
 
   </TabItem>
   <TabItem value="pnpm">
 
 ```bash
-pnpm i -D postcss-loader postcss postcss-preset-env tailwindcss autoprefixer css-loader style-loader
+pnpm i -D @remotion/tailwind postcss-loader postcss postcss-preset-env tailwindcss autoprefixer css-loader style-loader
```
Just-Moh-it: Just wondering if we'd still need to ask users to install all the other deps 

```shell
pnpm i -D @remotion/tailwind postcss-loader postcss postcss-preset-env tailwindcss autoprefixer css-loader style-loader
```

instead of just

```shell
pnpm i -D @remotion/tailwind
```

since I saw all the other deps are `devDependencies` of the `@remotion/tailwind` package!

packages/docs/docs/tailwind/enable-tailwind.md
```diff
@@ -0,0 +1,42 @@
...
+import { Config } from "remotion";
+import { enableTailwind } from "@remotion/tailwind";
+
+Config.overrideWebpackConfig((currentConfiguration) => {
+  const newConfig = enableTailwind(currentConfiguration);
+
+  return {
+    ...newConfig,
+    // Make other changes
+  };
```
Just-Moh-it:
```suggestion
  const newConfiguration = {
  	// Make other change here
  	...currentConfiguration
  };

  return enableTailwind(newConfiguration);
```

I believe we should ask users to make their changes before applying these override functions (just like middlewares in `next.config.mjs` as instructed by their docs) because in the `enableTailwind` function, we spread the `currentConfiguration` passed as the first params first, and then add the overrides for tailwind!

So, for example, (going with the current version), a user overrides the `module` key in the newConfig, then the tailwind middleware won't work, and they'd need to manually spread the module value from `newConfig`.

Just-Moh-it (changes requested):
Haven't tested this locally, only on GitHub... just a few minor suggestions but the rest lgtm... great work 🔥!

packages/docs/docs/tailwind.md
rjackson: Yeah it seems necessary, unfortunately. I was hoping their presence in devDependencies would avoid the need to do so, but that doesn't seem to be the case.

These are the errors I get trying to run `packages/example` with just `@remotion/tailwind`:

```
$ pnpm start

> @remotion/example@3.3.87 start /Users/rjackson/Code/remotion/packages/example
> remotion preview --props src/my-props.json

Server ready - Local: http://localhost:3000, Network: http://192.168.86.27:3000
ERROR in ./src/Video.tsx 1:0-21
Module not found: Error: Can't resolve 'style-loader' in '/Users/rjackson/Code/remotion/packages/example'
ERROR in ../cli/dist/previewEntry.js 9:0-31
Module not found: Error: Can't resolve 'style-loader' in '/Users/rjackson/Code/remotion/packages/example'
Built in 1275ms
```

This is similar to how the `@remotion/skia` was set up, so I assume we had the same issue with that package too.

JonnyBurger: @rjackson @Just-Moh-it What if they were dependencies instead of devDependencies?
That would be totally valid!

rjackson: Darn I got excited, but that doesn't seem to work either.

I feel like it should work but I've had a whole prune, clean, reinstall, and rebuild, but it's still giving the same errors :(

```sh
$ pnpm store prune
$ pnpm clean
$ pnpm i
$ pnpm build
$ cd packages/example
$ pnpm start

> @remotion/example@3.3.87 start /Users/rjackson/Code/remotion/packages/example
> remotion preview --props src/my-props.json

Server ready - Local: http://localhost:3000, Network: http://192.168.86.27:3000
ERROR in ./src/Video.tsx 1:0-21
Module not found: Error: Can't resolve 'css-loader' in '/Users/rjackson/Code/remotion/packages/example'
ERROR in ../cli/dist/previewEntry.js 9:0-31
Module not found: Error: Can't resolve 'css-loader' in '/Users/rjackson/Code/remotion/packages/example'
Built in 2204ms
```

packages/docs/docs/tailwind/enable-tailwind.md
rjackson: Happy to take guidance on this, I don't feel like I've got firm opinions either way.

This example is near enough a copy from the `@remotion/skia` docs, so if we're happy updating it shall I do so in both places?

packages/docs/docs/tailwind.md
JonnyBurger: Ah damn 😐 Maybe `require.resolve("css-loader")`?
That should resolve the `css-loader` package relative to `@remotion/tailwind`!

If that does not work, then okay, let's ship it like that.

Just-Moh-it (dismissed):

JonnyBurger (approved):
Just as hyped as you are @Just-Moh-it!

Well done machine gun @rjackson 🫡 Thanks a lot! 🙂 🙌🏼

rjackson:
Ah you cracked the dependency issue, great stuff!

PR #2311, opened 2023-05-22
WIP: Easier metadata configuration for compositions

JonnyBurger:
- [x] Clicking on number and clicking outside will trigger a refresh
- [x] Should debounce or cancel the update requests
- [x] Should indicate whether is Suspense Loading, or loading `calculateMetadata()`
- [x] Dynamic duration feature makes it possible to go outside timeline
- [x] RenderModal uses currentComposition in a wrong way
- [x] ResolveCompositionConfig might break `wrap-remotion-context`
- [x] Handle error in Remotion Bundle overview page
- [x] Should be able to do sync


packages/core/src/resolve-video-config.ts
```diff
@@ -0,0 +1,61 @@
...
+		);
+	}
+
+	const durationInFrames =
+		calculated?.durationInFrames ?? comp.durationInFrames ?? null;
+	if (!durationInFrames) {
+		throw new TypeError(
+			'Composition durationInFrames was neither specified via the `durationInFrames` prop nor the `calculateMetadata` function.'
+		);
+	}
```
sauldom102: Awesomee! 🤩🤩🤩

Issue #2313, opened 2023-05-22
`@remotion/player`: RSC support (prepend `'use client'` while bundling to player files)


Just-Moh-it:
Currently, to use `<Player />` in react server components, we need to manually create a file with the `'use client'` directive, and that doesn't work well either, since all of remotion's source files don't have 'use client', so they get treated as server files by default. 

Even though there are very less use cases where a <Player /> could be used in a server component, it still makes sense for <Player /> to have the directive, in case it is ever imported in a server component.

### Workaround (sort of)

Currently, you could create a custom component with a name like `<UseClientPlayer />` that is a wrapper around `<Player />`. But creating that component with typescript requires some serious generics and other stuff to completely copy `<Player />`'s props.

Would be great if `@remotion/player` files could have the directive.

### Useful, maybe

https://github.com/rollup/rollup/issues/4699

JonnyBurger:
I'm down to add these directives, manually if needed!

JonnyBurger:
Let's put it into the wild!

/bounty 50

JonnyBurger:
💎 This issue has a bounty on it!

Read our contributing guidelines: 
- **How to contribute**: https://www.remotion.dev/docs/contributing/
- **Implementing a new feature**: https://www.remotion.dev/docs/contributing/feature
- **Implementing a new option**: https://www.remotion.dev/docs/contributing/option
- **Write documentation**: https://www.remotion.dev/docs/contributing/docs
- **Bounty information**: https://www.remotion.dev/docs/contributing/bounty

PR #2319, opened 2023-05-23
add "use client" directive to @remotion/player bundle

il3ven:
This PR adds a `use client` directive to @remotion/player. I have used `rollup-plugin-banner2` to add the directive to the ESM bundle. Note, this marks all components exported by @remotion/player as client components. I believe it isn't a problem in our case.

/claim #2313


JonnyBurger:
Seems to work excellently! Thanks a lot @il3ven! 🙌

PR #2321, opened 2023-05-24
prepend 'use client' while bundling to player files #2313

scshiv29-dev:
Instead of making any changes to the packages/player file took the suggestion from the Issue and created a packages/client-player ,used rollup-plugin-banner for the job and passed all the test as it was just a copy of the player files. Also added docs for reference.

/claim #2313

### DIFF
--- a/packages/client-player/.eslintrc
+++ b/packages/client-player/.eslintrc
@@ -1,0 +1,4 @@
+{
+	"extends": ["@jonny", "plugin:require-extensions/recommended"],
+	"plugins": ["require-extensions"]
+}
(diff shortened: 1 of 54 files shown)


JonnyBurger:
Hey @scshiv29-dev!
Sorry I just realized that there are two PRs about this. Apologies for assigning you and then merging the other one!

I'll make sure to establish something so that only one person can work on an issue at a time, otherwise there will be unnecessary competition and double effort.

That being said, I liked the other PR better because it keeps everything in one package. We would not want this type of duplication.

Sorry for the mess and hopefully you'll find another chance to contribute to Remotion!

scshiv29-dev:
No issues @JonnyBurger  i followed the issue and created another package nvm, next time maybe

PR #2334, opened 2023-05-26
add frameRange to rendermediaonlambda docs

UmungoBungo:

### DIFF
--- a/packages/docs/docs/lambda/rendermediaonlambda.md
+++ b/packages/docs/docs/lambda/rendermediaonlambda.md
@@ -116,7 +116,7 @@ _available from v3.2.40_
 
 Overrides default composition height.
 
-### `framesPerLambda`
+### `frameRange`
 
 Specify a single frame (a number) or a range of frames (a tuple [number, number]) to be rendered.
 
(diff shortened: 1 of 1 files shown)


packages/docs/docs/lambda/rendermediaonlambda.md
```diff
@@ -116,6 +116,14 @@ _available from v3.2.40_
 
 Overrides default composition height.
 
+### `framesPerLambda`
+
+Specify a single frame (a number) or a range of frames (a tuple [number, number]) to be rendered.
```
JonnyBurger: This seems wrong, this is the documentation for `frameRange?`

UmungoBungo: my bad 😵 the heading is wrong, it should be frameRange. I'll update the PR title as well!

JonnyBurger:
Thanks a lot! 🙌 

On this note, I realized, passing a single number to renderMediaOnLambda() is a bit awkward, since that will still give a video. So I made an edit to point people towards renderStillOnLambda() instead.

JonnyBurger (approved):

PR #2362, opened 2023-05-30
Consistent indication of optional props - rendermediaonlambda.md

UmungoBungo:
- Some of the props are optional in the code, yet they weren't marked as optional in the docs.
- Some props has the _optional_ subheading, but were missing the question mark.
- dumpBrowserLogs was using the <AvailableFrom /> component but no other headings were.


JonnyBurger (approved):

JonnyBurger:
Super valuable! Thanks a lot! 🙌

PR #2365, opened 2023-05-31
Json editor local validation

patsalv:
Needs more work. 
JSON editor no longer crashes when providing invalid input type, but now json and schema editor are no longer synced


JonnyBurger:
Thanks!
You are right, this state is too complicated.
In the end I decided to implement a simpler logic and be fine with losing the state if you change the tab. 3386d0eb903d681b3d62c8ffd2034f8dc2298cf8

PR #2366, opened 2023-05-31
dynamic JSONEditorModal text area

patsalv:


JonnyBurger:
Nice try 😛 but I found a way simpler solution:
https://github.com/remotion-dev/remotion/commit/201e4b77c0104118eb77692fb6167d019f3e22e7

PR #2369, opened 2023-06-01
Update deploysite.md

UmungoBungo:
Some optional arguments were missing question marks.
Sub-properties of options need to be H4s
Sub-properties of options don't need to marked as optional, as options itself is optional
Swapped region and siteName, as region is not optional
Consistently type out available from message, as opposed to using the AvailableFrom component


JonnyBurger:
Thanks! 🙌

PR #2370, opened 2023-06-01
Update getorcreatebucket.md

UmungoBungo:
Changed property to properties.

Not certain of whether you want `<AvailableFrom v="v3.3.78" />` or `_available from v3.3.78_`, but I do think the component was non-obvious:

image

If you want to stick with the component approach, I can make a PR that changes all the `_available from_` text across the docs. I'll also make sure the component sits next to the heading, as opposed to underneath. Just let me know! 😁

### DIFF
--- a/packages/docs/docs/lambda/getorcreatebucket.md
+++ b/packages/docs/docs/lambda/getorcreatebucket.md
@@ -36,7 +36,7 @@ Allows to pass a callback after the bucket was created and before the S3 website
 
 ## Return value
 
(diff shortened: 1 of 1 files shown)


JonnyBurger:
Hmm, we started out with italic text and are in a conversion to badges, which are nice since you can also click them.
If you want to help that'd be nice! 🙌🏼

For larger changes, please PR to v4 branch to avoid too many merge conflicts 🤓

PR #2371, opened 2023-06-01
it's vs its - grammar correction

UmungoBungo:
Something I had noticed whlie being in the docs 😁. From [grammarly.com](https://www.grammarly.com/blog/its-vs-its/);
> its (without an apostrophe) is a possessive pronoun, like his or her, for nouns that don’t have a defined gender. In contrast, it’s (with an apostrophe) is the shortened form, or contraction, of it is or it has.


JonnyBurger (approved):
I need to learn it, its terrible 😛
Thank you! 🙌

PR #2372, opened 2023-06-01
Update getsites.md

UmungoBungo:
adding some specificity around multiple buckets.


JonnyBurger (approved):
Makes sense! Thanks 🙌🏼

Issue #2376, opened 2023-06-01
Add `mutechange` event to player


pawelkontek:
# Feature Request 🛍️

A way to keep` playerRef.current.isMuted()` in a state 

## Use Case

...so I could display messages like: "please turn on sound" if the player is muted.

JonnyBurger:
Acceptance criteria:

- Implement a `mutechange` event similar to `volumechange` and `fullscreenchange` (https://www.remotion.dev/docs/player/player)
- Document it in the player.md

💎 This issue has a bounty on it!

Read our contributing guidelines: 
- **How to contribute**: https://www.remotion.dev/docs/contributing/
- **Implementing a new feature**: https://www.remotion.dev/docs/contributing/feature
- **Implementing a new option**: https://www.remotion.dev/docs/contributing/option
- **Write documentation**: https://www.remotion.dev/docs/contributing/docs
- **Bounty information**: https://www.remotion.dev/docs/contributing/bounty

/bounty 100

rjackson:
Happy to have a look at this

/attempt #2376

JonnyBurger:
Thanks! Good luck @rjackson 🙌🏼

This issue is taken, please no further attempts 🙅‍♂️

PR #2378, opened 2023-06-01
Add `mutechange` event to Player

rjackson:
This PR adds a new `mutechange` event emitter for Player.

It's implemented the same as the `isMuted()` method, where the player is considered muted if the mute button is pressed or the volume is currently set to zero.

Alongside this PR I have included a fix for the `volumechange` event, to make sure it is emitted every time the volume changes. Previously it was only emitting were programmatically changed via the `setVolume()` method, which means it was not emitting when the volume slider was changed.

To test this change I have expanded the `packages/player-example`. When running this example with:

```sh
cd packages/player-example
pnpm dev
```

And interacting with the mute button or volume slider, you can see the mutechange and fixed volumechange events being logged.

In the following screenshot, the left player shows mute being directly toggled, whereas the right shows the volume slider going down to zero and back up.

image

The Player API reference docs have been updated to note the new event:

image

Solves #2376
/claim #2376

### DIFF
--- a/packages/docs/docs/player/api.md
+++ b/packages/docs/docs/player/api.md
@@ -597,6 +597,10 @@ useEffect(() => {
     console.log("scalechange", e.detail.scale);
   };
 
+  const onMuteChange: CallbackListener<"mutechange"> = (e) => {
+    console.log("mutechange", e.detail.isMuted);
+  };
+
   playerRef.current.addEventListener("play", onPlay);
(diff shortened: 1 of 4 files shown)


JonnyBurger (approved):

JonnyBurger:
Looks absolutely perfect! Thanks a lot. 

Thanks for the bugfix as well, well caught! /tip 20

PR #2381, opened 2023-06-02
fix @remotion/tailwind when used in ESM

JonnyBurger:


mcnaveen:
Thanks for the quick fix.

Issue #2153, opened 2023-04-26
Text To Speech Demo Does Not Work


odapplications:
# Bug Report 🐛 When connected to Azure and AWS it creates a 0 byte mp3 file and breaks because it can not be played. Using default code and presets. Just updating the .env 

![error-results](https://user-images.githubusercontent.com/27908286/234511274-3129af1f-700e-4458-a7ce-c41f3c76dcb6.jpg)

Note: I also tried in chrome not brave and still same results.

thecmdrunner:
Although v4 is not stable yet, and isn't recommended for production, you could try the new Google Text to speech tempalate and see if that works?

https://github.com/thecmdrunner/remotion-gtts-template

JonnyBurger:
Thanks for reporting!
To make it easier for us to debug - do you have any idea why it might have happened? Does the file exist? Does it maybe take some time to generate the audio and Remotion tries to render too early? Any error in the S3/Azure logs?

JonnyBurger:
OK, we have tried to reproduce this today ourselves, and indeed we get `StatusCode: 1006, undefined Reason:  Unexpected server response: 401` even for an existing key that worked before.

@FelippeChemello Do you have any insights, did it break for you as well recently? Did you do something to fix it?

FelippeChemello:
Not yet, however I haven't updated my deployment recently.

JonnyBurger:
OK, I improved the whole situation! Sorry for taking this long to take a look at it.
There were multiple issues leading to bad developer experience:

- Azure tokens don't become immediately valid, you have to wait, maybe 1-2 hours even? At least this was the case for us, but the account had billing enabled very freshly. --> Mention this now in the README
- If the Synthesis fails with an invalid token, it would still try to upload to S3. --> Fixed this in the code
- Remotion environment variables would not auto-reload as we suggest in the terminal --> Fixed this in Remotion.

Hope that in the future, this leads to less problems!

PR #2389, opened 2023-06-05
Hide TTS template

JonnyBurger:
I cannot use the TTS template anymore, the Azure service is just responding anymore. This is after another report we have received https://github.com/remotion-dev/remotion/issues/2153.

It is currently not clear whether this is a problem with our accounts or the API does not work generally anymore. Until it is clear what the problem is, we are hiding the template because people might have a bad experience with it.


JonnyBurger:
Ok, now it works actually, we just had to wait really long until the tokens start working

PR #2406, opened 2023-06-09
Update getting-started.md

ideabrian:
just playing with the language. Feel free to reject anything you don't like.


JonnyBurger:
Love it! Less words is always better in documentation 🤝

PR #2416, opened 2023-06-15
Remove FredokaOne from font-picker.md

UmungoBungo:
Fredoka One is no longer exported, changed it over to Fredoka

### DIFF
--- a/packages/docs/docs/font-picker.md
+++ b/packages/docs/docs/font-picker.md
@@ -281,8 +281,8 @@ export const top250 = [
     load: () => import("@remotion/google-fonts/FrankRuhlLibre"),
   },
   {
-    family: "Fredoka One",
-    load: () => import("@remotion/google-fonts/FredokaOne"),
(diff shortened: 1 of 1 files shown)


UmungoBungo:
image

This was causing builds to fail on the CloudRun docs. I assume this didn't occur on main repo of Remotion because Vercel used a cached version of google-fonts? If that is the case, how come it didn't detect a change in the package.json file?

JonnyBurger:
This list was programmatically generated, made a PR to update everything and added instructions now 👉🏼 #2418 

This is a good question! Turborepo should consider this:

> The contents of each workspaces's package.json and turbo.json (if it exists) are automatically considered for workspace hashes. This behavior is not configurable.

So I think it is a Docusaurus Twoslash cache which we don't specify correctly in our turbo.json 🤔 we actually have no configuration, so I think it will just cache it globally. Weird, I do not know the right thing to do right now.

PR #2420, opened 2023-06-17
change the referenced package for calls to remotion lambda

alexfernandez803:
use the new package name for the example


JonnyBurger:
Thanks a lot for this persistence! 🙌

Issue #2421, opened 2023-06-18
--force-bucket-name not working with CLI commands


UmungoBungo:
Looking at the code for removing a site with the cli (`src/cli/commands/sites/rm.ts`) it first gets a list of deployed sites, but the code within only gets sites from buckets with the remotionlambda- prefix. If you are forcing the bucket name, it may not have that prefix.
```tsx
const region = getAwsRegion();
const deployedSites = await getSites({
    region,
});
for (const siteName of args) {
    const bucketName =
        parsedLambdaCli['force-bucket-name'] ??
        (await getOrCreateBucket({region})).bucketName;
    const site = deployedSites.sites.find((s) => s.id === siteName.trim());
    if (!site) {
        Log.error(
            `No site ${siteName.trim()} was found in your bucket ${bucketName}.`
        );
        return quit(1);
    }
```

I also found that the create command also doesn't work with --force-bucket-name. This is due to options.bypassBucketNameValidation not being set from the CLI. 
```bash
pnpm exec remotion lambda sites create --force-bucket-name=not-remotion-name
```
![image](https://github.com/remotion-dev/remotion/assets/22192773/80e07bba-40f1-4cf6-8c3e-d3f2810daa4a)

JonnyBurger:
Very valid! Thanks for reporting.

/bounty 50 for the person who can fix this up:

1. Allow to remove a site with `--force-bucket-name` even if no sites exist with the default bucket prefix
2. Pass `bypassBucketNameValidation` in CLI when a bucket name is forced

As mentioned in the description above.

💎 This issue has a bounty on it!

Read our contributing guidelines: 
- **How to contribute**: https://www.remotion.dev/docs/contributing/
- **Implementing a new feature**: https://www.remotion.dev/docs/contributing/feature
- **Implementing a new option**: https://www.remotion.dev/docs/contributing/option
- **Write documentation**: https://www.remotion.dev/docs/contributing/docs
- **Bounty information**: https://www.remotion.dev/docs/contributing/bounty

tachyonicbytes:
I'll try. You can assign me, if you want

JonnyBurger:
@tachyonicbytes Thanks, good luck! 🙌🏼

tachyonicbytes:
I think it should be done now. Please let me know if there is anything left to add.

PR #2423, opened 2023-06-18
fix: #2421. Propagate --force-bucket-name in rm.ts and get-sites.ts

tachyonicbytes:
Solves #2421.

There was a bug where `--forced-bucket-name` was not propagated properly.

/claim #2421


JonnyBurger (approved):
Thanks for the quick PR! 🙌

Feedback: We use objects as the only parameter for public APIs and I renamed `forced` to `forceBucketName` to align the API across the board.

But given the sparse description from our side, great!

PR #2462, opened 2023-07-01
Go client lambda response update

alexfernandez803:
I've noticed that the lambda response signature, maybe related to the aws-lambda-go package update.
This MR fixes this issue and including update of example. Version is also added in getting progress values to prevent this issue

{"errorType":"Error","errorMessage":"Version mismatch: When calling getRenderProgress(), you called the function remotion-render-3-3-101-mem2048mb-disk2048mb-240sec which has the version 3.3.101 but the @remotion/lambda package is an older version. Deploy a new function and use it to call getRenderProgress(). See: https://www.remotion.dev/docs/lambda/upgrading","trace":["Error: Version mismatch: When calling getRenderProgress(), you called the function remotion-render-3-3-101-mem2048mb-disk2048mb-240sec which has the version 3.3.101 but the @remotion/lambda package is an older version. Deploy a new function and use it to call getRenderProgress().


JonnyBurger:
Oh, very well spotted! Thanks a lot for fixing this.
/tip 50

PR #2490, opened 2023-07-04
Fix typo of Parametrized to Parametrized

alexfernandez803:
Fix typo of Parametrized to Parametrized


JonnyBurger:
Mind blown 🤯

Thanks for fixing that!

PR #2498, opened 2023-07-05
`@Remotion/renderer` Console.log incomplete objects indicator

patsalv:
Tests passing for objects containing only strings, probably more testing needed before merging


JonnyBurger (changes requested):
Try to also log this:

```
console.log(new Array(1000).fill(true))
```

Probably we need to handle arrays as well.

Let's use `…` instead of `...` (different character) because if we resemble the Chrome DevTools maybe people will find it more intuitive

JonnyBurger (approved):

PR #2507, opened 2023-07-05
`@remotion/renderer`: Filter out mixed content warnings for localhost frames

patsalv:


JonnyBurger (approved):
Don't forget to reference the issue number and prefix it with "Fix"

JonnyBurger (approved):

PR #2513, opened 2023-07-06
Update rendermediaonlambda.md

UmungoBungo:
Number of returned properties different from the text description. Not 100% certain that my rewording is achieving what you intended.


JonnyBurger:
Thanks for catching that!

PR #2514, opened 2023-07-06
Update post.mjs - discord supports markdown

UmungoBungo:
Completely optional PR, and I haven't tested it either 🤪 but thought it would be nice to make this simpler, and let github and discord do the hard work!


JonnyBurger:
Thanks! I already tried this and I did not like it – I think either titles were unreasonably huge or not supported at all.
So there was some intention in replacing it with bold text.

PR #2525, opened 2023-07-07
Fix typo in pnpm-workspace.yaml

nmattia:
Spotted a typo in the pnpm workspace file while browsing the source; might be intentional in which case feel free to close


JonnyBurger:
Indeed! I'm trying to ignore this directory, but sometimes it ends up in our lockfile, thanks for catching that! Hope that leads to less changes in the lockfile. 

🙌🏼

PR #2558, opened 2023-07-10
`useDelayRender()` snippet

satelllte:
## About

`delayRender()` as a hook

I'm finding it useful and I'm actually already using this one internally in my projects, for instance: https://github.com/satelllte/remotion-audio-visualizer/blob/main/src/hooks/useDelayRender.ts

## Docs preview

https://remotion-git-fork-satelllte-use-delay-render-remotion.vercel.app/docs/use-delay-render


satelllte:
Hi @JonnyBurger 👋🏻 
Just tagging you in case you missed it. Does this change make sense?

JonnyBurger:
@satelllte Sorry for the long processing time of this PR. It looks very clean, but I will need to think about whether we need this API in core. In any case, this is very useful and we should definitely make a snippet out of it!

satelllte:
@JonnyBurger cool! So then up to you to decide :)

JonnyBurger:
Sorry for the long wait and thanks a lot for this useful function!

We recognize the core APIs are verbose and there is opportunity for syntactic sugar, but adding more core APIs can also get more overwhelming and over time will lead to increased bundle size.

So for now I added it to the snippets section where people can copy-paste it. We'll consider making a separate package for utility functions in the future!

satelllte:
Nice! Thanks, @JonnyBurger 🤝

PR #2568, opened 2023-07-12
`@remotion/cli`: Allow saving the input props using Cmd+S indicate unsaved changes

### LINKED ISSUE #2067: Consider: Save a field in a schema by pressing Cmd+S / Ctrl+S

### LINKED ISSUE #2496: Visually indicate in tab that there are unsaved prop changes

patsalv:
Fix #2496 Fix #2067


JonnyBurger:
Some things I noticed:

- [x] it should not display a click cursor, because there is no action there
- [ ] There is no graying out while the props are saving, like with the other icons
- [x] We should put a tooltip about how to save
- [ ] If `!process.env.KEYBOARD_SHORTCUTS_ENABLED`, let's not show the tooltip
- [ ] Let's add a notification "Default props saved" (notificationCenter.current.addNotification)

Issue #2573, opened 2023-07-12
"Jump to Beginning" button should jump to "Mark In" frame if active


BrodyRas:
# Bug Report 🐛
When the "Mark In" frame is active, the "Jump to Beginning" button should jump to that frame, instead of the actual first frame of the composition.

## Steps to Reproduce
- Open a composition
- Use the "Mark In" button on a frame that isn't the first frame of the composition
- Press the "Jump to Beginning" button
- Note how the playhead is now outside the active frames

![Jul-12-2023 10-16-50](https://github.com/remotion-dev/remotion/assets/49171024/b48e1f1c-b5ae-43cf-842d-68d2f7e2501e)

JonnyBurger:
Good point!

mnmt7:
@JonnyBurger, can I have a go at this?

JonnyBurger:
Happily! @mnmt7

PR #2581, opened 2023-07-13
fix: "Jump to Beginning" button should jump to "Mark In" frame if active

mnmt7:
This PR solves #2573. When the user presses on "Jump to Beginning" Button or presses "a" key, the pointer jumps to the "Mark In" frame if it is active instead of the first frame of the composition. Similarly, when the user presses "e" key, the pointer moves to the "Mark Out" frame if it is active instead of the last frame of the composition.
 
video - https://jam.dev/c/6dd9913a-0d57-4d50-8eee-6889ad3c5c81


JonnyBurger:
Thanks a lot! Looks right to me, I will also try to merge your other PR today!

JonnyBurger (approved):
🙌🏼

PR #2583, opened 2023-07-16
Add redirection for booking

MehmetAdemi:
Instead of using "cal.com/remotion", let's also have "remotion.dev/booking". I would like to use it on our Google Search profile.


JonnyBurger:
Sounds good to me!

PR #2592, opened 2023-07-18
Add booking link on expert profile

MehmetAdemi:
The booking link was only on the expert card on the exerts landing page. I added the booking link on the expert profile as well.


JonnyBurger:
Excellent! 💡

PR #2595, opened 2023-07-18
Add some hover effect to the expert card

MehmetAdemi:
I was thinking of adding a button to see the details from a Remotion Expert. But I think it's better to signalize that you can press the card with a hover effect. Feedback is appreciated.


JonnyBurger:
Nice!

PR #2631, opened 2023-07-25
`@remotion/docs` Mention in lambda docs that --gl=angle isn't supported in lambda

patsalv:


JonnyBurger:
It is supported on Lambda, but breaks WebGL. Can we express this more precisely?

PR #2649, opened 2023-07-31
Remotion Studio: Fix layout of frame range slider

### LINKED ISSUE #2647: Frame range slider is off
image

If you are **not** using Tailwind

patsalv:
fix #2647 

Only draft, since we should double check that
```js
corePlugins: {
		preflight: false,
	},
```

in the `tailwind.config.js` doesn't have implications on other styles we don't want!
Just to be careful :)


JonnyBurger:
Thanks a lot! 🙌 

The ultimate goal is that it looks good no matter which Tailwind config we use.
It all looks a bit random to me at the moment, merging for now, in case we get problems, then we'll make another issue :^)

PR #2651, opened 2023-07-31
`@remotion/cli`: npx remotion upgrade optionally takes --version flag

### LINKED ISSUE #2632: Allow `npx remotion upgrade` to select a specific version
Such as allowing to downgrade, or to select a alpha product such as Cloud Run.
Good idea by @jmaguirrei

patsalv:
close #2632

### DIFF
--- a/packages/cli/src/upgrade.ts
+++ b/packages/cli/src/upgrade.ts
@@ -42,8 +42,16 @@ export const upgrade = async (
 		packageJson.optionalDependencies ?? {}
 	);
 	const peerDependencies = Object.keys(packageJson.peerDependencies ?? {});
-	const versionToUpgrade = version ? version :  await getLatestRemotionVersion();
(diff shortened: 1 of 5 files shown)


packages/cli/src/upgrade.ts
```diff
@@ -45,7 +44,6 @@ export const upgrade = async (
 	const peerDependencies = Object.keys(packageJson.peerDependencies ?? {});
 	const latestRemotionVersion = await getLatestRemotionVersion();
```
JonnyBurger: Can you make it that we do not even fetch the latest version if you anyway provide a custom version? It will be faster.

```suggestion
	const versionToUpgrade = version ? version :  await getLatestRemotionVersion();
```

JonnyBurger (approved):
Thanks!
For next time, docs would also be nice 🙂

PR #2682, opened 2023-08-04
add Remotion Expert to directory

MehmetAdemi:
Don't understand why there is a change from "GL Options" to "--gl flag" in articles.ts, but I just left it like this.


JonnyBurger:
Thanks! Indeed, that is my fault, thanks for updating this!

PR #2736, opened 2023-08-15
Use code block for Lambda policies validate in setup.md

UmungoBungo:
This is ideal so the copy prompt appears on hover.


JonnyBurger (approved):
Thanks!

PR #2745, opened 2023-08-16
fix: Prevent exiting the page if there are unsaved changes to the default props

### LINKED ISSUE #2495: Prevent exiting the page if there are unsaved changes to the default props
If the Remotion Studio Props Editor has unsaved changes, and you are about to reload, then an alert should be triggered using `window.onbeforeunload` telling the user that there are unsaved changes.

To determine if there are unsaved changes, check only if the root object is changed.

Once the changes are saved, the handler should be removed.

**Special case:** 

There are 5 cases where Remotion Studio will reload itself (look for `window.location.reload()`):

- New input props or env variables (packages/cli/src/event-source.ts) ❌
- Fatal React error (packages/cli/src/preview-server/error-overlay/entry-basic.ts) ✅
- Hook order changed (packages/cli/src/preview-server/error-overlay/react-overlay/listen-to-runtime-errors.ts) ❌
- Failed fast refresh (packages/cli/src/preview-server/hot-middleware/process-update.ts) ❌

✅ The error is unrecoverable and we anyway lost our changes, so we can refresh
❌ The error is not recoverable, and we should not refresh, instead show a notification using `notificationCenter.current?.addNotification`

/bounty 100

rishi-raj-jain:
fix #2495
/claim #2495


rishi-raj-jain:
@JonnyBurger 

Can you give the message and duration you expect to set on this?

I added the event listener in the render modal, let me know if you think I shall shift it to somewhere else.

packages/cli/src/event-source.ts
```diff
@@ -32,7 +35,16 @@ export const openEventSource = () => {
 			newEvent.type === 'new-input-props' ||
 			newEvent.type === 'new-env-variables'
 		) {
-			window.location.reload();
+			if (window.unsavedProps) {
+				notificationCenter.current?.addNotification({
+					id: 'random',
+					content: 'Do not save hey',
```
JonnyBurger: - The content could be "There are unsaved changes to the props. Do you want to leave?"
- The notification center will do nothing here, you can only prevent a page from closing if you use `window.oneforeunload`

Example:
```
window.addEventListener('beforeunload', (event) => {
  event.returnValue = 'Are you sure you want to leave?';
});
```

packages/cli/src/editor/helpers/document-title.ts
```diff
@@ -6,12 +6,22 @@ let unsavedProps = false;
 let tabInactive = false;
 let renderJobs: RenderJob[] = [];
 
+declare global {
```
JonnyBurger: All global variables are defined in core/src/index.ts.
Can you move it there as well and update the variable name with a `remotion_` prefix?

JonnyBurger (changes requested):

JonnyBurger:
Thanks!

JonnyBurger (approved):

Issue #2703, opened 2023-08-07
Cloud Run: Bad error handling if a render is failing


JonnyBurger:
Currently, you have to inspect the logs to see the error, it does not propagate back to renderMediaOnCloudRun()

JonnyBurger:
Fixed by #2764

PR #2764, opened 2023-08-21
`@remotion/cloudrun`: Propagate back render errors

UmungoBungo:
Issue: #2703


UmungoBungo:
Development image available here: `us-docker.pkg.dev/remotion-dev/development/render:4.0.21`
You will need to copy+paste this url when deploying a new cloud run service, and give the service a name like `remotion-4-0-21dev-mem2gi-cpu1-0-t300`

image

UmungoBungo:
I tried a bunch of times, but I wasn't able to get the CLI error message to overwrite after Symbolicating was finished, so both messages remain in the CLI;

image

Not sure why this is happening as I am calling the Symbolicate code the same way that the Lambda package is 🤷🏻 
```ts
const frames = RenderInternals.parseStack(error.stack?.split('\n') ?? []);

const errorWithStackFrame = new RenderInternals.SymbolicateableError({
	message: error.message,
	frame: null,
	name: error.name,
	stack: error.stack,
	stackFrame: frames,
});
await CliInternals.handleCommonError(
	errorWithStackFrame,
	ConfigInternals.Logging.getLogLevel()
);
```

JonnyBurger (approved):
Excellent! Did not look at it too closely, since it is alpha, let's ship it and see if it holds up!

PR #2769, opened 2023-08-22
fix typo in durationInMiliseconds

UmungoBungo:
durationInMiliseconds is missing the second `l`

I wasn't 100% certain if you would want the changelog updated, as the spelling is correct for the changelog date 🤷🏻


JonnyBurger:
Thanks! 

For backwards compatibility, we need to support the previous, added that now.

JonnyBurger:
Lol thx!

PR #2772, opened 2023-08-22
Run dbus in docker

JonnyBurger:
TODO: If works well, also add it to https://remotion.dev/docs/docker


JonnyBurger:
While it gives a warning, we find that it is not needed at all.

PR #2782, opened 2023-08-23
Update Remotion Expert page

MehmetAdemi:
Added a website button and made some adjustments for alignment.

What needs to be done is handling the case when the browser window is narrow. Idea: organize buttons in three rows instead of only 2 rows when the window narrows.


MehmetAdemi:
I think it should be fine for now. @JonnyBurger can you check and give feedback?

JonnyBurger:
Yes 👏 

Looks excellent!

PR #2784, opened 2023-08-23
`@remotion/renderer`: Better error handling for renders

patsalv:


JonnyBurger (changes requested):
Some Lambda code which gets bundled now imports renderer code, which is causing it to break. Put it in `BrowserSafeApis` instead of `RenderInternals`.

JonnyBurger:
Sorry, too many merge conflicts, I recreated it here https://github.com/remotion-dev/remotion/pull/2789, will close

PR #2795, opened 2023-08-25
feat: check first if type of window is undefined in isIosSafari

Iamshankhadeep:


JonnyBurger:
Thanks Shank! 🙌 Very important!

PR #2810, opened 2023-08-28
`@remotion/cli`: Remotion Studio now warns if you are on a Remotion version with known bugs

patsalv:


JonnyBurger:
Merging later, I still want to bring in a feature to keep the bugs in the main repo

PR #2826, opened 2023-08-31
Update brownfield-installation.md

omercelik:
Using remotion as the folder name causes ```Module has no exported member``` errors as it is looking into that folder instead of node_modules/remotion folder.

### DIFF
--- a/packages/docs/docs/brownfield-installation.md
+++ b/packages/docs/docs/brownfield-installation.md
@@ -119,6 +119,10 @@ registerRoot(RemotionRoot);
 
 The file that calls [`registerRoot()`](/docs/register-root) is now your Remotion **entrypoint**.
 
+:::note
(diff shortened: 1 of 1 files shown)


JonnyBurger:
That is only if you have configured a TypeScript alias or baseUrl if I understand correctly, then "remotion" will resolve to the folder..

Having the folder called remotion is a convention now and the CLI will pick it up as a commonly known entry point. So I would rather warn about the TS config.

Is this a specific framework which has this as the default? Indeed this conflict is unfortunate!

JonnyBurger:
Appreciate it!

I wanted to verify that this is still the case, and the latest Next.js scaffold does not generate a config anymore which resolves `remotion` to the remotion folder.

Given that `remotion/index.ts` is in the list of known root file paths, I prefer the existing strategy but mentioning this as a pitfall.

PR #2828, opened 2023-08-31
Update incorrect 'see also' links in compositions.md

UmungoBungo:
One of the links was for the parent page, and the other link was incorrect


JonnyBurger (approved):
Thanks!

Issue #2747, opened 2023-08-17
Allow to display assets + outputs of compatible formats in the Remotion Studio


JonnyBurger:
For now, just if you make a video, we could allow the output video to directly play in the Remotion Studio, similar to when you play a video in VS Code. 

You could click on an asset in the "Assets" panel and it would show in place of the composition, the same for the output video

PR #2839, opened 2023-09-04
Remotion Studio: Assets from the public folder can be displayed in the Remotion Studio

patsalv:
Partially completes #2747


JonnyBurger:
Audio files would look better if centered:

Screenshot 2023-09-05 at 09 50 39

JonnyBurger:
The URL does get set, but then if you reload with Cmd+R, you do not land in the asset again

JonnyBurger:
A selection of great TODOS that you can pick from in case you are earlier at the office tomorrow than me @patsalv :

- [ ] Display a loading indicator when while the asset dimensions are loading (packages/cli/src/editor/components/Spinner.tsx)
- [ ] Make back button / forwards button work also when switching between assets and compositions (pushState / popState) implementation
- [ ] If URL suggests an asset (/assets/hi.png), but does not exist, show an error message
- [ ] If asset cannot be displayed (bangers.woff), show just the filename and file size
- [ ] Recursive folder expand when switching asset via URL

PR #2842, opened 2023-09-04
add internalRenderMediaOnCloudrun

UmungoBungo:
By creating internalRenderMediaOnCloudrun, found that these attributes were missing:

- colorSpace
- offthreadVideoCacheSizeInBytes
- x264Preset

Should be more reliable to add/remove properties in the future!


UmungoBungo:
I'm not sure about explicitly listing `offthreadVideoCacheSizeInBytes` in the `InternalRenderMediaOnCloudrun` type as it is missing from the lambda equivalent `RenderMediaOnLambdaInput`, and I _think_ it is covered off by the Partial..?  
image

JonnyBurger:
Excellent, thanks a lot for bringing this pattern also to Cloud Run!

PR #2874, opened 2023-09-09
Add `bun` package manager support

### LINKED ISSUE #2873: Add support for `bun` package manager when used for project init
# Feature Request 🛍️

When creating a Remotion project using `bunx create-video@latest`, it should use `bun` instead of `npm` as the package manager.

image

## Use Case

- Use `bun` for installing packages, just like `yarn, npm, pnpm` when a new remotion project is initialized.

- would save 4 seconds of life for everyone who uses `bun` to create a Remotion project.

## Possible Solution

#2874

thecmdrunner:
fix #2873

1. Check if bun was used to initialize the project: `bunx create-video@latest`
2. Include bun in functions that return commands
3. Handle installation via bun
4. Add tests for bun

Tested `bun` on these templates, and works on:

- [x] Tailwind template (Fixed duplicate dependencies: https://github.com/remotion-dev/template-tailwind/commit/7361decac6b1b14781dbe7285b3770c80560b5e2)
- [x] Remix (Some err)
- [x] Still Template
- [x] Three Fiber
- [x] Blank
- [x] Next /pages
- [x] Next /app
- [x] Hello World TS
- [x] Hello World JS
- [x] ... add more


JonnyBurger:
Thanks a lot! We will merge this when all the features work properly (especially the studio).
I'll file bugs with bun to ensure they are aware of the crashes.

thecmdrunner:
@JonnyBurger this PR only adds bun package manager support, which by default uses the node runtime for studio and renders. Since the bun package manager is decoupled with bun runtime, this seems to work flawlessly. Can we not have just the package manager support rn?

JonnyBurger:
@thecmdrunner Upon thinking about it further, I agree with you!

We can support Bun as a package manager before we support it as a runtime. Will test all templates and then merge this!

JonnyBurger (dismissed):

JonnyBurger (approved):

PR #2885, opened 2023-09-12
Fix: Thorium browser docs link

thecmdrunner:
"What's that new throium browser that Remotion is using now? I wanna kno... oh...oh no" 🤣

image


JonnyBurger:
@thecmdrunner Thanks a lot for catching that!
Also added a redirect so the old URL now starts working.

PR #2901, opened 2023-09-18
fix filter on bucket which even if non relevant buckets are shown

alexfernandez803:
Fix issue on cloudrun which I have an existing project with some buckets that does not start with "remotioncloudrun", still considered as remotion bucket.
`
 You have multiple buckets (gcf-v2-sources-XXXXXXX-australia-southeast2, gcf-v2-uploads-XXXXXXX-australia-southeast2, remotioncloudrun-592i3qw5fz) in your Cloud Storage region (australia-southeast2) starting with "remotioncloudrun-"`

Existing bucket does not start with "remotioncloudrun". This PR fixes this issue.


JonnyBurger:
Thanks a lot! 🙌

PR #2939, opened 2023-09-27
`@remotion/eslint-config`: Disable "no-unreachable" rule in TypeScript

JonnyBurger:
Because TypeScript takes care of this already

https://eslint.org/docs/latest/rules/no-unreachable#handled_by_typescript


JonnyBurger:
Ah, this is not the curlprit

Issue #2942, opened 2023-09-28
Add a `playbackRate` prop to the <Gif> component


JonnyBurger:
Could just divide the current frame with the `playbackRate` to make the GIF play faster or slower.

And then document the new prop in gif.md!

JonnyBurger:
Here is the file to get started in

https://github.com/remotion-dev/remotion/blob/main/packages/gif/src/useCurrentGifIndex.tsx

/bounty 100


💎 This issue has a bounty on it!

Read our contributing guidelines: 
- **How to contribute**: https://www.remotion.dev/docs/contributing/
- **Implementing a new feature**: https://www.remotion.dev/docs/contributing/feature
- **Implementing a new option**: https://www.remotion.dev/docs/contributing/option
- **Write documentation**: https://www.remotion.dev/docs/contributing/docs
- **Bounty information**: https://www.remotion.dev/docs/contributing/bounty

anikdhabal:
@JonnyBurger  Hey can you assign me this issue?

JonnyBurger:
@anikdhabal @DevGajjar28 

Whoops, that was almost at the same time! @anikdhabal was first, so I will give them this issue.

Sorry @DevGajjar28, more bounties coming soon!

Vishal-Sikka:
oh forgot to comment attempt here, i already figured it out

DevGajjar28:
I have also figured out but my bad it was assigned

JonnyBurger:
Sorry @Vishal-Sikka, we will for the bounty only consider a PR by the assignee. 

We want to avoid multiple people doing the same work. Nonetheless thanks for your willingness to contribute!

Vishal-Sikka:
@JonnyBurger it's ok

anikdhabal:
@Hitesh-sisara As you can see, this issue is assigned to me, and I'm working on it. @JonnyBurger clearly mentioned the guidelines.

JonnyBurger:
This is right, @anikdhabal you are eligible for the bounty. We are going to continue process and also work with Algora to make this more smooth.

PR #2943, opened 2023-09-28
fix:#2942 Added a playbackRate prop to the <Gif> component

Vishal-Sikka:
fix: #2942 
/claim #2942

Added playbackRate prop documentation to <Gif> component.
and also written doc


JonnyBurger:
Hi! This PR has few problems:

- Does not actually work, the function call `useCurrentGifIndex` is not actually updated, clearly not tested
- Messed up formatting
- Docs mention old version and mentions example which does not follow

While I appreciate the willingness to contribute, we cannot consider sloppy PRs for which we have to fix everything ourselves before merging.

PR #2970, opened 2023-10-04
`@remotion/cloudrun`: Credentials refactor towards supporting Google Cloud Tasks

alexfernandez803:
The goal of this Merge Request is to enable the execution of renderMediaOnCloudrun() on [Google Cloud Tasks](https://cloud.google.com/tasks). Cloud Tasks allows the rendering process to be executed as a separate server and provides queuing functionality.

Upon my investigation, I found that renderMediaOnCloudrun() currently relies on the following properties:

process.env.REMOTION_GCP_PROJECT_ID
process.env.REMOTION_GCP_CLIENT_EMAIL
process.env.REMOTION_GCP_PRIVATE_KEY

However, using these properties won't suffice for service-to-service communication, specifically between Cloud Tasks and renderMediaOnCloudrun(). Google Cloud offers a solution to enable communication between services using a service account, facilitating communication between Cloud Tasks and renderMediaOnCloudrun(), which runs on Cloud Run.

To determine if the rendering process is occurring within a Google Cloud Task, I have implemented the following check:

```
export const isInCloudTask = () => {
	return (
		process.env.K_CONFIGURATION &&
		process.env.GCLOUD_PROJECT &&
		!process.env.REMOTION_GCP_PROJECT_ID
	);
};
```

These properties are commonly present when the code is running in a Cloud Task environment.

If the code is running within a Cloud Task, we only need to provide the project ID. During execution, the service account associated with the services will automatically provide the necessary credentials for Cloud Tasks to execute renderMediaOnCloudrun(). Here's the code snippet for this scenario:

```

return new ServicesClient({
	projectId: getProjectId(),
});

```
To add to this, there is a challenge in google cloud task that when you pass REMOTION_GCP_PRIVATE_KEY as one of environment property it will not work, 

Note that I have not provided the unit testing for this yet. This is for your feedback.


JonnyBurger:
This is exciting! Thanks for doing this!

alexfernandez803:
@JonnyBurger  thanks for approving, based on my test this is good.

PR #2979, opened 2023-10-06
Fixes #2932 - Allow a toneFrequency prop to be passed to <Audio> for changing the pitch.

### LINKED ISSUE #2932: Allow a `toneFrequency` prop to be passed to `<Audio>` (Change pitch of Audio)
Allow to change the pitch (technically named "tone frequency") of an audio.

## Desired API

`<Audio toneFrequence={0.9} />` 

When calling [`registerRenderAsset()`](https://github.com/remotion-dev/remotion/blob/8fe522de3e06268ed576ac2820c010fb54de8bd3/packages/core/src/audio/AudioForRendering.tsx#L91), this information gets passed and when it arrives at preprocessAudioTrack, then a FFmpeg filter will be applied. 

For example `<Audio toneFrequency={0.9} />` could result in a `asetrate=44100*0.9,aresample=44100,atempo=1/0.9` filter being applied to the audio.

Filter taken from this answer: https://stackoverflow.com/questions/53374590/ffmpeg-change-tone-frequency-keep-length-pitch-audio

## Acceptance criteria

- Should implement a way to change the pitch of the audio during rendering.
- Should prevent changing the pitch over time (we do not need to support this right now)
- Add it to the `<Audio>` docs
- Mention limitations in docs: Must have static value per audio (no change over time), will only work during rendering and not in preview
- Implement min / max values that seem rational to your judgement

## Bounty

💎 This issue is part of Hacktoberfest and has a bounty on it!
- **General Rules**: https://www.remotion.dev/bounties

🎃 **Hacktoberfest rules**
- You can only work on 1 issue at a time.
- Only 1 person can work on the same issue.
- We explicitly assign a developer. Don’t work on an issue if we do not assign you.
- First-time contributors will get preferential treatment.
- You must set up the Remotion repository locally and test your changes. [Contributing to Remotion | Remotion | Make videos programmatically in React](https://www.remotion.dev/docs/contributing)
- Your PR must fulfill the acceptance criteria.
- If you use AI, you must verify that the output is correct.
- All tests (functionality and formatting) must pass.

Issue already taken? Find more issues under https://remotion.dev/bounties

/bounty 200

evoxf1:
### Issue

Fixes #2932

### Description

This PR addresses issue #2932 by introducing the `toneFrequency` prop to the `<Audio>` component. The `toneFrequency` prop allows users to specify the pitch adjustment for audio playback, with a valid range from 0.01 to 2(0 being not assignable). This feature enables precise control over audio pitch during rendering.

### Changes Made

- Added `toneFrequency` prop to the `<Audio>` component, allowing users to control pitch adjustment during audio playback.
- Updated documentation to include information about the `toneFrequency` prop, specifying its usage and valid range.
- Ensured that the test suite passes with the new changes.
- Documented potential tradeoffs, if any, related to the implementation of the `toneFrequency` prop.

### Tradeoffs Considered

Not at the moment. 

### How to Test

Please refer to **audio.md** Docs.

### DIFF
--- a/packages/core/src/CompositionManager.tsx
+++ b/packages/core/src/CompositionManager.tsx
@@ -133,6 +133,7 @@ export type TRenderAsset = {
 	mediaFrame: number;
 	playbackRate: number;
 	allowAmplificationDuringRender: boolean;
+	toneFrequency: number | null;
 };
 
 export const compositionsRef = React.createRef<{
--- a/packages/core/src/CompositionManager.tsx
+++ b/packages/core/src/CompositionManager.tsx
@@ -124,6 +124,17 @@ export type TSequence = {
 	loopDisplay: LoopDisplay | undefined;
 } & EnhancedTSequenceData;
 
+
+type ToneFrequencyConstraints = {
+	min: number;
+	max: number;
+};
+
+const TONE_FREQUENCY_CONSTRAINTS: ToneFrequencyConstraints = {
+	min: 0,  // Minimum allowed value for toneFrequency
+	max: 1, // Maximum allowed value for toneFrequency
+};
+
 export type TRenderAsset = {
 	type: 'audio' | 'video';
 	src: string;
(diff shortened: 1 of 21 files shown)


JonnyBurger (changes requested):
This goes into the right direction.

However, the PR did not actually work, it was not actually possible to pass a prop and then the filter was not actually applied to the audio. I helped a bit here!

Now it works, but I noticed regarding documentation and validation there is one major flaw:
It is currently implemented to take values between 0 and 1, however `1` is the normal pitch, and a value like 1.2 will make the pitch 20% higher, while a value like 0.8 will make the pitch 20% lower.

It is not a range between 0 and 1.
Can you fix this in the validation logic and documentation?

evoxf1:
@JonnyBurger 

Ya sure it can be fixed maybe altering the function, so according to you the pitch should be between 0-2, meaning value less that one will reduce the pitch and more than 1 will increase?

JonnyBurger (approved):
Nice! Thanks a lot! 🙌

PR #2994, opened 2023-10-10
`@remotion/renderer`:  add extractAudio() function

### LINKED ISSUE #2900: `extractAudio()` API
# Feature Request 🛍️
There should be an ability in the library to separate the audio track from the video.

## Use Case

This is beneficial in cases where we need the audio to be transcribed, as some cloud providers don't transcribe video out of the box; some require the content to be in audio format. Sending video would waste bandwidth and slow down the upload and transcription process, as only the audio is needed.

## Acceptance criteria

- Implement a new function in Rust which takes a local file path and an output path and extracts only the audio stream of the video.
- It should avoid re-encoding the audio if possible. A setting for would be a nice-to-have but not necessary.
- Documentation should be written for it with the same quality as other APIs in `@remotion/renderer`

## Resources
- getVideoMetadata() implementation, another function from @remotion/renderer: https://github.com/remotion-dev/remotion/blob/main/packages/renderer/src/get-video-metadata.ts
- getVideoMetadata() documentation as a template: https://www.remotion.dev/docs/renderer/get-video-metadata
- How to work with Rust code: https://www.remotion.dev/docs/contributing/rust
- FFmpeg Rust examples: https://github.com/zmwangx/rust-ffmpeg/tree/master/examples

## Bounty
💎 This issue is part of Hacktoberfest and has a bounty on it!
- **General Rules**: https://www.remotion.dev/bounties

🎃 **Hacktoberfest rules**
- You can only work on 1 issue at a time.
- Only 1 person can work on the same issue.
- We explicitly assign a developer. Don’t work on an issue if we do not assign you.
- First-time contributors will get preferential treatment.
- You must set up the Remotion repository locally and test your changes. [Contributing to Remotion | Remotion | Make videos programmatically in React](https://www.remotion.dev/docs/contributing)
- Your PR must fulfill the acceptance criteria.
- If you use AI, you must verify that the output is correct.
- All tests (functionality and formatting) must pass.

Issue already taken? Find more issues under https://remotion.dev/bounties

josefeliuf:
- closes #2900 
- Created extract_audio function in ffmpeg.rs
- Added ExtractAudio option to CliInputCommandPayload
- The function was tested with a mp4 video using the following:
`cargo run '{"payload":  {"type": "ExtractAudio",
    "params": {
      "input_path": "video.mp4",
      "output_path": "output.aac"
    }
  },
  "nonce": "66"
}'`

- It sucessfully created an output.aac file with the audio of the video.


josefeliuf:
@JonnyBurger hey! i added the ts wrapper and the documentation. I noticed that the docs also have a json in articles.ts but there is no info about that in the [contributing to the docs](https://www.remotion.dev/docs/contributing/docs) page. Should i add one? Also in extract-audio.md what should i put in the image path? (check image below). Finally, do i need to add tests?

image

JonnyBurger (changes requested):
I added a test, which can be run with:

```bash
cd packages/renderer
node build.mjs
pnpm exec vitest src/test/extract-audio.test.ts
```

Unfortunately what I tried led to a crash in Rust:

```
Only AAC streams can be muxed by the ADTS muxer
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: ffmpeg::Error(22: Invalid argument)', renderer/rust/ffmpeg.rs:305:25
```

The feature does not have to necessarily convert the audio, but it should do some good error handling and not crash.

josefeliuf:
@JonnyBurger okay no problem, i'll fix it and commit again, thanks!

packages/renderer/src/extract-audio.ts
```diff
@@ -0,0 +1,22 @@
...
+		logLevel: options?.logLevel ?? 'info',
+		indent: false,
+	});
+	const metadataResponse = await compositor.executeCommand('ExtractAudio', {
+		input_path: options.videoSource,
+		output_path: options.audioOutput,
+	});
+	compositor.finishCommands();
+	await compositor.waitForDone();
+	return JSON.parse(metadataResponse.toString('utf-8')) as VideoMetadata;
```
josefeliuf: @JonnyBurger why did u add a metadata response? extract-audio just returns Ok(vec![])

josefeliuf: i just realize i didn't push the ts wrapper, sorry my bad but i did almost exactly the same

JonnyBurger: Oh yeah, that's why I did push it, I forgot to rename it here.

Thanks for going back to this PR, looking mostly good though!

josefeliuf:
@JonnyBurger I was thinking, what if we optionally ask the user for the directory in which he wants the output audio file to be saved (default to the same level as input path) and identify the audio codec programmatically to prevent the user from entering an incorrect output extension? The other way would be that even if the user enters an incorrect extension, to change the extension programmatically anyway, but perhaps this could lead to confusion.

JonnyBurger:
@josefeliuf Cool! Yes I think that approach works 😊
I need a few extra days to review it but if the approach you described works well, then I'll approve 👍

Thanks a lot for the effort!

JonnyBurger:
Very happy with this! Thanks for the contribution 🙌

JonnyBurger (approved):

josefeliuf:
@JonnyBurger Heyy that's awesome, I'm glad you liked it. Thanks you too for your disposition and time. I was wondering, how do i claim the bounty?

JonnyBurger:
@josefeliuf Sorry for missing this - you can claim a bounty by using the `/claim #2994` command in the PR body. Paid out now!

PR #3013, opened 2023-10-16
`@remotion/renderer`: Add `angle-egl` as a new Chromium renderer

UmungoBungo:
Rendering with a GPU in headless mode was not working previously. After adding this flag, and monitoring the GPU via the `nvidia-smi` command, you can see a GPU process get added and the memory spike.

I'm not sure if there are other ramifications for setting this, and whether it would be preferable to bubble this option up and expose it via an API, or to leave it as a default as I have written here.

https://github.com/remotion-dev/remotion/assets/22192773/a3408b79-ebee-4e1f-998a-de44214dd5cf


UmungoBungo:
[ChatGPT session](https://chat.openai.com/share/a64c7684-91b1-4b63-9e2f-7a7ae22bfd7f) around use-gl=egl vs setting use-gl=angle and use-angle=gl-egl.

I didn't approach it this way intentionally, I was just trying different things to get it working. However, it does seem that this method can provide wider support.

bilalmughal:
I did numerous tests and iterations with different driver versions and even different OS/Kernel and took me good 30-40 hours. And my research and results led me to the conclusion that the headless mode using `egl` directly was crashing GPU drivers during initialization and was unreliable. Therefore, it involves a trade-off, although a very minor one. That's why I opted to use Puppeteer with `args: ['--use-gl=angle', '--use-angle=gl-egl']` while utilizing the Angle layer.

image

----

It's important to note that running this setup on AWS, especially with Nvidia GPU, acceleration, won't work out of the box. The GPU needs to be initialized first. To address this, I've included a startup service in the workflow, which I've shared in the draft on Discord. This information should be mentioned as well.

image

JonnyBurger:
Thanks a lot, taking some time to do my own experiments now to also catch up with the things you've learned!

UmungoBungo:
Hey Jonny, this is holding me up a bit as I'm using a fork to test things out, and it would simplify things a whole lot if this were in the main branch. Is there anything I can do to help get this in quicker? Perhaps for now I can change it to something like this?
```ts
if (renderer === 'angle-egl') {
    return [`--use-gl=angle`, `--use-angle=gl-egl`];
}
```

so that if anyone is using `use-gl=angle` it won't change their flow at all?

JonnyBurger:
Hope to add it for todays release!

JonnyBurger:
To ship it quickly and to still be confident, I am adding this as a new option: `{chromiumOptions: {gl: "angle-egl"}}`

PR #3014, opened 2023-10-16
Update LICENSE.md

MehmetAdemi:
add FAQs to License.md + update broken link to Support.md

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -7,7 +7,7 @@ Depending on the type of your legal entity, you are granted permission to use Re
 
 ## Free license
 
-Copyright © 2023 [Remotion](https://www.remotion.pro/license)
+Copyright © 2023 [Remotion](https://www.remotion.dev)
 
 ### Eligibility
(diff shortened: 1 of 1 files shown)


LICENSE.md
```diff
@@ -7,7 +7,7 @@ Depending on the type of your legal entity, you are granted permission to use Re
 
 ## Free license
 
-Copyright © 2023 [Remotion](https://www.remotion.dev/)
+Copyright © 2023 [Remotion](https://www.remotion.pro/license)
```
JonnyBurger: It's just meant to be a link to the company website, no? Why link to the license page?

MehmetAdemi: We can leave this one as it was. However, I think it makes sense to link directly to the pricing page. Since it's about the license anyway. So I suggest alternatively to use the link on line 44.

MehmetAdemi:
@JonnyBurger What do you think about this last adjustment?

PR #3063, opened 2023-10-24
Warn if Bun cannot determine musl or glibc and assume glibc

### LINKED ISSUE #3062: renderMedia fails on Bun
When using renderMedia to ssr render a composition with bun, it fails returning the following error.

**TypeError: undefined is not an object (evaluating 'process.report.getReport')** 

Here: ```@remotion/renderer/dist/compositor/get-executable-path.js:8:12```

Bun Version: 1.0.7
Node Version: 18.16.0
OS: Fedora Linux 38
Remotion version: 4.0.51

Since apparently the ```process.report``` obj. isn't available on Bun, I also tried to disable the ```isMusl``` func and running ```@remotion/compositor-linux-x64-gnu``` directly and that removed the error but the script never went to the end neither rendered out anything.

JonnyBurger:
Fixes #3062

### DIFF
--- a/packages/cli/src/ffmpeg.ts
+++ b/packages/cli/src/ffmpeg.ts
@@ -1,28 +1,33 @@
 import {RenderInternals} from '@remotion/renderer';
 import {spawnSync} from 'node:child_process';
 import {chmodSync} from 'node:fs';
+import {ConfigInternals} from './config';
 
 export const ffmpegCommand = (_root: string, args: string[]) => {
-	const binary = RenderInternals.getExecutablePath('ffmpeg');
+	const logLevel = ConfigInternals.Logging.getLogLevel();
+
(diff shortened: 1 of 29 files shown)


OmarKhled:
If possible, can you release it? I've a release dependent on it 😅
Thanks for your quick response!

JonnyBurger:
I can later today!

JonnyBurger:
@OmarKhled Out now, sorry it took another night!

OmarKhled:
@JonnyBurger Thanks for the help! 
Just created a small fix at #3068 to check if process.report is undefined (the latest changes was checking if it's not undefined, though it's undefined in Bun) 
Would much appreciate it if you reviewed that and schedule another release as I'm still pending on that 😅

Thanks for your great contributions!

PR #3068, opened 2023-10-25
check if process.report is undefined with bun

OmarKhled:
Related to #3062
Just a small fix on the changes at #3063 to check if process.report is undefined in case of using Bun.


JonnyBurger (approved):

JonnyBurger:
Thanks! Yes can make another release, sorry for the wrong logic

PR #3069, opened 2023-10-25
Updated reference article in cloud gpu docs

bilalmughal:
Updated the reference article URL from draft to published one in "Using the GPU in the Cloud"


JonnyBurger:
Thanks a lot again! 🙌🏼

PR #3098, opened 2023-10-30
feat: Add flag to enable lambda insights

### LINKED ISSUE #2740: lambda create function flag for enabling lambda insight
a flag for rendering lambda function to enable lambda insight could be very useful
with that aws will provde memory and cpu usage metrics which could help optimizing lambda functions

aswathkk:
Adds `--enable-lambda-insights` flag to the `npx remotion lambda functions deploy` command so that AWS Lambda Insights will be enabled for the function.

/claim #2740
Fixes #2740


JonnyBurger:
Thanks a lot for the PR! Should be able to review it in 1-2 days.

JonnyBurger (approved):
I couldn't get it to work until I added the `CreateLogGroups` permission, but other than that, pretty nice work! Well done 🙌🏼

aswathkk:
Even though in the docs it was mentioned to add `CreateLogGroups` permission, it was working without it when I tested using my account. That's why I didn't add the permission explicitly. Thanks for the meaningful additions on top of this PR!

PR #3121, opened 2023-11-03
Update Showcase section on website

MehmetAdemi:


MehmetAdemi:
@JonnyBurger 
Can you have a look at it, give feedback, and afterward we merge it?

Later (in another PR), I will reduce the number of videos and add specific thumbnails.

MehmetAdemi:
The last thing to do: 
reduce the number of videos and then choose a good thumbnail for each showcase video

JonnyBurger (approved):

PR #3124, opened 2023-11-04
Performance and Memory Boost: React JS Dependency Updates 🚀

sanjaiyan-dev:
Hello, this pull request aims to bring about minor improvements in performance and memory management by updating the React JS dependencies.

1) We have updated the dependency array to use the object's value instead of the object itself. This change will improve performance because when an object is added as a dependency, it retains its reference rather than its value. This prevents the object from being recreated in every render with a new reference in memory. This modification will benefit React JS hooks like `useCallback` and `useMemo`. The code has been changed from:

```js
const selectedEditorProps = useMemo(() => {
	return composition ? allEditorProps[composition.id] ?? {} : {};
}, [allEditorProps, composition]);
```

to:

```js
const selectedEditorProps = useMemo(() => {
	return composition ? allEditorProps[composition.id] ?? {} : {};
}, [allEditorProps[composition?.id], composition]);
```

2) We have made the dependency array for `useMemo` or `useCallback` conditional. This change is expected to result in a performance gain. For instance, by changing the dependency to `[mediaVolume === 0]`, it will only create a new instance when the volume is 0, and it won't have any effect when the volume has any other numerical value. This change will lead to improvements in memory and performance.

Extremely sorry if I made any mistakes :)


JonnyBurger:
Thanks a lot @sanjaiyan-dev! Very much appreciated 🙌🏼
I'm only going to merge a few of them since others introduced linter warnings and errors.

JonnyBurger (approved):

PR #3131, opened 2023-11-07
alphabeticalized third-party.md

tomByrer:
There are enough integrations to need to be in alphabetical order ;)
Removed outdated CSS example

(doc only)


JonnyBurger:
Makes sense! Thanks 🙌

PR #3136, opened 2023-11-08
Remotion Studio: Remove console.log() statement that was forgotten

loalf:


JonnyBurger:
Thanks a lot! We should have caught this, re-enabled the lint rule for this package again

PR #3237, opened 2023-11-30
`@remotion/renderer`: If system is about to run out of memory, halfen the cache size and halfen the maximum cache size

JonnyBurger:


Amatewasu:
Nice idea!

PR #3258, opened 2023-12-07
Return `pixelFormat` in `getVideoMetadata()`

### LINKED ISSUE #3216: Return `pixelFormat` in `getVideoMetadata()`
Currently, we need to explicitly state if the source video is transparent in `OffthreadVideo`.
This sometimes causes problems, as video containers such as `.mov` can have or not have an alpha channel.

# Feature Request 🛍️
It would be very useful to detect whether a video has or not transparency, without requiring this explicit information from the user.

## Use Case
That would enhance support and generalization for different video encodings, especially when we don't know beforehand which types of videos are we going to receive.

## Possible Solution
A feature could be implemented to detect whether a video has a transparency channel or not in the already existing function [get-video-metadata](https://www.remotion.dev/docs/renderer/get-video-metadata).

jkcs:
close #3216
/claim #3216


JonnyBurger:
It's ok, thats because we disable windows build right now because it is broken.

no need to do anything!

JonnyBurger:
Looks good, but I'll review it only by next week!
Sorry for the delay 🙈

jkcs:
No problem, have a great weekend!

JonnyBurger:
Very well done! Thanks a lot and sorry for the long review time!

PR #3262, opened 2023-12-08
feat: make sure `renderMediaOnLambda` works on edge

Iamshankhadeep:
runtime

@JonnyBurger This PR  removes `node:crypto` from `renderMediaOnLambda` so it should make sure that `renderMediaOnLambda` should work on `edge` runtime now.


JonnyBurger:
Thanks a lot! I wanted to not introduce a new dependency for this, but I figured out a way to achieve the same goal - get rid of Crypto APIs not available on the Edge. Let's see if that unblocks it or if there is more we need to do :)

PR #3302, opened 2023-12-27
`@remotion/three`: useOffthreadVideoTexture()

samohovets:
Hey team, Merry Christmas and Happy Holidays! 🎅

This is a WIP version of implementing the `useOffthreadVideoTexture` hook for the `@remotion/three` package. 
In the context of my project:
- I encountered well-known limitations of `<Video>` component such as throttling by Chrome if the composition is too heavy (since we're creating 3D scenes this is pretty common).
- I need the ability to encapsulate some 3D object and it's video texture in a React component without having to render `<Video>` at the top level alongside the `<ThreeCanvas>`.

This is a draft version since it needs some unexposed APIs from Remotion core:
- [ ] The video playing in the preview is a bit laggy (I am not sure if I'm syncing Remotion's time to video tag correctly). For the workaround, I render `<Video>` in my project just for the preview and passing videoRef to the `useOffthreadVideoTexture` without using the current `useVideoRef`. 
- [x] it requires the user to render `<OffthreadVideo>` somewhere in the composition so `compositor` will start serving the video frames

Original Discord conversation: https://discord.com/channels/809501355504959528/1178662345476681878/1178675456447356989


JonnyBurger:
Thanks a lot! 🙌

Looks pretty good actually! Would you be up to write some docs as well?  
We can also do it, not mandatory, it would just accelerate it a bit I guess!

samohovets:
Thanks for the refactors and fixes! 🙏

Some thoughts:

1. I haven't looked through the code of the original `<OffthreadVideo>` yet, but is there a way to provide the ability to use `useOffthreadVideoTexture` without manually rendering offthread video in the composition to make `compositor` respond to image requests?

2. Is synchronizing video in the `useVideoRef` done the correct way? I've been experiencing some lags/freezes of the video after seeking in Remotion Player. Maybe we should use `useMediaPlayback` here? My guess is it's freezing because there's no mechanism of seek throttling as in `useMediaPlayback`

JonnyBurger:
This is awesome, thanks a lot!

No, it should not be necessary to mount another OffthreadVideo component. I'll play around with it in the testbed.
And if it works, then the use of useVideoTexture seems fine!

samohovets:
Just checked out the example in the `ThreeScene` folder and it looks pretty interesting! 

I can see you untightened the API a bit which is great, but how do we create refs and `<Video>`s dynamically?

I want to render an array of "scenes" where each scene may have a 3D model with a video texture. 

Let's say my main `.map` that renders `<Series.Sequence>` is located inside of `ThreeCanvas`. So to display the preview I should render multiple `<Video>`'s somewhere on the top level? 🤔

Something like this?
https://stackoverflow.com/questions/55995760/how-to-add-refs-dynamically-with-react-hooks

JonnyBurger:
@samohovets In this case, you could:

- Render multiple components which each render a Video or useVideoRef(). There is no requirement for it to be on the top-level.

- Use `createRef()` instead to create many refs, bypassing the requirement to use a React hook which you cannot dynamically change how many hooks you call. Something like:

`const refs = new Array(5).fill(true).map(() => React.createRef(null))`

```tsx
refs.map(r => <Video ref={r} />)
```

samohovets:
> * Render multiple components which each render a Video or useVideoRef(). There is no requirement for it to be on the top level.

Sorry by top-level I meant "outside of `ThreeCanvas`". React Three Fiber doesn't let you render Remotion's `<Video>` inside the canvas.

`createRef` looks like a solution, thanks!

PR #3338, opened 2024-01-09
docs: avoid common pitfall from code reuse

Just-Moh-it:


JonnyBurger:
So true! Thanks for bringing this into the docs

PR #3353, opened 2024-01-12
docs: add behavior for single number passed as `frameRange` to `renderMediaOnLambda`

Just-Moh-it:
Had to go into the source code to see the single-frame behavior... while creating this pr, realized there were docs on the behavior for a single number being passed to the `frameRange`, just for `renderMedia` though... this change might help 😎


JonnyBurger:
Lol, I was surprised we support a single number!

Well then, thanks for fixing this! 🙌🏼

JonnyBurger (approved):

PR #3363, opened 2024-01-15
`@remotion/lambda`: Don't throw error if default AWS profile is set

JonnyBurger:


shellscape:
Looks good. I'd also recommend throwing in the short-circuit option into this. An environment variable would work.

PR #3369, opened 2024-01-17
`@remotion/renderer`: Add `--repro` flag 

### LINKED ISSUE #3341: Make a `--repro` flag for `npx remotion render`
## Motivation

I always ask people to provide a repro, but why not make it easy to provide a repro?

## Acceptance criteria

- Add an optional `repro` flag to renderMedia(). If set to `true`, make it do the following things:
  - Create a folder in the temporary directory and log it out. This is the repro folder
  - Save the bundle:
  	- a) If the `serveURL` is a URL, print it to a file in the temporary directory
  	- b) If the serveURL is a directory, copy the directory to a subdirectory of the repro folder
  - Save the input props:
 	- Save the serialized input props to a file in the repro folder
 - Every time `Log.verbose`, `Log.warn`, etc. gets called, serialize the message to JSON and append it to a log file in the repro folder. Make this efficient so you don't have to first read the existing file to append to it
 - Write the `process.argv` array to the repro folder so we know which command was executed
 - If the render is successful, copy the output to the repro folder
 - If everything is successful, ZIP the repro folder without using an additional dependency (on macOS for example, spawn a child process using the `zip` command)
 - copy the repro folder into the current working directory and print out the path. don't overwrite any files (name it repro.zip, repro1.zip, repro2.zip).
 - Allow the option to be specified via CLI (--repro) in the same way like other boolean flags
 - Document the flag in renderMedia() docs and `npx remotion render` docs and indicate that it is available from `currentVersion + 0.01`

## Bounty

/bounty 350

jkcs:
Close #3341
/claim #3341

### DIFF
--- a/packages/bundler/src/index-html.ts
+++ b/packages/bundler/src/index-html.ts
@@ -34,6 +34,7 @@ export type RenderDefaults = {
 	colorSpace: string;
 	multiProcessOnLinux: boolean;
 	beepOnFinish: boolean;
+	repro: boolean;
 };
 
 declare global {
--- a/packages/cli-autocomplete/src/source.ts
+++ b/packages/cli-autocomplete/src/source.ts
@@ -207,6 +207,11 @@ const localRenderOptions: Fig.Option[] = [
     name: "--muted",
     description: "Mute the output video",
   },
+  {
+    name: "--repro",
+    description:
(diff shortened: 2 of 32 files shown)


JonnyBurger:
Thanks a lot! Reviewin tomorrow!

PR #3386, opened 2024-01-20
docs: remove unused arg in use-current-player-frame convinience-hook

Just-Moh-it:
I refer to this doc often since i implement custom remotion playback controls a lot, and since `useSyncExternalStore` isn't used very often, it takes some time to understand how the `useCurrentPlayerFrame` hook works. But every time I do, I can't help but notice and end up being left confused for a while after I read the callback function and see the frame is being fed in from two locations, but only one of those is actually returned by the hook.

Doesn't change any functionality, but might help understand this hook for people who don't use `useSyncExternalStore` a lot.


JonnyBurger:
💡

Indeed, seems like I didn't understand useSyncExternalStore fully either 😄

Thanks!

PR #3418, opened 2024-01-25
Add support for negative number in interpolate styles

ChetanKarwa:


JonnyBurger:
Nice! Thanks for catching that! 🙌🏼

PR #3439, opened 2024-02-01
`@remotion/renderer`: Disable `fromSurface` for better screenshot reliability on Lambda

JonnyBurger:
As explained in the comment


packages/renderer/src/screenshot-task.ts
```diff
@@ -75,7 +75,12 @@ export const screenshotTask = async ({
 								width,
 							},
 				captureBeyondViewport: true,
-				optimizeForSpeed: true,
+				optimizeForSpeed: false,
```
dcodesdev, line 78: @JonnyBurger This change made my renders drastically slower, was it intentional?

PR #3448, opened 2024-02-05
Removed --save-exact option from pnpm command

Dawsoncodes:
`--save-exact` option does not exist on `pnpm install` command.

It only exists on `pnpm add`

Running `pnpm exec remotion upgrade` in a pnpm project results in an error: `--save-exact` option does not exist.

https://pnpm.io/cli/install#options


JonnyBurger:
Hmm, thanks a lot! I use pnpm myself, surprised I never ran into that.

PR #3454, opened 2024-02-06
docs: New solution to background-image flickering.

alexanderson1993:
The existing solution works for `background-image`, but not for `mask-image`. The solution proposed in the commit works for both, and in some cases might be simpler to implement.


JonnyBurger:
Thanks a lot for this!

I rewrote it a bit, because most of the time the code can just be written to layer an AbsoluteFill on top of an Img and the background-image code can be completely avoided. I put your code snippet as an alternate solution now!

PR #3462, opened 2024-02-10
`@remotion/renderer`: Allow for seamless AAC stitching

JonnyBurger:
TODO: 

- [x] Function to determine when to use this
- [x] Fix no audio in output
- [x] Add padding to chunks
- [x] Fix progress function


packages/lambda/src/functions/renderer.ts
```diff
@@ -101,10 +101,7 @@ const renderHandler = async (
 
 	const outdir = RenderInternals.tmpDir(RENDERER_PATH_TOKEN);
 
-	const chunkCodec: Codec =
-		params.codec === 'gif' || params.codec === 'h264'
-			? 'h264-mkv'
-			: params.codec;
+	const chunkCodec: Codec = params.codec === 'gif' ? 'h264-mkv' : params.codec;
```
JonnyBurger: Is this really needed? GIFs don't have sound

PR #3475, opened 2024-02-13
docs: Add extension outname

Just-Moh-it:
Had issues with rendering when not including file extensions. The docs don't say it explicitly, so this might save someone a few minutes 😄


JonnyBurger:
Makes sense! That's indeed a gotcha, thanks for adding it!

JonnyBurger (approved):

PR #3483, opened 2024-02-15
fix typo in serveless framework docs

BrunnerLivio:
Don't think `ts-nodwe` exists 😄


JonnyBurger:
I checked, it indeed does not 😄 

Thanks!

PR #3492, opened 2024-02-18
Update overwriting-webpack-config.md

UmungoBungo:
I'm fumbling my way through sass, but I couldn't get this to work without installing these other packages. Are they required? They are listed in the webpack config;

```
...
use: [
   { loader: "style-loader" },
   { loader: "css-loader" },
   { loader: "sass-loader", options: { sourceMap: true } },
],
```

### DIFF
--- a/packages/docs/docs/overwriting-webpack-config.md
+++ b/packages/docs/docs/overwriting-webpack-config.md
@@ -210,14 +210,14 @@ values={[
 <TabItem value="npm">
 
 ```bash
-npm i sass sass-loader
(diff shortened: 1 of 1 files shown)


JonnyBurger:
Although those modules are dependencies of `@remotion/bundler`, it appears that when you override, it now resolves to the directory of where the config file is in.

This could be a reason why it works in my testing in the monorepo but not for you. So it's probably a good idea to just add it to the docs. Thanks a lot! 🙌🏼

JonnyBurger (approved):

Issue #3326, opened 2024-01-08
Enable drag-and-drop upload to Remotion Studio Assets panel with progress bar and server integration.


JonnyBurger:
[From message posted on Discord by **jonnyburger**](https://discord.com/channels/809501355504959528/817306238811111454/1193900990705967104)
----
There should be a way to drag items into the Remotion Studio Assets panel and them being uploaded to the public folder
there should be a progress bar and the files should go through our Remotion Studio server

PR #3498, opened 2024-02-19
`@remotion/studio`: Allow to upload assets by dragging and dropping them into the assets panel

ChetanKarwa:
Resolves: https://github.com/remotion-dev/remotion/issues/3326


pnpm-lock.yaml
```diff
@@ -1949,6 +1949,9 @@ importers:
       '@remotion/studio-shared':
         specifier: workspace:*
         version: link:../studio-shared
+      formidable:
+        specifier: ^3.5.1
+        version: 3.5.1
```
JonnyBurger: Can we make this work without dependencies?  
Unfortunately this adds 6 dependencies to a lean package otherwise.
I also mentioned this in the specification for this issue

packages/studio/src/components/AssetSelector.tsx
```diff
@@ -96,15 +103,47 @@ export const AssetSelector: React.FC = () => {
...
+		},
+		[],
+	);
+
+	const onDrop: React.DragEventHandler<HTMLDivElement> = useCallback(
+		(e) => {
+			e.preventDefault();
+			e.stopPropagation();
+			handleUploadFile(e.dataTransfer.files[0], dropLocation || '/');
+			setDropLocation(null);
```
JonnyBurger, line 130: If the Studio server is disconnected or you deploy the Studio statically, the drop handler is enabled and fails with a generic error.

It should not accept dragging at all.

Use the `readOnly` props and the `const connectionStatus = useContext(StudioServerConnectionCtx)
		.previewServerState.type;` to determine if the Studio is connected to a server.

packages/studio-server/src/routes.ts
```diff
@@ -195,6 +197,74 @@ const handleOpenInEditor = async (
...
+	req: IncomingMessage,
+	res: ServerResponse,
+	publicDir: string,
+): void => {
+	const TWO_GIGABYTES = 2 * 1024 * 1024 * 1024;
+	const form = formidable({
+		maxTotalFileSize: TWO_GIGABYTES,
+		maxFieldsSize: TWO_GIGABYTES,
+		maxFileSize: TWO_GIGABYTES,
+	});
```
JonnyBurger: Uploading a 10GB file does not seem to get rejected or display a useful error message

packages/studio/src/components/AssetSelectorItem.tsx
```diff
@@ -141,13 +160,53 @@ export const AssetFolderTree: React.FC<{
...
+	const isDropDiv = useMemo(() => {
+		return dropLocation === combinedParents;
+	}, [combinedParents, dropLocation]);
+	useEffect(() => {
+		if (dropLocation === null) {
+			dragDepthRef.current = 0;
+		}
+	}, [dropLocation]);
 	return (
-		<div>
```
JonnyBurger, line 150: https://github.com/remotion-dev/remotion/assets/1629785/28c1c086-43e0-4792-a9fd-367465fb9b40


It's weird that you can drop the file into a folder, but not over a folder. I thought it was a bug that all folders get highlighted at first, it's even weirder when the folders are expanded.

ChetanKarwa: So folders are a part of the parent tree, when we hover over a folder we highlight all the files and folders of that level.

PR #3546, opened 2024-03-03
fix: increase minimum value for shutterAngle slider

thecmdrunner:
In Camera Motion Blur example, the player crashes when shutterAngle is zero.
image


JonnyBurger:
Oh, thanks a lot for catching that, and sending a PR!

PR #3547, opened 2024-03-04
fix: remix template link

thecmdrunner:


JonnyBurger:
Damn, how many mistakes do we have in the docs 😅 thanks for staying alert!

PR #3555, opened 2024-03-06
Allow wrapping in interpolate extrapolation

Jonhyfun:
Sorry for not documenting this in an issue, i just stumbled across the need to use a wrapping feature and figured i could implement it quickly (its a very small change).

Now when you call interpolate, you can use `'wrap'` in `extrapolateLeft` or `extrapolateRight`

Meaning that if you have

`interpolate(frame, [0,5], [0,50], {extrapolateRight:'wrap'})`

then

`interpolate(6, [0,5], [0,50], {extrapolateRight:'wrap'}) //yields 0`
`interpolate(7, [0,5], [0,50], {extrapolateRight:'wrap'}) //yields 10`
`interpolate(11, [0,5], [0,50], {extrapolateRight:'wrap'}) //yields 50`

and so on, wrapping arround your outputRange so that it loops on the intended values forever. A great example usecase would be a forever scrolling background, that instead of forever expanding, just returns to your initial x position.

### DIFF
--- a/packages/animation-utils/src/transformation-helpers/interpolate-styles/index.tsx
+++ b/packages/animation-utils/src/transformation-helpers/interpolate-styles/index.tsx
@@ -1,4 +1,4 @@
-import type {InterpolateOptions} from 'remotion';
+import type {ExtrapolateType, InterpolateOptions} from 'remotion';
 import {interpolate, interpolateColors} from 'remotion';
 import type {
 	CSSPropertiesKey,
@@ -7,8 +7,6 @@ import type {
 	UnitNumberAndFunction,
 } from '../../type';
 import {breakDownValueIntoUnitNumberAndFunctions} from './utils';
-
-type ExtrapolateType = 'extend' | 'identity' | 'clamp';
 
 const interpolatedPropertyPart = ({
 	inputValue,
--- a/packages/core/src/interpolate.ts
+++ b/packages/core/src/interpolate.ts
@@ -1,6 +1,6 @@
 // Taken from https://github.com/facebook/react-native/blob/0b9ea60b4fee8cacc36e7160e31b91fc114dbc0d/Libraries/Animated/src/nodes/AnimatedInterpolation.js
 
-export type ExtrapolateType = 'extend' | 'identity' | 'clamp';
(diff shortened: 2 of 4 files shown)


JonnyBurger:
Hey @Jonhyfun!
This is great idea, thanks for just going ahead and implementing this.

I am down to ship this, I am just surprised about the output.

Consider:

```
interpolate(1.5, [0, 1], [0, 2], {extrapolateRight: 'wrap'})
```

I expected it to yield `1`, since I want it to loop the output range between 0 and 2. So I would expect it to give the same result as passing `0.5` as input, by first running `1.5 % 1` (1 being the last item in the inputRange).

Would that not make more sense?

Also in your example:

> interpolate(6, [0,5], [0,50], {extrapolateRight:'wrap'}) //yields 0

This is weird, what if you pass a value between `5` and `6`? It always stays zero? It seems like a "one-off" error to me, I would expect it to yield `(6 % 5) / 5 * (50 - 0) + 0 = 10` instead.

Jonhyfun:
> Hey @Jonhyfun! This is great idea, thanks for just going ahead and implementing this.
> 
> I am down to ship this, I am just surprised about the output.
> 
> Consider:
> 
> ```
> interpolate(1.5, [0, 1], [0, 2], {extrapolateRight: 'wrap'})
> ```
> 
> I expected it to yield `1`, since I want it to loop the output range between 0 and 2. So I would expect it to give the same result as passing `0.5` as input, by first running `1.5 % 1` (1 being the last item in the inputRange).
> 
> Would that not make more sense?
> 
> Also in your example:
> 
> > interpolate(6, [0,5], [0,50], {extrapolateRight:'wrap'}) //yields 0
> 
> This is weird, what if you pass a value between `5` and `6`? It always stays zero? It seems like a "one-off" error to me, I would expect it to yield `(6 % 5) / 5 * (50 - 0) + 0 = 10` instead.

Youre right, i really didnt think about non integer numbers. Ill be looking into that soon.

UmungoBungo:
this is awesome 🔥

JonnyBurger:
I finished it up now with the logic that I think makes sense!

Jonhyfun:
Hey, many thanks, sorry for taking so long, i got hooked up with work and i couldnt figure out some edge-cases

PR #3577, opened 2024-03-12
`@remotion/install-whisper-cpp`: transcribe() API

patsalv:


JonnyBurger (approved):
The documentation for this is stellar! 👏

One of the best documented APIs we have

patsalv:
> The documentation for this is stellar! 👏
> 
> One of the best documented APIs we have

Thanks!🙏

PR #3630, opened 2024-03-25
Fix Docusaurus Vercel URL

slorber:
The env variables do not seem to work as you expect, and this is harmful for Remotion site SEO which has bad canonical urls.

Page metadata: 

```html
<link data-rh="true" rel="canonical" href="https://remotion-ebgrex03y-remotion.vercel.app/blog/faster-lambda">
```

RSS feed: https://www.remotion.dev/blog/rss.xml

CleanShot 2024-03-25 at 18 39 26

IMHO it is preferable to harcode the URL. This way if Google ever tries to index a deploy preview, Google will consider that the deploy preview is a "copy" and only consider your prod website as the canonical source of content without penalizing it.

### DIFF
--- a/packages/docs/docusaurus.config.js
+++ b/packages/docs/docusaurus.config.js
@@ -6,10 +6,7 @@
 const config = (mode) => ({
   title: "Remotion | Make videos programmatically in React",
   tagline: "Make videos programmatically",
-  url:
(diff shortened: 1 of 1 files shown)


JonnyBurger:
Thanks a lot! I trust you on this one 😃

PR #3660, opened 2024-04-03
`@remotion/install-whisper-cpp`: Semantic version check on Windows & docs update

patsalv:


packages/install-whisper-cpp/src/install-whisper-cpp.ts
```diff
@@ -16,6 +16,13 @@ const installForWindows = async ({
 	to: string;
 	printOutput: boolean;
 }) => {
+	const isSemVer = /^[\d]{1}\.[\d]{1,2}\.+/;
```
JonnyBurger: This looks like it is always `true` 😆 caught it

JonnyBurger (approved):

PR #3684, opened 2024-04-08
fix(docs): wrong watercolor template url in templates

tewarig:
fix(docs): Wrong watercolor template URL

In `https://www.remotion.dev/templates/`, the WaterColor Map is redirecting to `https://www.remotion.pro/mapbox-globe`. It should redirect to `https://www.remotion.pro/mapbox-globe`.


Just-Moh-it (approved):

JonnyBurger:
Oh, nice catch! Thanks a lot for sending a PR! 🙌🏼

PR #3685, opened 2024-04-09
feat: `additionalElementStyles` prop on `fitText` and `measureText`

Just-Moh-it:
With all the existing inputs `fitText` and `measureText` expose, it's not possible to use these functions for more complex inputs where you need more control over the underlying `<span>` element that gets created to measure the text

### Before the addition

Adding a property like this:

Screenshot 2024-04-08 at 11 15 35 PM

Would make the fitText and measureText break and wrap the output.

image

<hr />

### After the addition,

Screenshot 2024-04-08 at 11 17 04 PM

It properly fits the text in now!

Screenshot 2024-04-08 at 11 18 25 PM

## Why not just expose `textTransform` in as an input instead of creating a totally new `additionalElementStyles`

I use this function in a project that uses this function in a complex text element inside a editor that supports various font properties on an element, like textTransform, transform, etc., which I don't think would make sense to have as explicit inputs to this function.

As for the risk of specifying fontFamily or fontSize, or any other property through `additionalElementStyles`, I've excluded them from the type for this property, hence leaving no scope of accidentally specifying the same property twice, once through the direct function params, and again through `additionalElementStyles`.

## What could be improved

- I think the name `additionalElementStyles` is not the best and could be better. It says `element` in the name, whereas the docs don't specify what the `element` actually is. Unless someone digs through the source code and finds out the `measureText` function actually creates an invisible text element in the DOM, they won't know this abstraction exists. Feels like it exposes parts of the abstraction that haven't been documented.
- Naturally, the second would be 'how it works' section on `fitText` docs, which says something like "it uses `measureText` under the hood to calculate the size of the text with a standard fontSize, then uses the ratio of the resulting element's width to proportionally find the fontSize that would fit into the passed in `withinWidth` property. This property exposes the styles on the underlying invisible span element that is created", and something similar on `measureText` as well.

### DIFF
--- a/packages/docs/docs/layout-utils/fill-text-box.md
+++ b/packages/docs/docs/layout-utils/fill-text-box.md
@@ -80,7 +80,7 @@ _string_
 
 Same as CSS style `font-variant-numeric`.
 
-### `textTransform`<AvailableFrom v="4.0.140"/>
+#### `textTransform`<AvailableFrom v="4.0.140"/>
 
 _string_
 
@@ -91,6 +91,12 @@ Same as CSS style `text-transform`.
 _boolean_
 
 If set to `true`, will take a second measurement with the fallback font and if it produces the same measurements, it assumes the fallback font was used and will throw an error.
+
+#### `additionalStyles`<AvailableFrom v="4.0.140"/>
+
+_object, optional_
+
+Additional CSS properties that affect the layout of the text.
 
 ### Return value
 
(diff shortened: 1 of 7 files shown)


JonnyBurger:
Thanks a lot for the PR! 🙌🏼 💙
Sounds good to me.

I changed it to `additionalStyles` for brevity.
And added support for textTransform as a top-level option!

JonnyBurger (approved):

PR #3691, opened 2024-04-10
Studio: Fix splitter layout issue

XeroAlpha:
## Issue

In Remotion Studio, when the title of a video/audio is too long and exceeds the screen width (`innerWidth`), the timeline container also extends accordingly, causing the timeline to not be fully displayed at the minimum zoom level.

Here is an example that triggers this issue:

```typescript
export const MyComposition = () => {
	const videoUrl = 'http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4';
	return (
		<AbsoluteFill>
			<Video src={`${videoUrl}?${'x'.repeat(600)}`} />
		</AbsoluteFill>
	);
};
```

Before:

image

After:

image

This pull request fixes the issue. It is currently unclear whether this fix might cause other bugs.

### DIFF
--- a/packages/studio/src/components/Splitter/SplitterContainer.tsx
+++ b/packages/studio/src/components/Splitter/SplitterContainer.tsx
@@ -12,6 +12,7 @@ const containerRow: React.CSSProperties = {
 	flexDirection: 'row',
 	flex: 1,
 	height: '100%',
+	width: '100%',
 };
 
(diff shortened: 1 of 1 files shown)


JonnyBurger:
Thanks a lot for reporting it and sending a fix alongside! 🙌🏼 Much appreciated!

PR #3703, opened 2024-04-11
Add parameter maxLength to transcribe

### LINKED ISSUE #3688: Add whisper max-length parameter
# Feature Request 🛍️

It would be practical to have the option to omit (don't pass any) or set the max-length parameter passed to whisper.cpp in the transcribe function.

 https://github.com/remotion-dev/remotion/blob/main/packages/install-whisper-cpp/src/transcribe.ts
 
This would make it possible to easily generate sentence transcription: Each transcription would contain multiple tokens instead of one, which is whisper.cpp's default behavior.

## Use Case

Create movie-style subtitles for a video, using whisper.cpp's default token grouping.

## Possible Solution

Add an optional parameter tokens per transcript, with a default of 1 (backwards compatibility). If set to null then don't pass any value to the whisper command.

Option 2: add an optional argument processArgs that applies changes to the args passed to whispercpp.

youssefbenlemlih:
Fixes #3688

This PR adds the possibility to configure the `max-length` parameter that is passed to whisper.cpp (previously hard coded to 1).

The function `transcribe` now also accepts a parameter `maxLength`. 

The parameter can only be also be set to null, to use `whisper.cpp`'s default token grouping (useful for generating a movie-style transcription).

## Backwards compatibility

The function is backward compatible. If `maxLength` is not provided, a default value of 1 is used (same behaviour as before)

## Note
Since `convertToCaptions` groups the transcription-items and not the tokens, the parameter `maxLength` can now only be used when `tokenLevelTimestamps` is set false.

Question: Would `tokensPerTranscriptionItem` be a batter/more accurate name?

/claim 3688

### DIFF
--- a/packages/docs/docs/install-whisper-cpp/transcribe.mdx
+++ b/packages/docs/docs/install-whisper-cpp/transcribe.mdx
@@ -95,6 +95,18 @@ We recommend using at least the `medium` model to get satisfactory results when 
 
 Whether to print the output of the transcription process to the console. Defaults to `true`.
 
+### `maxLength?`<AvailableFrom v="4.0.141" />
+
+_optional - default: `1`_
+
+The maximum amount of tokens included in each transcription item.
+
+Set this flag to `null`, to use `whisper.cpp`'s default token grouping (useful for generating a movie-style transcription).
+
+:::info
+`maxLength` can only be set when `tokenLevelTimestamps` is set to `false`.
+:::
+
 ## Return value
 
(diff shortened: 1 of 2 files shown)


JonnyBurger:
This is perfect! Thanks a lot!
And I agree with your idea, I renamed it to `tokensPerItem`.

youssefbenlemlih:
Thanks @JonnyBurger

PR #3705, opened 2024-04-14
Add code to support bilingual language

alexfernandez803:
This is add functionality for transcribe to explicitly set the language of the the audio, hence forcing whisper to transcribe the audio in the specified language.

I've noticed that when the audio is bi-lingual, so I was speaking filipino/tagalog the transcribed word is automatically converted to english even if **translateToEnglish**' is set to false.

I've added the **-l** flag to allow this.

  -l LANG,   --language LANG     [en     ] spoken language ('auto' for auto-detect)


JonnyBurger:
This is so nice! Thanks a lot for sending it, it looks perfect!

PR #3707, opened 2024-04-15
Experts: update Karel Nagel info

karelnagel:


JonnyBurger:
Thanks for the PR!
You must use the same pnpm version that we use for this to work, ideally with corepack.
I clarified this in the docs!

PR #3729, opened 2024-04-18
`@remotion/renderer`: Handle ERR_TIMED_OUT when offthreadvideo images take too long

JonnyBurger:


JonnyBurger:
I think this is not the root cause for the issue, we have other freezes

PR #3763, opened 2024-04-23
docs: add `loop` difference for video vs offthread video

### LINKED ISSUE #3762: `loop` not available on `<OffthreadVideo />`
Not really a major one but someone who's using `<Video />` would expect `<OffthreadVideo />` to be a drop in replacement for `<Video />`.

Was having issues with rendering not pausing for buffering to finish for `<Video />` elements, resulting in the first frame being black. But it worked on `<OffthreadVideo />` so tried replacing it throughout the app but had this type error of `loop` missing from `<OffthreadVideo />`

Just-Moh-it:
Fixes #3762


JonnyBurger:
Thanks!

PR #3801, opened 2024-04-28
`@remotion/lambda`: Bundle function with Bun and optimize build for speed

JonnyBurger:


JonnyBurger:
Doesn't work on Windows -> no-go

PR #3813, opened 2024-04-30
`@remotion/docs`: WIP - Recorder Documentation 

patsalv:


JonnyBurger:
@patsalv I figured out why dark mode doesn't work: It uses the system dark mode, but Docusaurus can override it. We can use the useColorTheme() hook in Docusaurus instead

PR #3830, opened 2024-05-03
fix startInSeconds in convert-to-captions

philipheinser:
before all startInSeconds values where the same all -1 now they correct


philipheinser:
never mind seems to be an upstream issue https://github.com/ggerganov/whisper.cpp/issues/2036

JonnyBurger:
In remotion.dev/recorder, we use this:

```tsx
const getTokenToTimestamp = (token: Token | undefined): number | null => {
  if (!token) {
    return null;
  }

  if (token.t_dtw === -1) {
    return token.offsets.from;
  }

  return token.t_dtw * 10;
};

```

PR #3874, opened 2024-05-15
Add contact page

MehmetAdemi:
We don't have a contact page on https://www.remotion.dev/ 

It would make sense to add one. I started with it. As for now, it's more of a draft: 
https://www.remotion.dev/contact

Screenshot 2024-05-15 at 17 31 41


MehmetAdemi:
Here is the contact page I created. What do you think @JonnyBurger?
Image

JonnyBurger:
Love it!

PR #3881, opened 2024-05-17
`@remotion/lambda`: Allow configuring a VPC

timello:
Add VPC config support so remotion render lambda can download assets internally from the VPC.

With the VPC config option, the assets can be downloaded via an internal endpoint. That increases security and allows cost reduction. One clear use case is downloading assets from an S3 private bucket via an S3 VPC endpoint.


JonnyBurger:
Thanks a lot for this! This was done super carefully, which I appreciate a lot!

Does this already unlock additional functionality! Remotion Lambda only downloads via HTTP, or have you tricked it into doing something different? Or do you pass the internal endpoint as the `src` for assets?

I know the internal endpoint is interesting, and we might allow proxying S3 assets natively in the future.

timello:
> Thanks a lot for this! This was done super carefully, which I appreciate a lot!
> 
> Does this already unlock additional functionality! Remotion Lambda only downloads via HTTP, or have you tricked it into doing something different? Or do you pass the internal endpoint as the `src` for assets?
> 
> I know the internal endpoint is interesting, and we might allow proxying S3 assets natively in the future.

Indeed, @JonnyBurger , we do use internal endpoints as the `src` for assets to avoid significant expenses.

PR #3897, opened 2024-05-21
what if we used bun as package manager

JonnyBurger:
- Custom MDX components don't work anymore
- Need to update contribution guidelines


JonnyBurger:
Almost works, but Docusaurus dependencies are hoisted, leading to breakage. Will revisit!

PR #3923, opened 2024-05-26
Render a video on the homepage

JonnyBurger:
Old branch, making it a WIP PR now


MehmetAdemi:
Looking at the Vercel's preview:
For some reason, the player is taking the whole width of the browser window. Is this intentional? It also effects other elements on the homepage. Here a screenshot:
image

JonnyBurger:
Here are some things that can be done as next step: @UmungoBungo 

- Show a loading state (right now therer is a layout shift when loading the page)
- During rendering, only render the animated emoji that is selected (using `getRemotionEnvironment()`)
- During rendering, don't show the arrows for selecting an emoji
- Store the order of the cards and the selected emoji as input prop so that we can pass it while rendering
- Handle a data fetching issue gracefully (show an empty card with an error message)

Meanwhile I am working on setting up the backend infrastructure so that soon we can start rendering a video!

JonnyBurger:
One more thing:

We are making a call to `https://bugs.remotion.dev/trending?lng=9&lat=8&country=US` and now it also returns the country name and SVG paths.

Which means we can use this and get rid of all the SVG paths that we currently bundle, making the site faster.

@UmungoBungo

PR #3957, opened 2024-06-03
assign proper start index for 'input range' arg of interpolate styles function, fixes #3922

### LINKED ISSUE #3922: `interpolateStyles` throws `Cannot convert undefined or null to object`
# Bug Report 🐛

interpolateStyles throws a weird error when I use the following block to style one of the images in my composition:

```tsx
	const lastImageExitStyle = interpolateStyles(
		frame,
		[eachImgDur * 2, eachImgDur * 3],
		[{opacity: 1}, {opacity: 0}]
	);
```

### What I wanted to achieve:
1. Fade in effect for when the first image shows up
2. Fade out for the last image when the video ends

### Expected output
A more helpful error about what's wrong :)

### Current output
```
Cannot convert undefined or null to object
```

Here's my composition:
Repro:
https://stackblitz.com/edit/github-bpjydj?file=src%2FRoot.tsx,src%2FComposition.tsx

```tsx
export const MyComposition = () => {
	const frame = useCurrentFrame();
	const {durationInFrames, fps} = useVideoConfig();

	const firstImageEnterStyle = interpolateStyles(
		frame,
		[0, fps / 2],
		[{opacity: 0}, {opacity: 1}]
	);

	const lastImageExitStyle = interpolateStyles(
		frame,
		[eachImgDur * 2, eachImgDur * 3],
		[{opacity: 1}, {opacity: 0}]
	);
	
	return ( <> a bunch of images here </> );
};
```

abhishek818:
root cause of this issue is the scenario when input arg's value is less than the first element of input range array. Invalid starting index for input range was getting assigned in this case.

/claim #3922


abhishek818:
@JonnyBurger , up for review..

JonnyBurger:
Nice! That solves it, thanks for demonstrating it with a test 🙂

PR #4060, opened 2024-07-09
`@remotion/renderer`: Correctly enable multi-threading for frame extractor

JonnyBurger:
…actor


JonnyBurger:
Will merge next Monday, as this is a bit of a risk and I will be more ready to respond if something goes wrong

PR #4067, opened 2024-07-09
fix: Resolve the issue where CJK-named composition IDs are not automatically reselected after page refresh in Remotion Studio.

hylarucoder:
Resolve the issue where CJK-named composition IDs are not automatically reselected after page refresh in Remotion Studio.

image


JonnyBurger (commented):
This is great! Thanks for sending a PR our way 💙

JonnyBurger (approved):
This is great! Thanks for sending a PR our way 💙

PR #4076, opened 2024-07-10
Ensuring movflags faststart is applied correctly when using Remotion Lambda

DaveMiscampbell:


packages/renderer/src/combine-video-streams.ts
```diff
@@ -63,8 +63,8 @@ export const combineVideoStreams = async ({
 		encoder,
 		codec === 'h265' ? '-tag:v' : null,
 		codec === 'h265' ? 'hvc1' : null,
-		codec === 'h264' ? '-movflags' : null,
-		codec === 'h264' ? 'faststart' : null,
+		codec === 'h264' || codec === 'h264-ts' ? '-movflags' : null,
```
DaveMiscampbell: I'm not sure if there are more possibilities here that we need to handle

packages/renderer/src/mux-video-and-audio.ts
```diff
@@ -39,6 +39,8 @@ export const muxVideoAndAudio = async ({
 		videoOutput ? 'copy' : null,
 		audioOutput ? '-c:a' : null,
 		audioOutput ? 'copy' : null,
+		'-movflags',
```
DaveMiscampbell: I appreciate this isn't ideal because the output codec may not support these flags but it wasn't immediately obvious to me how to check that from this point

packages/renderer/src/stitch-frames-to-video.ts
```diff
@@ -359,7 +359,7 @@ const innerStitchFramesToVideo = async (
 			x264Preset,
 			colorSpace,
 		}),
-		codec === 'h264' ? ['-movflags', 'faststart'] : null,
+		codec === 'h264' || codec === 'h264-ts' ? ['-movflags', 'faststart'] : null,
```
DaveMiscampbell: Again, I'm not sure if there are more possible codec values we need to handle here

JonnyBurger:
Thanks!

I think we only need to add this flag in the final assembly, no need to spam the flag.
I just tested with h264 and h265 output formats, and it works!

Empirically I also tested ProRes, and it doesn't work.
Containers other than MP4 and MOV don't support Faststart anyway.

DaveMiscampbell:
Thanks @JonnyBurger - that all makes sense to me - appreciate you taking a look and merging

JonnyBurger:
Appreciate you too!

PR #4087, opened 2024-07-16
Fix: save media to another bucket (python)

teyou:
- RenderMediaParams to be able to pass OutNameInputObject and follow the same naming 
- add "no-acl" privacy


JonnyBurger:
Thanks a lot for this PR! Very nice, TIL about `Union`!

Issue #4079, opened 2024-07-12
Remotion + Rive + Referenced Assets


zackphilipps:
# Feature Request 🛍️

@JonnyBurger Thanks for the Rive support in #4069! I'm curious what the best approach would be for using Rive + Remotion with [referenced or dynamic assets](https://rive.app/community/doc/loading-assets/doct4wVHGPgC#referenced-assets)?

I see that `ref` support was added with the above PR, but I'm unclear on the use case of that or if that could be leveraged here.

I can make it work with [`useRive`](https://codesandbox.io/s/rive-out-of-band-images-react-y5fmx6?file=/src/App.js), but since I'm not using RemotionRiveCanvas, that doesn't sync Rive to Remotion's clock. Is there an easy solution for that?

## Use Case

For our use case, we're trying to make images dynamic, and we'll eventually need to do the same thing with a custom typeface.

**Update:** We also need `useRive` support for [dynamic text](https://rive.app/community/doc/text/docn2E6y1lXo).

## Possible Solution

Not sure.

JonnyBurger:
You can already change the text run using `ref.current.getArtboard()`: https://rive.app/community/doc/text/docn2E6y1lXo#low-level-api-usage

I'm adding support for custom asset loading in the next version!
Refer to the updated docs.

zackphilipps:
Hey @JonnyBurger, thanks for this. I was able to get it almost working with the changes in #4091. However, I'm still running into this issue, and I have no idea why it's happening:

![image](https://github.com/user-attachments/assets/29b53f6b-a06c-4657-ad1f-07348f777290)

Here's the relevant code:

```tsx
import {useCallback, useRef} from "react";
import {staticFile} from "remotion";
import {RemotionRiveCanvas, RiveCanvasRef} from "@remotion/rive";
import {FileAsset, ImageAsset} from "@rive-app/canvas-advanced";
import {decodeImage} from "@rive-app/react-canvas";

export const Scene02 = () => {
    const ref = useRef<RiveCanvasRef>(null);

    const assetLoader = useCallback((asset: FileAsset, bytes: Uint8Array) => {
        if (!ref.current) return false;

        console.log("Asset properties to query", {
            name: asset.name,
            fileExtension: asset.fileExtension,
            cdnUuid: asset.cdnUuid,
            isFont: asset.isFont,
            isImage: asset.isImage,
            isAudio: asset.isAudio,
            bytes,
        });

[...]

PR #4091, opened 2024-07-17
`@remotion/rive`: Correctly pass `assetLoader` param

zackphilipps:
This PR is a follow-up to #4081. It correctly passes the `assetLoader` param as indicated [here](https://github.com/rive-app/rive-wasm/blob/2cad929c47491a1a338ccac0c10a592685bf8aa0/js/src/rive.ts#L1349-L1354).

Related issue: https://github.com/remotion-dev/remotion/issues/4079


JonnyBurger:
Thank you for sending a PR! This seems sound.

PR #4147, opened 2024-08-02
`remotion`: new `onVideoFrame` callback allowing an OffthreadVideo to be drawn to a canvas

JonnyBurger:
How to draw OffthreadVideo to canvas: https://www.remotion.dev/docs/video-manipulation

### DIFF
--- a/packages/docs/components/GreenscreenExamples/index.tsx
+++ b/packages/docs/components/GreenscreenExamples/index.tsx
@@ -44,7 +44,7 @@ export const Greenscreen: React.FC<{
 				<OffthreadVideo
 					style={{opacity: 0}}
 					onVideoFrame={onVideoFrame}
-					src="https://remotion-assets.s3.eu-central-1.amazonaws.com/just-do-it.mp4"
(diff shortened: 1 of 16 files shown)


packages/docs/components/GreenscreenExamples/index.tsx
```diff
@@ -1,187 +1,150 @@
...
+		[height, width],
+	);
+
+	return (
+		<AbsoluteFill>
+			<AbsoluteFill>
+				<OffthreadVideo
+					onVideoFrame={onVideoFrame}
+					style={{opacity: 0}}
+					startFrom={300}
```
khanzzirfan, line 85: This works nicely if startFrom > 0. Consider a situation when startFrom=0 and there are multiple sequences and generating  video using Lambda. 
if startFrom=0, generally canvas sync will show blankscreen, since the video is not started. To remediate we have to offset 0th second to 0.1 frame to show the initial start screen to sync with canvas and avoid having the empty screen.

JonnyBurger: @khanzzirfan It has nothing to do with startFrom, it works with any value.

I only put 300 as value because the video starts with a black frame.

PR #4169, opened 2024-08-07
`@remotion/cloudrun`: Fix cloudrun and Debian docker image

JonnyBurger:
Cloud Run was prone to breaking with Chrome upgrades.
Now pinning and using Chrome Headless Shell.

### DIFF
--- a/packages/cloudrun/container/Dockerfile
+++ b/packages/cloudrun/container/Dockerfile
@@ -1,13 +1,28 @@
 FROM node:20-bookworm
 
-RUN apt-get update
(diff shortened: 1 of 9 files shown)


ssijak:
@JonnyBurger hey, I was testing cloudrun, and the whole setup went well without issues, but the actual run command just timeseout. Tried it on a new hello-world project too, same. Here is the log output (504 is when it timesout), seems like nothing is happening

CleanShot 2024-08-08 at 00 32 18@2x

JonnyBurger:
@ssijak The PR was not released yet, but I'll do that shortly!
Then it should work again

PR #4231, opened 2024-08-23
Add support for captions/subtitle generation via transformers.js

D4ve-R:
Add support for captions/subtitle generation via transformers.js (WASM/WebGPU)

https://github.com/xenova/transformers.js/tree/v3/examples/whisper-word-timestamps


JonnyBurger:
🔥🔥

JonnyBurger:
Seems like this never materialized and is not functional in the current state, therefore closing.

Good idea for a module though!

PR #4235, opened 2024-08-25
Improve Docker instructions

JonnyBurger:


Antonwy:
Screenshot 2024-08-25 at 23 46 37@2x

Step 3 should be probably removed...

JonnyBurger:
Thanks for pointing this out! @Antonwy

PR #4237, opened 2024-08-25
Remove unused `path` import

Et3rnel:
The `path` import is not used in the example.
I think it can be removed safely.


JonnyBurger:
Indeed! Thanks a lot! 🙌🏼

PR #4244, opened 2024-08-27
fix(lambda): Always parse arguments as string

mecirmartin:
This PR fixes the fact that minimist is parsing numeric arguments as numbers. Remotion expects strings (in internal TS types).
I've found this fix while passing numeric ids like `1` to `npx remotion lambda render <SITE_URL> 1`.
This results in `Error: Could not find composition with ID 1. Available compositions: 1, 2`

Based on docs of minimist (https://github.com/minimistjs/minimist):
```
Numeric-looking arguments will be returned as numbers unless opts.string or opts.boolean contains that argument name. To disable numeric conversion for non-option arguments, add '_' to opts.string.
```

This PR disables implicit type conversion of numeric arguments to javascript number type.

Please test if this really fixes the issue, i'm not familiar enough with remotion to test this myself

### DIFF
--- a/packages/cli/src/benchmark.ts
+++ b/packages/cli/src/benchmark.ts
@@ -320,7 +320,7 @@ export const benchmarkCommand = async (
 
 	const ids = (
 		remainingArgs[0]
-			? remainingArgs[0]
+			? String(remainingArgs[0])
 					.split(',')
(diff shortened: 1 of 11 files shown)


mecirmartin:
Excelent, thank you

PR #4254, opened 2024-08-30
Homepage update

MehmetAdemi:
This is still a WIP.

I created 3 new components on the homepage. All of them are working based on functionality and responsiveness. However, they need to be refactored. 

- TrustedByBanner.tsx
Might need a better approach to display the logos, but it could be more responsive regarding scaling the window.

- EvaluateRemotionSection.tsx
It should be fine.

- DeveloperCommunityStats.tsx
It needs a refactor since all styles and sub-components are in one file. I tried to separate it, but I had issues with how the styles applied.


JonnyBurger:
It seems like the Evaluate section is good to ship, great job!
The rest I disabled at the moment, it can be reactivated and worked on a new branch.

JonnyBurger (approved):

PR #4283, opened 2024-09-05
Add community stats

MehmetAdemi:
The stats appear well. However, I would like to make them clickable, but adding a <a> tag messes up the layout. Need to work on this.


JonnyBurger:
Section itself looks good, I am just wondering about the placement.
Inbetween pricing and evaluation call seems odd.

Maybe it should be above pricing?

MehmetAdemi:
Experience-wise, I think this change could be made better. 


Which one do you prefer?

JonnyBurger:
Thanks!

PR #4284, opened 2024-09-05
Implement zGoogleFont zod type

### LINKED ISSUE #2692: implement zod type for google fonts
# Feature Request 🛍️
It would be beneficial to have a combobox or picker to select google fonts similar to zColor. This will improve the dev experience to easily select a font while development.
## Possible Solution
Extend https://www.remotion.dev/docs/zod-types/ to add a z-google-font with a zGoogleFont type

erayerdin:
Closes #2692.

This PR is a draft.


packages/zod-types/src/z-google-font.ts
```diff
@@ -0,0 +1,7 @@
+import {getAvailableFonts} from '@remotion/google-fonts';
```
JonnyBurger, line 1: Thanks a lot for starting this initiative!
Unfortunately I think it's not a good idea to use `@remotion/google-fonts` as a dependency.

`@remotion/zod-types` is a lightweight package and many people use it to for example show a color picker.
Due to it's massive database, `@remotion/google-fonts` is a 50MB package that most people will not even need in this case.

For a `zGoogleFont` type, we should use the Google Fonts package and write a script to generate a list of available fonts, then use that instead of depending on the whole package.

JonnyBurger (changes requested):

packages/zod-types/src/z-google-font.ts
erayerdin: Just out of curiosity and to clarify: Why do we even have to write a generator script here? I have checked `packages/google-fonts/scripts` and it does not seem like it fetches fonts from Google Fonts API. And, in this case, all we need (I presume) is to get the names of the fonts and that's all.

So, doesn't that mean I can just copy all `FontName.ts` files under `packages/google-fonts/src` into `packages/zod-types/src/z-google-fonts`? Or maybe a create a big `packages/zod-types/src/z-google-fonts/fonts.ts` and create something like:

```typescript
const fonts: string[] = [
  // ...
  'Roboto',
  'Ubuntu',
  // ...
];
```

JonnyBurger: The generator script does only retrieve metadata: List of available fonts, weights, styles, unicode ranges and most importantly the URLs at which it should fetch the actual fonts. The metadata powers also the TypeScript type completion and the `getInfo()` APIs. Yes even this data is massive.

I would say yes, you can just copy all font names into an array. Then the developer can itself retrieve the metadata from that and load fonts.

JonnyBurger:
closing due to staleness

PR #4294, opened 2024-09-10
[docs] Update resources with 12+ Remotion Example

StevePhuc:
12+ Remotion Example  https://remotion-example.vercel.app/


JonnyBurger:
Wow, this is super valuable! Thanks for making this!

JonnyBurger (approved):

PR #4296, opened 2024-09-11
Docs: Add React Video Editor as a Template

sambowenhughes:
Contains:
- Added the React Video Editor as a template under resources


JonnyBurger:
Nice, didn't even realize there is a light open source version!

PR #4331, opened 2024-09-27
fix: signal is aborted without reason

andrzej-woof:
Fixes issue when `free()` is called inside different `useEffect`

CleanShot 2024-09-27 at 10 59 43

### DIFF
--- a/packages/core/src/prefetch.ts
+++ b/packages/core/src/prefetch.ts
@@ -202,7 +202,7 @@ export const prefetch = (
 				canceled = true;
 				if (canBeAborted) {
 					try {
-						controller.abort();
+						controller.abort('free() called');
 					} catch (e) {}
 				}
 			}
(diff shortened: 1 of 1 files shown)


JonnyBurger:
Nice, thanks! TIL

Should the reason be an `Error` object instead of a `string` maybe?

https://dom.spec.whatwg.org/#ref-for-dom-abortsignal-reason%E2%91%A0 seems to think so because if there is no reason, it uses an `Exception` which is an `Error`

If a string is thrown, then one could have a problem in their `.catch` handler if they do `err.message.includes`

andrzej-woof:
@JonnyBurger you're obviously correct, I just made a quick patch for myself without giving it much thought.
To be fair a nicer solution would be to have default reason and maybe expose passing custom reason as an argument `free(reason)`

JonnyBurger:
Nice, thanks!

Issue #4355, opened 2024-10-03
`crossOrigin` prop not being passed to video element when rendering the `<Video />` component using `startFrom` or `endAt` props.


gdaunton:
With the change made in https://github.com/remotion-dev/remotion/pull/4248 we are no longer getting `crossorigin` applied to certain video elements. This is because there is a different render function for `<Video />` components that are given the `startFrom` or `endAt` props. [That render function](https://github.com/remotion-dev/remotion/blob/main/packages/core/src/video/Video.tsx#L111) does not include the same props as the [base render function](https://github.com/remotion-dev/remotion/blob/main/packages/core/src/video/Video.tsx#L147). For us that means our videos aren't being given the proper `crossOrigin` prop 

# Bug Report 🐛
To reproduce this issue
```tsx
<Video
  // ... required video props
  startFrom={123}
  crossOrigin="anonymous"
/>
```
observe that the `crossorigin` attribute is not applied to the resulting video element.

PR #4358, opened 2024-10-03
Fix crossOrigin prop not being passed to VideoForPreview.

gdaunton:
resolves https://github.com/remotion-dev/remotion/issues/4355

this pr ensures that the `crossOrigin` prop will always make it to the underlying video element.
Currently if passing either the `startFrom` or `endAt` props the `crossOrigin` prop will be dropped along with the `showInTimeline` and `onAutoPlayError` props.

I also made a second commit to propose a refactor to make the prop destructuring in this file easier to read. The first commit only fixes the `crossOrigin` prop being dropped but is a much smaller and less risky change.

### DIFF
--- a/packages/core/src/video/Video.tsx
+++ b/packages/core/src/video/Video.tsx
@@ -34,7 +34,6 @@ const VideoForwardingFunction: React.ForwardRefRenderFunction<
 		_remotionInternalNativeLoopPassed,
 		showInTimeline,
 		onAutoPlayError,
-		crossOrigin,
 		...otherProps
 	} = props;
 	const {loop, _remotionDebugSeeking, ...propsOtherThanLoop} = props;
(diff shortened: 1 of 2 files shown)


packages/core/src/video/Video.tsx
```diff
@@ -85,15 +80,19 @@ const VideoForwardingFunction: React.ForwardRefRenderFunction<
 				name={name}
 			>
 				<Video
-					{...propsOtherThanLoop}
+					{...loopedVideoProps}
```
JonnyBurger: Previously `startFrom` and `endAt` were passed, but now they are not anymore.

Looks like this could be a regression.
Can we make the PR more minimal so that only `crossOrigin` is affected?

JonnyBurger (changes requested):
Thanks a lot for filing an issue and sending a solution too!
I only have one doubt still.

JonnyBurger (approved):

JonnyBurger:
Thanks a lot! 🙌🏼

Issue #4367, opened 2024-10-05
Export `BufferingProvider`?


bennobuilder:
# Feature Request 🛍️

The `BufferingContextReact` required by `useMediaPlayback` is currently not provided by Vidstack (Player), causing the following error:
```
Uncaught Error: useMediaPlayback must be used inside a <BufferingContext>
```

## Use Case

Make Remotion work with Vidstack (Remotion Provider) by ensuring the `BufferingContextReact` is available for `useMediaPlayback`.

## Possible Solution

1. [TEMP] Manually export `BufferingProvider` in `node_modules` or automate this via a script.
2. Provide an official export of `BufferingProvider` from the Remotion `core` package to ensure `BufferingContextReact` is accessible in `useMediaPlayback`.

```tsx
<BufferingProvider>
	<MediaPlayer
		src={src}
		ref={mediaPlayerRef}
	>
		<MediaProvider loaders={[RemotionProviderLoader]} />
	</MediaPlayer>
</BufferingProvider>
```

JonnyBurger:
This does not seem the right way to fix it.
Both `useMediaPlayback` and BufferingProvider is internal API from Remotion.

It seems like Vidstack does not use it either, so I think the actual problem here is a version mismatch!
If multiple versions of Remotion are in a project, then the React contexts don't work anymore.

Can you check your lockfile if this is the case?

bennobuilder:
@JonnyBurger Good point 👍 

But although I use the same Remotion Version across the packages, it still can't identify the `BufferingContextReact` 🤷 

![Image](https://github.com/user-attachments/assets/fb5181a5-eed4-4f5a-aedd-141c4281ea8b)

Remotion Versions in use (Verified via Dependency Tree)
- `@repo/web` Remotion Version: `4.0.193`
- `@repo/video` Remotion Version: `4.0.193`
- `@vidstack/react` Remotion Version: `4.0.193` [[Dev Dependency]](https://github.com/vidstack/player/blob/main/packages/react/package.json)

<details>

<summary>Dependency Tree (PNPM, Depth: 2)</summary>

```
dependencies:
@blgc/utils 0.0.20
@radix-ui/react-icons 1.3.0
└─┬ react 18.3.1 peer
  └── loose-envify 1.4.0
@radix-ui/react-scroll-area 1.2.0
├── @radix-ui/number 1.1.0
├── @radix-ui/primitive 1.1.0
├─┬ @radix-ui/react-compose-refs 1.1.0
[...]

PR #4368, opened 2024-10-05
#4367 Export BufferingProvider

bennobuilder:
#4367


JonnyBurger:
Thanks for sending a PR right away! As per https://github.com/remotion-dev/remotion/issues/4367#issuecomment-2396282546 I'd like to solve it in another way.

PR #4370, opened 2024-10-07
Gt planar font in demo

UmungoBungo:
Had an issue getting this to actually build and display due to the captions docs throwing errors. But changing the value in devtools was fixing the fonts.

image


UmungoBungo:
wow excuse my noob moment, forgot how a monorepo even works 🤦🏻 

I built all of the packages, restarted the docs server, it is all G

image

UmungoBungo:
I see GT Planar is not available on Google Fonts. What considerations need to be made for this to work when rendering?

JonnyBurger:
Nice!
If we don't already do it, we need to load the font into the video also with `FontFace` / `@remotion/fonts` APIs.
We cannot just rely on the font loading on the website.

PR #4409, opened 2024-10-15
Prevent "NotAllowedError: play() failed"

UmungoBungo:
While the audio has not been requested (un-mute hasn't been clicked, or the volume hasn't been changed) don't mount an audio component. If rendering, mount it regardless.

This is to prevent this browser message

image


UmungoBungo:
Considering the audio tag is mounted just at the moment it is required for playback, I have added a `preload()` outside of the `<Player />`. I also trimmed the audio, bringing the file size down from 1.76mb to 60kb

PR #4413, opened 2024-10-17
`@remotion/studio`: Fix problematic Zod schema

### LINKED ISSUE #4157: Problematic Zod schema
These dropdowns don't work well in Remotion Studio:

```tsx
import {z} from 'zod';
import {themeSchema} from './theme';

export const width = z.discriminatedUnion('type', [
	z.object({
		type: z.literal('auto'),
	}),
	z.object({
		type: z.literal('fixed'),
		value: z.number().min(1080),
	}),
]);

export const schema = z.object({
	theme: themeSchema,
	width,
});

```

avanderbergh:
Fixes #4157

In the case where a `min` check exists on a `Number` type, we use that value instead of `0`.


packages/studio/src/components/RenderModal/SchemaEditor/create-zod-values.ts
```diff
@@ -17,7 +17,9 @@ export const createZodValues = (
 		case zodRuntime.ZodFirstPartyTypeKind.ZodString:
 			return '';
 		case zodRuntime.ZodFirstPartyTypeKind.ZodNumber:
-			return 0;
+			return (
+				(def as z.ZodNumberDef).checks.find((c) => c.kind === 'min')?.value ?? 0
+			);
```
avanderbergh: @JonnyBurger This fixes this particular issue with the `min()` or `gte()`. Would it make sense to also cover other checks like `gt()` and `positive()`, `multipleOf()`, etc...?

And then would it make sense to also extend the same method to other types like `string` and `BigInt` for example?

JonnyBurger: I thought about it and the only one I can think of is `max && max.value < 0`.

Otherwise, all possible checks should pass with `0`.

JonnyBurger:
Thanks a lot! Very valuable.

JonnyBurger (approved):

JonnyBurger (approved):

PR #4426, opened 2024-10-24
Building a Timeline: Fix the dependency array in useMemo hook

ADTC:
Without the dependency, the callback will never return the updated tracks.


JonnyBurger:
👑

Thanks king!

PR #4432, opened 2024-10-27
Correct behaviour for volume control

UmungoBungo:


UmungoBungo:
I've changed the animation for the country card, as scaling causes text to jitter:

| Before  |  After |
| ------------- | ------------- |
| slideJitter2 | slideText2  |

PR #4439, opened 2024-10-30
[Docs update] Custom controls - fixes and improvements to sample code & docs

ADTC:
- It's "full control" not "full controls".
- Improved listing of prerequisites for custom controls outside the player.
- `SVGProps` is unused. Removed.
- Fix initial state of Play/Pause button. It has to update when `playerRef` changes.
- `TimeDisplay` was not exported. Fixed.
- Name of some components made consistent with the file name.
- `durationInFrames - 1 ?? 0` is invalid since the left hand side is never nullish. Use `Math.max` instead to enforce `>= 0`.
- `inFrame` and `outFrame` props are optional in `<Player/>` so they should be optional in `<SeekBar/>` as well.
- `VolumeSlider.tsx` file name was missing. Fixed.

### DIFF
--- a/packages/docs/docs/player/custom-controls.mdx
+++ b/packages/docs/docs/player/custom-controls.mdx
@@ -33,16 +33,20 @@ Use the following APIs to customize the individual controls:
 Use this approach if you:
 
 - Want to implement custom controls anywhere on the page
-- Want full controls over the look and behavior of the controls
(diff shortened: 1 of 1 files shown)


JonnyBurger:
Wow thank you! Immensly helpful! 🤗

PR #4447, opened 2024-11-01
Experiment: Migrate to Rspack

JonnyBurger:


JonnyBurger:
We're still gonna do it, but start off clean and do it a bit different :)

PR #4460, opened 2024-11-04
Docusaurus 3.6 with Rspack

JonnyBurger:


packages/docs/docusaurus.config.js
```diff
@@ -13,6 +13,9 @@ const config = (mode) => ({
 	favicon: 'img/favicon.png',
 	organizationName: 'remotion-dev', // Usually your GitHub org/user name.
 	projectName: 'remotion', // Usually your repo name.
+	future: {
+		experimental_faster: true,
+	},
 	webpack: {
 		jsLoader: (isServer) => ({
 			loader: require.resolve('swc-loader'),
```
slorber, line 21: you can remove this, this is used instead of the built-in Rspack SWC loader and actually slow things down by introducing a "js bridge" in between 2 rust things 😅

JonnyBurger: Thanks! Testing with https://github.com/remotion-dev/remotion/pull/4462

PR #4485, opened 2024-11-07
docs: fix link to X264 preset CLI option doc

rbarbazz:
This PR fixes an incorrect internal link I noticed while reading the docs.


JonnyBurger:
Thanks a lot! Good catch!

Issue #4487, opened 2024-11-08
Memory leak in printUsefulErrorMessage


liyipeng3:
My code runs as a service and when over time alreadyPrinted expands indefinitely, there doesn't seem to be any failing logic in the code, is there any optimization that can be done here or is there anything I can do about this to avoid it?

# Bug Report 🐛

Code in packages/renderer/src/print-useful-error-message.ts

![Image](https://github.com/user-attachments/assets/369607ae-d812-462a-b155-797115172344)

![Image](https://github.com/user-attachments/assets/4924c803-af02-4a7a-be69-d08a258a0c69)

PR #4499, opened 2024-11-11
`@remotion/renderer`: Fix memory leak in `printUsefulErrorMessage()`

liyipeng3:
Trying to fix memory leaks in #4487


JonnyBurger:
Thanks a lot for taking this into your hand!
I think a more simple solution is to just keep the last 10 errors at most.

The previous implementation was buggy too in that the error message would be the same, but the reference would be different. So with this it should also be fixed.

JonnyBurger:
Sorry, this will only be out in v4.0.230, I missed to merge it in time! In 1-2 days this will be out

Issue #4572, opened 2024-11-28
can't install astro-example, 'not in registry'


tomByrer:
Win10, fresh node install

# Bug Report 🐛
```bash
> npm install @remotion/astro-example --save-exact
(node:11632) ExperimentalWarning: CommonJS module C:\Program Files\nodejs\node_modules\npm\node_modules\debug\src\node.js is loading ES Module C:\Program Files\nodejs\node_modules\npm\node_modules\supports-color\index.js using require().
Support for loading ES Module in require() is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
npm error code E404
npm error 404 Not Found - GET https://registry.npmjs.org/@remotion%2fastro-example - Not found
npm error 404
npm error 404  '@remotion/astro-example@*' is not in this registry.
npm error 404
npm error 404 Note that you can also install from a
npm error 404 tarball, folder, http url, or git url.
npm error A complete log of this run can be found in: C:\Users\tomby\AppData\Local\npm-cache\_logs\2024-11-28T21_27_32_690Z-debug-0.log
PS C:\Users\tomby\OneDrive\Documents\github\react\remotion\astro-remotion> cd ..
PS C:\Users\tomby\OneDrive\Documents\github\react\remotion> npm install @remotion/astro-example --save-exact
(node:1248) ExperimentalWarning: CommonJS module C:\Program Files\nodejs\node_modules\npm\node_modules\debug\src\node.js is loading ES Module C:\Program Files\nodejs\node_modules\npm\node_modules\supports-color\index.js using require().
Support for loading ES Module in require() is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
npm warn ERESOLVE overriding peer dependency
npm warn While resolving: rollup-plugin-banner@0.2.0
npm warn Found: peer eslint-plugin-node@">=7.0.0" from eslint-config-standard@12.0.0
npm warn node_modules/.registry.npmjs.org/rollup-plugin-banner/0.2.0/node_modules/rollup-plugin-banner/node_modules/eslint-config-standard       
npm warn   dev eslint-config-standard@"^12.0.0" from rollup-plugin-banner@0.2.0
npm warn   node_modules/.registry.npmjs.org/rollup-plugin-banner/0.2.0/node_modules/rollup-plugin-banner
npm warn
npm warn Could not resolve dependency:
[...]

PR #4577, opened 2024-11-29
Remove installation instructions from internal packages

tomByrer:
fixes https://github.com/remotion-dev/remotion/issues/4572


JonnyBurger:
Thanks, this was generated by a script, so I also updated the script

PR #4582, opened 2024-11-30
add quote around paths to take into account if there are spaces in file paths

oga-sven:
Just tried the tiktok template. I have a space in my path where the project is installed so when i ran sub.mjs i got an error. Adding a quote around it will avoid this issue


JonnyBurger:
Thank you! 😃🙌

PR #4623, opened 2024-12-11
`remotion`: Don't store OffthreadVideo frames in cache

JonnyBurger:


mxmisterdev:
Hey @JonnyBurger, just for curiosity, wouldn't this make the rendering slower when using looped video?

PR #4627, opened 2024-12-11
Docs: Add missing libcups2 dependency for Debian

acj:
We were updating some dependencies that seemed unrelated to Remotion but started getting a "Failed to launch the browser process!" error, which we traced to the libcups2 package. It was added to the Debian dockerfile [here](https://github.com/remotion-dev/remotion/commit/25cb4ff055304adb4f3147dd7a79a4ff811b8777#diff-cd61de8244d9267e03f2ee40c6e8a2f936dfaf543443ebb7b2da6e51ab709860R20) but not the docs. I don't know whether it's needed on Ubuntu.

Thanks!

### DIFF
--- a/packages/docs/docs/miscellaneous/linux-dependencies.mdx
+++ b/packages/docs/docs/miscellaneous/linux-dependencies.mdx
@@ -57,6 +57,7 @@ RUN apt install -y \
   libxdamage1 \
   libpango-1.0-0 \
   libcairo2 \
+  libcups2 \
   libatk-bridge2.0-0
 ```
 


JonnyBurger:
Thanks for sending this and sorry for the breakage.

Indeed, we forgot to add this to the docs.
Chrome is now managed by Remotion so no accidential upgrades start happening.

Probably you realized this and removed Chrome from your Docker image, but now this library is missing.

With your fix, everything should be pinned now going forward, so there should be no further breakage hopefully!.

JonnyBurger (approved):

acj:
We were removing `libvips` from our stack. Maybe it had a transitive dependency on libcups? Or maybe we were relying on cached image layers with different package versions that happened to work. It seems healthy now, anyway. 👍

PR #4667, opened 2024-12-24
Docs: Mark nixOS as unsupported

JonnyBurger:

### DIFF
--- a/packages/docs/docs/miscellaneous/linux-dependencies.mdx
+++ b/packages/docs/docs/miscellaneous/linux-dependencies.mdx
@@ -84,3 +84,8 @@ RUN yum install -y \
 ## Alpine Linux
 
 Not supported due to to unsupported Libc symbols.
+
+## nixOS
+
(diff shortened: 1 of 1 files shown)


tcurdt:
@JonnyBurger Is there some explanation somewhere why nixOS is explicitly unsupported?

JonnyBurger:
@tcurdt Remotion ships with Chrome/FFmpeg binaries. I think nixOS has a hard time executing binaries that are not part of the bootstrap. Help getting this to work is appreciated!

tcurdt:
@JonnyBurger Nice.

Nix sure makes the packaging a bit more complicated (or actually just different - maybe even easier) but since there are also [playwright packages](https://search.nixos.org/packages?channel=25.11&query=playwright&show=playwright) this should be very much doable.

Maybe it would be good to rephrase the docs a little so it sounds more like "looking for help to support it".
I see if I can give this a try - but my nix foo is still limited.

Thanks for the quick clarification!

PR #4676, opened 2024-12-29
`@remotion/openai-whisper`: Fix matching words which contain characters with special RegEx meaning

JonnyBurger:


thecmdrunner:
@JonnyBurger Thanks! 🙌

Issue #4712, opened 2025-01-09
Add `renderWebhook` support for `renderMediaOnCloudrun`


njho:
# Feature Request 🛍️

Add support for webhooks in renderMediaOnCloudrun to enable external progress tracking and custom data handling during the render process.

The renderUpdateWebhook would receive:
 - progress: number (0-1 indicating render progress)
 - renderId: string (unique identifier for the render)
 - data: any (custom user-provided data passed through)

## Use Case

* Saving `renderId` somewhere alongside progress and collating with an internal `jobId`
* External Progress Tracking
* Triggering notifications

## Possible Solution

```
import { renderMediaOnCloudrun } from "@remotion/cloudrun/client";

// Example custom data to pass through webhook
interface CustomWebhookData {
  projectId: string;
  userId: string;
  metadata: {
    campaign: string;
    version: string;
  };
}

async function renderVideo() {
  console.log("Making render request");

  // Example custom data that will be passed to the webhook
  const customData: CustomWebhookData = {
    projectId: "proj_123",
    userId: "user_456",
    metadata: {
      campaign: "summer_2024",
      version: "1.0.0",
    },
  };

  await renderMediaOnCloudrun({
    region: "us-east1",
    cloudRunUrl: "https://<anonymized-cloud-run-url>",
    serveUrl: "https://<anonymized-serve-url>",
    forceBucketName: "<anonymized-bucket-name>",
    composition: "my-video",
    codec: "h264",
    // Webhook for external progress tracking
    renderUpdateWebhook: {
      url: "https://api.your-service.com/render-updates",
      headers: {
        Authorization: `Bearer ${process.env.WEBHOOK_SECRET}`,
        "Content-Type": "application/json",
      },
      data: customData, // Custom data that will be included in webhook payload alongside `progress` and `renderId`
    },
  });
}

renderVideo().catch(console.error);

export {};
```

JonnyBurger:
Thanks! Merged!

Issue #4711, opened 2025-01-09
Adding Declarative `RenderId` Support for Cloud Run


njho:
# Feature Request 🛍️
I would like to have the ability to make the `renderId` declarative and provided as an arg to `renderMediaOnCloudRun`

## Use Case

This would allow individuals to avoid needing to await a long-running process and instead be able to poll for completion at the URL where the render is uploaded to.

## Possible Solution

Suggestion to add `renderId` to `renderMediaOnCloudRun` and the CLI

PR #4714, opened 2025-01-09
`@remotion/cloudrun`: Add `Webhook` and `renderId` support - #4711 and #4712

njho:
Please see both #4712 and #4711 

---
### Documentation (In PR as well)

**renderStatusWebhook**
Type: _optional_

Allows you to define a webhook URL that will receive updates on the progress of a render.

**Arguments**

- url: The endpoint to which progress updates will be sent.
- headers: An object containing headers to be included in the webhook request. Note: Content-Type: application/json is added automatically.
- data: Custom data to be sent alongside the progress updates.
- webhookProgressInterval (optional): Specifies how frequently the webhook will be called, as a percentage of progress completion. Defaults to 0.1 (10%). Minimum: 0.01 (1%), Maximum: 1 (100%).

**Example Config**
```
{
  "url": "https://example.com/webhook",
  "headers": {
    "Authorization": "Bearer 1234567890"
  },
  "data": {
    "projectId": "1234567890"
  },
  "intervalPercentage": 0.1
}
```

**Example Webhook**
```
{
  "progress": 0.1,
  "renderedFrames": 100,
  "encodedFrames": 100,
  "renderId": "1234567890",
  "projectId": "1234567890"
}
```

---

**renderIdOverride**
Type: _optional_

Allows you to provide a custom render ID for the render. If not specified, a random render ID will be generated automatically.

Note: It is your responsibility to ensure the provided render ID is unique. If the render ID is not unique, it may overwrite existing renders with the same ID.

---

### Potential Tradeoffs
* Declarative method allows for overwriting of the output if the user defines it incorrectly
* Processing overhead - mitigated through `webhookProgressInterval`


njho:
@JonnyBurger tests passing!

JonnyBurger:
Thanks a lot! 💙

I am going to merge this for now, since this is very helpful.

The long-term plan is to use the same runtime for Cloud Run and Lambda, so we can have feature parity.
I'm progressing already with this by having extracted the runtime into `@remotion/serverless`.

PR #4769, opened 2025-01-20
`@remotion/media-parser`: `parseAndDownloadMedia()` API

JonnyBurger:


packages/media-parser/src/internal-parse-media.ts
```diff
@@ -0,0 +1,327 @@
...
+		name,
+		contentType,
+		supportsContentRange,
+	} = await readerInterface.read({src, range: null, signal});
+
+	if (contentLength === null) {
+		throw new Error(
+			'Cannot read media without a content length. This is currently not supported. Ensure the media has a "Content-Length" HTTP header.',
+		);
+	}
```
ycmjason, line 55: @JonnyBurger 

I notice that this breaks support for data urls used as src in `<Audio>`. Is this intentional?

PR #4820, opened 2025-01-27
handle quiet audio waves

UmungoBungo:
When processing extremely quiet audio (max amplitude < 0.001), the waveform visualization would over-amplify the noise due to normalization. This would make silent or near-silent sections appear as prominent waves.

For example, when I'm splitting this audio out into individual stems, it is common that one of the instruments is not present. In this screenshot, there is no piano:
image

This now treats audio below this threshold as silence by returning zeros instead of normalizing. Here are what the stems look like now:
image

Note that I still place a dot for the waveform when amplitude is zero, but that is in the canvas logic.


JonnyBurger:
Thanks a lot! 🙌

PR #4837, opened 2025-01-31
Remotion Lambda Python client: Better error handling

JonnyBurger:


packages/lambda-python/remotion_lambda/remotionclient.py
```diff
@@ -103,16 +105,28 @@ def _parse_stream(self, stream):
...
 
         client = self._create_lambda_client()
-        response = client.invoke(
-            FunctionName=function_name, Payload=payload, )
-        result = response['Payload'].read().decode('utf-8')
-        decoded_result = self._parse_stream(result)[-1]
+        try:
+            response = client.invoke(
+                FunctionName=function_name, Payload=payload)
+            result = response['Payload'].read().decode('utf-8')
```
AliMirjahani, line 109: in case result is empty which happens when there is timeout it will get catch in line 122 which wouldn't make it better

JonnyBurger: @alimirjahani7 Actually in case of timeout, it would not fail here.

Empty payload is a different bug that was solved here: https://github.com/remotion-dev/remotion/pull/4850/files#diff-2e9ffa290280b3c6a071b88ad15c452e13e12c5946457a00e52dcb2dbc043077

JonnyBurger:
The empty response bug was also present in the JS SDK. Fixed now!
Merging, and if there is something to do improve after the next version, let me know!

PR #4872, opened 2025-02-07
Make timeline controls accessible on mobile

HiteshChoudhary0098:
UI update

Old UI 

Screenshot 2025-02-07 at 11 55 04 PM

New UI
Screenshot 2025-02-07 at 11 34 43 PM
Screenshot 2025-02-07 at 11 34 47 PM

Changes
1. Remove Timeline Zoom Slider in mobile
2. Show playback control on left.
3. Move all other controls to the right and make the control panel scrollable.

### DIFF
--- a/packages/studio/src/components/FpsCounter.tsx
+++ b/packages/studio/src/components/FpsCounter.tsx
@@ -11,6 +11,7 @@ const label: React.CSSProperties = {
 	color: 'white',
 	fontSize: 15,
 	fontFamily: 'Arial, Helvetica, sans-serif',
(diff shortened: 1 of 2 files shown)


JonnyBurger:
Thanks not bad! A few things that don't make sense for me yet:

1. The space here is not needed anymore since it was for aligning the controls in the center. It can be removed.
image

2. There is no indication that one can scroll to the right. Could we add a fade on both sides if it is scrollable? Could look like this: 
image

3. I suggest we remove the FpsCounter as it causes a layout shift when play/pause is toggled

PR #4895, opened 2025-02-15
Separate types exports from regular exports, bun was failing because of them.

happylinks:
Bun was failing on the types exports not being marked with `type`, so I fixed that for media-parser and webcodecs package, I can do it for others if you want, was just not using those yet.

Verified that `pnpm build` still works.


JonnyBurger:
Thanks!

PR #4915, opened 2025-02-21
fix city display in temperature card on promo page

fobos531:
I noticed that the city name in the Temperature card of the promo page was displayed incorrectly when the city name has some special characters. We can use a simple `decodeURIComponent` to fix this

Production:

CleanShot 2025-02-21 at 10 56 27@2x

Local:

CleanShot 2025-02-21 at 10 56 55@2x


JonnyBurger:
Thanks a lot for fixing this! 🙌

PR #4919, opened 2025-02-22
Upgrade to Chrome 133

### LINKED ISSUE #4918: Memory leak on MacOS when rendering using --gl=angle
The issue affects all users using Remotion 4.0.146 and above.

To reproduce, 
1. Clone the Remotion Three.js Starter template
2. Start rendering a video 

The RAM used by `chrome-headless-shell` will go up and not free up. This is a bug in `chrome-headless-shell`.

References:
- https://discord.com/channels/809501355504959528/817306238811111454/1298679270683443241
- https://discord.com/channels/809501355504959528/817306238811111454/1299581935177170976
- https://www.remotion.dev/docs/gl-options#using-the-gpu

samohovets:
Fixes #4918.

Tested for a week in production on https://banger.show.


samohovets:
@JonnyBurger should the Playwright version also be updated? I guess so.

JonnyBurger:
Thanks a lot for this!

Need to think a bit about it.
We want to also have the same version on Lambda at all times, but don't have a binary for it right now.

But I see the need for it, and maybe we should expose an option for it.

JonnyBurger:
Seems to work, thanks a lot!

PR #4948, opened 2025-02-27
Docs: Update OffthreadVideo while rendering snippet

tomh05:
Fix for code snippet that threw errors when I tried it in my project.

- Remove deprecated imageFormat field `'imageFormat' is deprecated`
- Refactor to satisfy the common eslintreact rule Function component is not a function declaration `(eslintreact/function-component-definition)`
- Video doesn't allow onAutoPlayError to be null, but Offthread video does, so convert it `Type '(() => void) | null | undefined' is not assignable to type '(() => void) | undefined'.`


JonnyBurger:
This a bug that was already fixed in https://github.com/remotion-dev/remotion/commit/03eb830a56ddee5148e8f8a9c2d625529b1d97d8

Happy to merge the change to function though, thank you!

JonnyBurger (approved):

PR #5012, opened 2025-03-14
Fix cloudrun service version mismatch error message

philippbosch:
The following error messages shows up, when the local `@remotion/cloudrun` package version is different from the one used in the service deployed to Google Cloud Run. 

> Error: Version mismatch: When calling renderMediaOnCloudRun(), you called a service, which has the version **`4.0.274`**, but the @remotion/cloudrun package you used to invoke the function has version **`4.0.274`**. Deploy a new service and use it to call renderMediaOnCloudrun().

But the same version variable (`VERSION`) was used in both places – for the local and the remote installation. This is fixed in this PR.


JonnyBurger:
Thank you, that is indeed another bug!

PR #5074, opened 2025-03-30
fix: select composition logLevel is not passed correctly

ycmjason:
I noticed that `selectComposition` will not respect the passed `logLevel`. Inspected the code and find out this issue!


JonnyBurger:
Oh, very good catch! Thanks for sending a PR!

PR #5108, opened 2025-04-04
Render server template

### LINKED ISSUE #5103: Create "Rendering server" template
Notes:
- express.js server with `/render`, `/status/:jobId`, `/cancel-render/:jobId`
- in memory jobs with keyv
- should be based on `template-helloworld`
- should bundle the composition before starting the server or during the build process
- should include Dockerfile

samohovets:
Fixes #5103 

**Features:**
- express.js render server
- dockerfile
- in-memory job queue
- start, track, cancel renders via endpoints

Notes:

- to support hot reload in dev command in `template-render-server`, I've replaced `ts-node` to `tsx` throughout the project. `validate-templates.test.ts` has also been updated to use tsx.


packages/create-video/src/templates.ts
```diff
@@ -184,6 +185,27 @@ export const FEATURED_TEMPLATES: Template[] = [
...
+	{
+		homePageLabel: 'Render Server (Express.js)',
+		shortName: 'Render Server',
+		org: 'remotion-dev',
+		repoName: 'template-render-server',
+		description: 'An Express.js server for rendering videos with Remotion',
+		longerDescription:
+			'A template that provides an Express.js server starting, tracking, and canceling Remotion renders.',
+		promoVideo: {
+			muxId: 'ccgoUm00vQBSZV01xKYTNMtxMiPiVJYyAFl1DSwuPBXb00',
```
JonnyBurger: Does not seem to be a valid Mux ID

Maybe for this template we should use an image instead of a video

packages/template-render-server/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+
+COPY . .
+
+RUN mkdir -p renders
+
+# Install browser
+RUN npx remotion browser ensure
+
+# Bundle the Remotion composition
+RUN npx remotion bundle remotion/index.ts
```
JonnyBurger: `remotion/index.ts` will be auto-detected

packages/template-render-server/package.json
```diff
@@ -0,0 +1,38 @@
...
+    "remotion:studio": "remotion studio"
+  },
+  "repository": {},
+  "license": "UNLICENSED",
+  "dependencies": {
+    "@remotion/bundler": "workspace:*",
+    "@remotion/cli": "workspace:*",
+    "@remotion/renderer": "workspace:*",
+    "@remotion/zod-types": "workspace:*",
+    "express": "4.21.0",
```
JonnyBurger: Can we use Express v5? https://expressjs.com/2024/10/15/v5-release.html

packages/template-render-server/package.json
```diff
@@ -0,0 +1,38 @@
...
+    "react": "19.0.0",
+    "react-dom": "19.0.0",
+    "remotion": "workspace:*",
+    "zod": "3.22.3"
+  },
+  "devDependencies": {
+    "@remotion/eslint-config-flat": "workspace:*",
+    "@remotion/eslint-plugin": "workspace:*",
+    "@types/express": "^5.0.1",
```
JonnyBurger: Let's pin it

```suggestion
    "@types/express": "5.0.1",
```

JonnyBurger (changes requested):

packages/template-render-server/package.json
samohovets: yep! updated

packages/template-render-server/package.json
samohovets: thanks, pinned

packages/template-render-server/Dockerfile
samohovets: nice!

packages/create-video/src/templates.ts
samohovets: fixed! that was an asset id, not playback id 🤦 

also fixed resolution

PR #5118, opened 2025-04-08
Add bufferutil to `optionalDependencies` to avoid resolve bug in Next.js API endpoint when using `selectComposition()`

### LINKED ISSUE #5117: `selectComposition()` in Next.js API route fails with `bufferutil` error
# Bug Report 🐛

Context: https://discord.com/channels/809501355504959528/1358797458112512111

**Steps to reproduce:**

1. Use Remotion's Next.js template
2. Put the code below to `/api/lambda/render` route:
```ts
import { selectComposition } from "@remotion/renderer";
import path from "node:path";

export const GET = async (req: Request) => {
  const bundlePath = path.resolve("./build");

  const composition = await selectComposition({
    serveUrl: bundlePath,
    id: 'MyComp'
  });

  return new Response("Hello, world!");
};
```
3. The endpoint will fail with `TypeError: bufferUtil.mask is not a function`

It seems like the `install` script in bufferutil isn't always triggering.

## Workaround

**If you're using pnpm v10+:**
 Make sure `bufferutil` is added to `onlyBuiltDependencies` in your package.json:

```json
"pnpm": {
    "onlyBuiltDependencies": [
      "bufferutil",
    ]
  }
```

**If you're using npm or yarn:**
Add `bufferutil` as an additional dependency:

```
npm install --save-optional bufferutil
```

samohovets:
Fixes #5117

### DIFF
--- a/packages/renderer/package.json
+++ b/packages/renderer/package.json
@@ -49,7 +49,8 @@
 		"@remotion/compositor-linux-arm64-musl": "workspace:*",
 		"@remotion/compositor-linux-x64-gnu": "workspace:*",
 		"@remotion/compositor-linux-x64-musl": "workspace:*",
-		"@remotion/compositor-win32-x64-msvc": "workspace:*"
+		"@remotion/compositor-win32-x64-msvc": "workspace:*",
+		"bufferutil": "4.0.9"
 	},
 	"keywords": [
 		"remotion",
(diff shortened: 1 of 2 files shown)


samohovets:
🤔

samohovets:
Closing this. To use `@remotion/renderer`, you should externalize it in next.config.js.
https://github.com/remotion-dev/remotion/issues/5117#issuecomment-2798873424

PR #5163, opened 2025-04-18
Do not publish .ts files

cofenster-michal:
@remotion/renderer publishes .ts files to NPM, some of them with internal paths (../monorepo)

### DIFF
--- a/packages/renderer/.npmignore
+++ b/packages/renderer/.npmignore
@@ -14,5 +14,4 @@ tsconfig.json
 install-toolchain.mjs
 toolchains
 toolchains.mjs
-*.ts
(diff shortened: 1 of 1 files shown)


JonnyBurger:
Actually, `.d.ts` files are TypeScript definition files, without them, you don't get any type-safety when using the package!

But indeed we can remove the `happydom.ts` file that you probably found in the package, it is a file only used for testing :)

JonnyBurger (approved):

JonnyBurger:
Ah, I misread! .d.ts files were not excluded! That makes sense then, thanks a lot for the PRè

PR #5171, opened 2025-04-22
Do not publish config files/build scripts to NPM.

cofenster-michal:
I've added some additional guards to npm ignore files.

I would suggest though changing from blacklisting to [whitelisting](https://stackoverflow.com/questions/15288712/gitignore-whitelist-on-directory-and-its-contents) pattern.


JonnyBurger:
Thank you!

I wanted to clean up the files in the package for a while but never got around to it. Appreciate this PR!

PR #5181, opened 2025-04-24
`@remotion/layout-utils`: use textTransform inside measureText

camilla-ferreri:
I noticed that `takeMeasurement` (used by `measureText`) accepts `textTransform` as a parameter, but it was never passed to the function. 
This caused `measureText` to return the same size for uppercase and lowercase texts. It also influenced the fontSize returned by `fitText`.


JonnyBurger:
Great catch! Thanks for noticing it and sending a PR!

PR #5201, opened 2025-04-28
fix(template-render-server): ensure sequential execution in render queue

DimplesY:
Fix the Promise chain in queueRender function to properly wait for each  processRender task to complete before starting the next one. This prevents  potential concurrent execution of render jobs that could lead to resource  contention and unexpected behavior.

The fix changes the callback from:

``` ts
queue.then(() => { processRender(jobId); })
```
to:
``` ts
queue.then(() => processRender(jobId))
```
This ensures the Promise returned by processRender is properly integrated  into the Promise chain.

### DIFF
--- a/packages/template-render-server/server/render-queue.ts
+++ b/packages/template-render-server/server/render-queue.ts
@@ -120,9 +120,7 @@ export const makeRenderQueue = ({
       },
     });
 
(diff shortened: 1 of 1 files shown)


JonnyBurger:
This is important! Thanks for fixing this oversight 🙌

PR #5203, opened 2025-04-28
ensure that input props are passed

alexfernandez803:
I've noticed that the inputProps are not being passed to the render, hence this is to fix it.


JonnyBurger:
Thank you 🙇‍♂️

PR #5220, opened 2025-05-01
Remove `entry` files from React Router template

samohovets:
The `entry.client.tsx` and `entry.server.tsx` files are **optional** in React Router ([docs](https://reactrouter.com/explanation/special-files)), and exposing them adds unnecessary maintenance overhead to the template.

These files are tightly coupled with React Router’s internal behavior and can change between versions, which would require us to manually keep them in sync across updates.

To avoid locking ourselves into a brittle setup, it’s better to omit them unless absolutely needed.


JonnyBurger:
I agree with you on this! 
Thanks for submitting a fix 🙌

JonnyBurger (approved):

PR #5221, opened 2025-05-02
docs: Updated create-tiktok-style-captions.mdx for better understanding

JustinBenito:
Explicit mention of how a custom dataset json for caption generation should be ( including spaces )

Personal Note: I was breaking my head over why my video was not rendering with a custom subtitle, but then after getting into the source code of createTikTokStyleCaptions. I understood that the spaces are a delimiter and modified the captions json file and voila it worked. I wish the next person not to undergo the same headache, as the current Note is not very explanatory and easy to scroll past it.


JonnyBurger:
Makes sense! Thank you!

JustinBenito:
Thanks a ton @JonnyBurger for a very quick response. Have fallen in love with remotion 🩵

JonnyBurger:
@JustinBenito Nice! 🥳❤️

PR #5233, opened 2025-05-02
docs: Updated setup.mdx

JustinBenito:
Swapped the 2 to maintain logical hierarchy since giving name is the 1st task in the page after clicking next and followed by ignoring the tags page.

AWS beginners like me would get confused.
because I created a policy with a different policy after reading the final instruction had to recreate a new policy with the given name remotion-lambda-policy.

QoL improvement


JonnyBurger:
Thanks a lot! 

I think AWS changed the UI a bit, appreciate you bringing this up to date!

PR #5248, opened 2025-05-06
`@remotion/studio`: pageup/down to switch compositions

UmungoBungo:
image

https://github.com/user-attachments/assets/a515c10d-c511-40b3-85ee-1478db4bfabf


packages/studio/src/components/CompositionSelector.tsx
```diff
@@ -17,6 +17,61 @@ import {
...
+		const previousIndex =
+			(currentIndex - 1 + compositions.length) % compositions.length;
+		const previousComposition = compositions[previousIndex];
+		selectComposition(previousComposition, true);
+	}, [canvasContent, compositions, selectComposition]);
+
+	return {
+		navigateToNextComposition,
+		navigateToPreviousComposition,
+	};
```
JonnyBurger: Can we memoize this as well? `return useMemo()`

JonnyBurger (changes requested):
Oh I love this! Just one nit

packages/studio/src/components/CompositionSelector.tsx
UmungoBungo: cool, have implemented :)

JonnyBurger (approved):

PR #5302, opened 2025-05-19
`@remotion/whisper-web`: fix package not working in Vite and upgrade `whisper.cpp` to 1.7.5

samohovets:
- [x] Upgrade whisper.cpp to 1.7.5
- [x] Re-compile with emscripten 4.0.8
- [x] Remove workaround to initialize FS with separate `worker.js` file
- [x] Remove workaround with global `moduleOverrides`, override directly on initialization using `createModule` constructor
- [x] ~Test with Vite to resolve https://github.com/emscripten-core/emscripten/issues/22394~


samohovets:
After a long investigation, I've decided to leave a note in the docs about Vite instead of patching the compiled `main.js`.

Simplified, the current problem looks like this:

- Webpack doesn't work with `new Worker(new URL(import.meta.url))` syntax. It was [fixed](https://github.com/webpack/webpack/pull/19279) recently; however, a lot of frameworks and setups are using the old webpack version, and we should require everyone to upgrade to use the package.

- Vite doesn't work correctly when spawning the worker like this: `new Worker(new URL('./main.js', import.meta.url)) ` (default `emscripten` way). It's a [known issue](https://github.com/vitejs/vite/issues/11672#issuecomment-1397855641).

So, it's just that Vite optimizer doesn't support this. For now, let's recommend using `optimizeDeps.exclude`. 

Hopefully, this will be fixed on the Vite side in the future.

Leaving the list of issues here for reference:

- https://github.com/webpack/webpack/issues/12638
- https://github.com/emscripten-core/emscripten/issues/22521
- https://github.com/emscripten-core/emscripten/issues/23957
- https://github.com/vitejs/vite/issues/8427
- https://github.com/vitejs/vite/issues/11672

JonnyBurger:
Thank youè

Issue #5307, opened 2025-05-20
Audio not playing in remotion player >= 3.0.2


dillionverma:
# Bug Report 🐛

All versions >= 3.0.2 causes audio to not work any longer in remotion player. Setting version to 3.0.1 immediately fixes the issue.

Tested on latest version of chrome and safari.

PR #5308, opened 2025-05-20
Update AudioForPreview.tsx so that crossOrigin is sent to useSharedAudio

juan-altatech:
https://github.com/remotion-dev/remotion/issues/5307

because we destructure props and crossOrigin is not part of nativeProps, it does not get passed to the audio tag generated in useSharedAudio https://github.com/remotion-dev/remotion/blob/main/packages/core/src/audio/AudioForPreview.tsx#L70


JonnyBurger:
Wow, good catch!

Unfortunately we have to disable the Web Audio API by default in the new version because it just doesn't work for every asset. So you'll have to opt in, but this is still useful, thank you!

PR #5315, opened 2025-05-23
Support more audio formats in `useWindowedAudioData`

### LINKED ISSUE #5314: `useWindowedAudioData`: support more audio formats
Now it only supports WAV 😿 

it'd be great to support more browser-supported audio formats!

Using `@remotion/media-parser`, we can achieve partial decoding on the fly as @JonnyBurger has shown in the [example snippet](https://discord.com/channels/809501355504959528/844143007183667220/1369218340622434354).

samohovets:
This closes #5314 

- [ ] Use `useWindowedAudioData` in music visualization template
- [ ] `AudioDecoder` queue will grow without any control and cause memory problems. Implement throttling for parsing as it's faster than decoding. 
- [ ] Test with really long audio files (1hr+)


JonnyBurger:
If you pick up work on this in the future, it should use this new API: https://www.remotion.dev/docs/webcodecs/get-partial-audio-data

samohovets:
@JonnyBurger thanks for putting that into webcodecs package! Will get back to this once Editor Starter is finished 👍

packages/media-utils/src/get-partial-audio-data.ts
```diff
@@ -0,0 +1,75 @@
...
+	if (signal.aborted) {
+		throw new Error('Operation was aborted');
+	}
+
+	const audioSamples: Float32Array[] = [];
+
+	// mediabunny docs: constructing the sink is virtually free and does not perform any media data reads.
+	const sink = new AudioBufferSink(track);
+
+	for await (const {buffer, timestamp, duration} of sink.buffers(
```
JonnyBurger, line 42: I predict a bug that for matroska based files, you will get a lot of zeroes at the beginning because the packets are dependant on the previous packets. see here what I mean: https://discord.com/channels/@me/1409810025844838481/1415028953093111870

we have introduced a mitigation for this by requesting 1.5sec of packets beforehand as well, and throwing them away
see an example fix for this: https://github.com/remotion-dev/remotion/blob/f319050080ba7ba14a00d63296412c11831a25a2/packages/media/src/audio-extraction/audio-iterator.ts#L13-L52

Can you also implement it for this?

(There is also an imprecision issue for Matroska, but I think we can disregard this for now because for visualization, this will not have such an effect)

samohovets: Fixed! Thank you 👍

PR #5348, opened 2025-06-03
Update jumpcuts.mdx

AlexHayton:
Without this, you can sometimes see a single-frame gap between jump cuts


AlexHayton:
Oops, our mistake.

PR #5360, opened 2025-06-09
Update getAbsoluteSrc to check for absoluteness of the URL too

### LINKED ISSUE #5359: getAbsoluteSrc is not respecting real absolute URLs
# Bug Report 🐛

## Summary

The current implementation of [getAbsoluteSrc](https://github.com/remotion-dev/remotion/blob/4af8e6238b9f471b56c4abef235f7bca0712c371/packages/core/src/absolute-src.ts) relies on the presence or absence of the `window` property to determine the absolute nature of a URL. It does **not** validate whether the given `src` is actually an absolute URL. This leads to issues when integrating Remotion into existing projects that already use a global `window` property in a conflicting way.

## Suggested Solution (non-breaking)

Update `getAbsoluteSrc` to validate whether the given `src` is a valid absolute URL (e.g., starts with `http://`, `https://`, etc.)—similar to how [staticFile](https://github.com/remotion-dev/remotion/blob/4af8e6238b9f471b56c4abef235f7bca0712c371/packages/core/src/static-file.ts#L100) handles this check.

If this sounds good, I’m happy to submit a PR with the fix.

## Steps to Reproduce

1. Define a global `window` property in your app (e.g., for SSR).
2. Use the `<Audio>` component with a fully qualified URL as `src` - as suggested within [docs](https://www.remotion.dev/docs/audio/importing).
3. Observe that Remotion misinterprets the URL and throws.

Error:
```
Error in generateVideoHandler: SymbolicateableError [TypeError]: Failed to construct 'URL': Invalid URL
    at getAbsoluteSrc (http://localhost:3001/bundle.js:19500:10)
    at AudioRefForwardingFunction (http://localhost:3001/bundle.js:21883:37)
    at renderWithHooks (http://localhost:3001/bundle.js:3692:21)
    at updateForwardRef (http://localhost:3001/bundle.js:6226:15)
    at beginWork (http://localhost:3001/bundle.js:7702:14)
    at performUnitOfWork (http://localhost:3001/bundle.js:11045:14)
    at workLoopSync (http://localhost:3001/bundle.js:10926:37)
    at renderRootSync (http://localhost:3001/bundle.js:10907:7)
    at performWorkOnRoot (http://localhost:3001/bundle.js:10512:40)
    at performWorkOnRootViaSchedulerTask (http://localhost:3001/bundle.js:11816:3)
```

fatemeh-khoshkam:
This PR fixes #5359 (getAbsoluteSrc is not respecting real absolute URLs).

Currently, getAbsoluteSrc is looking only at the `window` property to determine whether the given URL is absolute or not. This PR adds the check for the protocol too as described in the issue.

### DIFF
--- a/packages/core/src/absolute-src.ts
+++ b/packages/core/src/absolute-src.ts
@@ -3,5 +3,9 @@ export const getAbsoluteSrc = (relativeSrc: string) => {
 		return relativeSrc;
 	}
 
+	if (relativeSrc.startsWith('http://') || relativeSrc.startsWith('https://')) {
+		return relativeSrc;
+	}
+
 	return new URL(relativeSrc, window.origin).href;
 };
--- a/packages/core/src/absolute-src.ts
+++ b/packages/core/src/absolute-src.ts
@@ -3,7 +3,14 @@ export const getAbsoluteSrc = (relativeSrc: string) => {
 		return relativeSrc;
 	}
 
-	if (relativeSrc.startsWith('http://') || relativeSrc.startsWith('https://')) {
+	// https://github.com/remotion-dev/remotion/issues/5359
+	if (
+		relativeSrc.startsWith('http://') ||
+		relativeSrc.startsWith('https://') ||
+		relativeSrc.startsWith('file://') ||
+		relativeSrc.startsWith('blob:') ||
+		relativeSrc.startsWith('data:')
+	) {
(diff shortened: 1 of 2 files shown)


JonnyBurger:
Thanks a lot for fixing the issue! 🙌

PR #5379, opened 2025-06-15
Fix lambda-python type hints

webmonch:
## Overview
Fixed typing so **mypy** checks pass.

## Changes
- removed string enum redundant types
- replaced download_behavior field default_factory value with typed object instead of dict
- fixed payload object construction in construct_render_request 
- fixed function result hints that return optional objects
- removed redundant var assignments that didn't result in any output


JonnyBurger:
Thanks a lot!

PR #5420, opened 2025-06-18
Whisper Web Example

hunxjunedo:
image
image


JonnyBurger:
Thank you this is awesome!

Just one thing, if I am in light mode, the whole design breaks.
Can we maybe limit it to just dark mode?

image

JonnyBurger:
This is awesome!

/tip 200

PR #5473, opened 2025-06-26
@remotion/lambda: expose getLayers

atticoos:
Externally exposes the `getLayers` under `LambdaInternals`.

This does not advertise it as a public API, but does allow for an escape hatch to build a lambda manually with CDK (something we have been doing for a while).

Currently we copy the list of layers, but would like to use `getLayers` as a source of truth
image


JonnyBurger:
Good idea! To hide it in internals is also something I can be okay with :)

Issue #5459, opened 2025-06-23
Support a --concurrency flag on Lambda


JonnyBurger:
A common complaint is that --frames-per-lambda is cumbersome to use

- One does not know the duration of the video yet when invoking the command
- You just want to have a specific concurrency like 100

Goals:

- Support this new flag in renderMediaOnLambda() and npx remotion lambda render
- Should ignore if this would mean a framesPerLambda < 4
- If both framesPerLambda and concurrency are set, then throw
- Document it

JonnyBurger:
💎 This issue has a bounty on it!

Read our contributing guidelines: 
- **How to contribute**: https://www.remotion.dev/docs/contributing/
- **Implementing a new feature**: https://www.remotion.dev/docs/contributing/feature
- **Implementing a new option**: https://www.remotion.dev/docs/contributing/option
- **Write documentation**: https://www.remotion.dev/docs/contributing/docs
- **Bounty information**: https://www.remotion.dev/docs/contributing/bounty

/bounty 200

JonnyBurger:
## No more attempts

This issue has been assigned to @HiteshChoudhary0098

Do not send pull requests for this issue, do not ask to be in the queue, do not bother the assignee.  
**We will not accept any PRs not from the assignee and not award any bounties.**

Follow the #bounties channel on the Remotion Discord to get notified of new bounties.

PR #5474, opened 2025-06-26
`@remotion/lambda`: Add `--concurrency` flag as an alternative to `--frames-per-lambda`

HiteshChoudhary0098:
Issue :-  https://github.com/remotion-dev/remotion/issues/5459


JonnyBurger:
Is it already meant for review?

HiteshChoudhary0098:
Its ready now.

JonnyBurger:
Thanks a lot! I do need a few days to review this properly.
(I need to do civil service this week and my time is very limited)

By the weekend or Monday, I will have this reviewed!

packages/lambda-client/src/render-media-on-lambda.ts
```diff
@@ -51,6 +51,8 @@ export type RenderMediaOnLambdaInput = {
 	jpegQuality?: number;
 	maxRetries?: number;
 	framesPerLambda?: number;
+	concurrency?: number;
+	durationInFrames?: number;
```
JonnyBurger: This part must be wrong.

We do not know the duration yet! Therefore this should not be a parameter
We only know this after the Lambda launcher routine is invoked, and the calculation and validation needs to happen afterwards.

HiteshChoudhary0098: Will try to move it down the calling sequence. 😓

PR #5480, opened 2025-07-01
Change asset bucket from S3 to R2

MehmetAdemi:
The assets in the S3 bucket are causing high bandwidth. We should consider using Cloudflare R2 as an alternative.

I've updated all the links.


JonnyBurger (approved):
Nice thank you!

A few are left but I think the worst offenders we fixed

PR #5484, opened 2025-07-03
Change asset bucket from S3 to R2 (part 2)

MehmetAdemi:


packages/docs/learn/2022-12-22-apple-wow.mdx
```diff
@@ -556,7 +556,7 @@ export const MyComposition = () => {
 
 The animated explosion should look like this:
 
-<img src="https://remotion-assets.s3.eu-central-1.amazonaws.com/tutorial/Explosion.gif" />
+<img src=https://pub-646d808d9cb240cea53bedc76dd3cd0c.r2.dev/Explosion.gif" />
```
JonnyBurger: missing quote

MehmetAdemi: You are right, thanks!

PR #5502, opened 2025-07-14
`@remotion/install-whisper-cpp`: Add `flashAttention` and `additionalArgs` parameters to `transcribe()`

bkniffler:
I've added 2 possible arguments to transcribe:
- flashAttention (true/false)
- additionalArgs: array of strings

Flash Attention has been recently added and improves performance on mac:
- https://github.com/ggml-org/whisper.cpp/releases/tag/v1.7.6

Additional Args will be useful to support any args not (yet) exposed as options.

What do you think?

### DIFF
--- a/packages/docs/docs/install-whisper-cpp/transcribe.mdx
+++ b/packages/docs/docs/install-whisper-cpp/transcribe.mdx
@@ -137,7 +137,7 @@ Additional args to be passed to whisper, in an array. The array can contain stri
 ```js
 transcribe({
   ...,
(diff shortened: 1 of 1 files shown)


JonnyBurger:
Thanks! Can you also add these parameters to the docs? `packages/docs/docs/install-whisper-cpp/transcribe.mdx`

bkniffler:
Done @JonnyBurger

JonnyBurger:
Thanks a lot for this! 🙌

PR #5533, opened 2025-07-28
Lambda Python SDK: Update type hints

webmonch:
Fixed _input_props_ type from **list** to **Dict[str, Any]** 
Removed **Download** dataclass which seems duplicate of class **ShouldDownload** but without literal constraint. 
Replaced raw dict value of PlayInBrowser with object.


JonnyBurger:
Thanks a lot! 💙

PR #5559, opened 2025-08-05
fix(create-video): Add macOS GitHub Actions path to git-status test

scotthavird:
## 🐛 Bug Fix

Fixes the failing `create-video` test in GitHub Actions on macOS runners.
[mac-latest recent update to mac-15](https://github.blog/changelog/2025-07-11-upcoming-changes-to-macos-hosted-runners-macos-latest-migration-and-xcode-support-policy-updates/)

## 📋 Problem

The `git-status.test.ts` test was failing in GitHub Actions after the migration from `macos-13` to `macos-14` runners. The test checks if the git repository location matches expected paths, but was missing the explicit GitHub Actions macOS path.

**Error:**
```
error: expect(received).toEqual(expected)
Expected: true
Received: false
```

## 🔍 Root Cause

- GitHub Actions recently migrated from `macos-13` to `macos-14` runners (commit `f2465c25bc`)
- The test had hardcoded path conditions for Windows (`D:/a/remotion/remotion`) but not for macOS GitHub Actions
- The path resolution logic should work but apparently has subtle differences on `macos-14` runners

## 🛠️ Solution

Added the explicit GitHub Actions macOS path (`/Users/runner/work/remotion/remotion`) to the test conditions:

```typescript
status.location === 'D:/a/remotion/remotion' ||
status.location === '/Users/runner/work/remotion/remotion'  // ← Added this line
```

## ✅ Testing

- [x] Test passes locally with the fix
- [x] Verified this follows the same pattern as previous Windows path fix (commit `5625f951e7`)
- [x] Maintains backwards compatibility with existing path resolution logic

## 📚 Context

This test has a history of needing platform-specific path fixes:
- **Jan 2023**: Fixed for Windows with path normalization
- **May 2023**: Added Windows GitHub Actions path (`D:/a/remotion/remotion`)
- **Now**: Adding macOS GitHub Actions path

Follows the same approach as commit `5625f951e7` which added Windows GitHub Actions support.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -159,7 +159,7 @@ jobs:
             node_version: 16
           - os: windows-latest
             node_version: 16
-          - os: macos-14
+          - os: macos-latest
             node_version: 16
           - os: ubuntu-latest
             node_version: 20.5
--- a/packages/create-video/src/test/git-status.test.ts
+++ b/packages/create-video/src/test/git-status.test.ts
@@ -15,7 +15,8 @@ test('Get git status', async () => {
 			path.posix.join(__dirname, '..', '..', '..', '..').replace(/\\/g, '/') ||
 			status.location ===
 				path.join(__dirname, '..', '..', '..', '..').replace(/\\/g, '/') ||
-			status.location === 'D:/a/remotion/remotion',
+			status.location === 'D:/a/remotion/remotion' ||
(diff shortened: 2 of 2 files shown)


JonnyBurger:
Oh nice! 
I was wondering about this, thanks for sending this! 😊 All green on the first attempt! 🤩

PR #5572, opened 2025-08-08
Fix renderStill error when rendering negative frame index 

uen:
Docs say rendering frame -1 will render the last frame
https://www.remotion.dev/docs/renderer/render-still

If you pass a negative number it will error as there is a small bug in that logic

```
durationInFrames: 100

renderStill({frame: -1}) = frame 101
```


JonnyBurger:
Wow! Thanks for catching that 🙌

PR #5595, opened 2025-08-20
Fix media parser not accepting MPEG 2.5 mp3 file

samohovets:
Fixes `parseMedia` call with this mp3 file: [test-silence-enhanced.mp3](https://github.com/user-attachments/files/21903603/test-silence-enhanced.mp3)

Original issue in the Remotion Studio:

image


JonnyBurger:
LGTM!

PR #5608, opened 2025-08-26
`@remotion/media-parser`: Fix infinite loop in parseMedia for specific mp3

samohovets:
## What broke

For MP3 frames with Version ID 00 (MPEG 2.5), we computed an invalid frame length. This  made `isMp3PacketHeaderHereAndInNext()` fail its “next header” check and the syncword 
scan loop kept spinning when seeking (e.g. 90–95s).

## Spec alignment

- Version ID (B bits): 00 = MPEG 2.5 per the spec
- Sampling rate (F bits): MPEG 2.5 uses half of MPEG‑2’s rates: 11025/12000/8000 Hz. We now halve the MPEG‑2 value when audioVersionId === 0b00 ([source](MP3 frame header spec))
- Bitrate index (E bits): Table is keyed by Version (V1 vs V2) and Layer (L1/L2/L3). MPEG 2.5 shares the V2 table with MPEG‑2. Therefore, the “layer” must be used to select the L1/L2/L3 column; feeding the raw version bits as “level” was incorrect.

## Why this fixes the loop

Frame length for Layer III is proportional to bitrate_kbps / sampleRate (plus padding). With the correct V2/Lx bitrate and the halved 2.5 sample rate, `frameLength` is accurate. The “next header at current_offset + frameLength” check succeeds, so `discardUntilSyncword()` exits instead of spinning.

### DIFF
--- a/packages/media-parser/src/containers/mp3/seek/wait-until-syncword.ts
+++ b/packages/media-parser/src/containers/mp3/seek/wait-until-syncword.ts
@@ -12,6 +12,10 @@ export const discardUntilSyncword = ({
 			continue;
 		}
 
+		if (iterator.bytesRemaining() === 0) {
+			break;
+		}
+
 		const nextByte = iterator.getUint8();
 		const mask = 0xe0; // 1110 0000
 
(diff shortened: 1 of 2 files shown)


samohovets:
Related to #5595

PR #5616, opened 2025-08-29
Doc should install @remotion/eslint-plugin into devDependencies

kingh0730:
- Updated brownfield-installation.mdx


JonnyBurger:
Makes sense, thanks a lot!

PR #5639, opened 2025-09-07
Python sdk improvements

webmonch:
- added MyPy for static typing
- fixed types
- formatted with Black for consistency 
- changed RenderMediaProgress to dataclass to match other types


JonnyBurger:
Amazing! 

Thanks a lot, I tested it and it still works fine 🙌

PR #5654, opened 2025-09-13
Python sdk add more tests

webmonch:
- minor refactoring (common strings into constants, some type hints, readability of functions)
- added some tests for RemotionClient


JonnyBurger:
Thanks a lot! Tested it and still works, so LGTM 🙌

Issue #5708, opened 2025-10-02
Webhook callbacks not awaited, causing premature termination in serverless environments


andreisocaciu:
### Bug Description
The `appRouterWebhook` and `pagesRouterWebhook` helpers don't await async callbacks before returning responses. This causes async operations (database updates, API calls) to be terminated prematurely in serverless environments.

### Reproduction
```typescript
// This fails silently - database never updates
export const POST = appRouterWebhook({
  secret: 'mysecret',
  onSuccess: async ({ renderId, outputUrl }) => {
    // This database update never completes
    await db.update(jobs)
      .set({ status: 'completed', outputUrl })
      .where(eq(jobs.renderId, renderId));
  },
});
```

### Expected Behavior
Async operations in callbacks should complete before the response is returned.

### Actual Behavior
- Response returns immediately
- Serverless function terminates
- Async operations are killed mid-execution
- No errors logged (silent failure)

### Environment
- Next.js App Router / Pages Router
- Vercel / AWS Lambda / Any serverless platform
- @remotion/lambda version: 4.0.352+

### Root Cause
Lines 54-60 in `app-router-webhook.ts` and lines 46-52 in `pages-router-webhook.ts` call callbacks without `await`.

JonnyBurger:
Fixed in #5709

PR #5709, opened 2025-10-02
@remotion/lambda: Await async callbacks in appRouterWebhook to prevent premature termination

andreisocaciu:
## Problem
The appRouterWebhook helper doesn't await the onSuccess, onError, and onTimeout callbacks before returning the response. This causes issues in serverless environments (Vercel, AWS Lambda, etc.) where:
The callback starts executing async operations (e.g., database updates)
The response is immediately returned
The serverless function terminates
Async operations are killed mid-execution
This results in database updates, credit consumption, and other async operations silently failing without errors.

## Real-World Impact
Users performing database updates or API calls in webhook callbacks experience:
* Database records not updating despite successful renders
* No error logs (operations silently fail)
* Credits/billing not being processed

## Solution
Change callback signatures to support Promise<void> and await them before returning the response.

### DIFF
--- a/packages/docs/docs/lambda/webhooks.mdx
+++ b/packages/docs/docs/lambda/webhooks.mdx
@@ -139,7 +139,7 @@ Instead of validating the signature yourself, you can use the [`validateWebhookS
 
 ## Example webhook endpoint (Express)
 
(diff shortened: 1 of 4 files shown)


andreisocaciu:
Closes #5708

JonnyBurger:
Thanks a lot for sending a PR 🥳

PR #5724, opened 2025-10-05
Fix premount/postmount timeline indicators for `<Video>` and `<Audio>` in Remotion Studio

samohovets:
When a `<Video>` or `<Audio>` component is placed inside a `<Sequence>` with `premountFor` or `postmountFor` props, the Remotion Studio timeline only shows diagonal stripes for the Sequence itself, but not for the media elements inside it.

Before the fix:

before

After the fix:

after

Not sure if it was intended, however "after the fix" seems more correct to me! 
Since underlying component is also premounted.


samohovets:
Initially started investigating the need of `premountDisplay` and `postmountDisplay` in #5714  and found this! Seems like we should follow the same pattern in the new `@remotion/media` components.

JonnyBurger (approved):

PR #5751, opened 2025-10-08
`@remotion/media`: fallback preview to old `<Video>` or `<Audio>` in case of unsupported input format or CORS errors

samohovets:
- Upgrade mediabunny
- Fix CORS error handling in Studio
- Add fallback to old `<Video>`

### DIFF
--- a/packages/docs/docs/media/audio.mdx
+++ b/packages/docs/docs/media/audio.mdx
@@ -207,11 +207,6 @@ export const MyComposition = () => {
 };
 ```
 
-:::note
-Currently only works during rendering.  
-Planning to support it during preview as well.
-:::
-
 ### `toneFrequency?`
 
 Accepts a number between `0.01` and `2`, where `1` represents the original pitch. Values less than `1` will decrease the pitch, while values greater than `1` will increase it.
--- a/packages/media/src/audio/audio-for-preview.tsx
+++ b/packages/media/src/audio/audio-for-preview.tsx
@@ -139,6 +139,7 @@ const NewAudioForPreview: React.FC<NewAudioForPreviewProps> = ({
 					: undefined,
 				canvas: null,
 				playbackRate,
+				audioStreamIndex: audioStreamIndex ?? 0,
 			});
 
 			mediaPlayerRef.current = player;
@@ -259,6 +260,7 @@ const NewAudioForPreview: React.FC<NewAudioForPreviewProps> = ({
 		trimBefore,
 		playbackRate,
 		videoConfig.fps,
+		audioStreamIndex,
 		disallowFallbackToHtml5Audio,
 	]);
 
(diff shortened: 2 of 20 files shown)


JonnyBurger (approved):

packages/docs/docs/media/audio.mdx
```diff
@@ -241,6 +241,10 @@ Maps to [`<Audio />` -> `useWebAudioApi`](/docs/audio#usewebaudioapi)
 
 Maps to [`<Audio />` -> `acceptableTimeShiftInSeconds`](/docs/audio#acceptabletimeshiftinseconds)
 
+#### `pauseWhenBuffering?`
```
JonnyBurger, line 235: Already documented as a top-level prop!

packages/media/src/audio/audio-for-preview.tsx
```diff
@@ -30,6 +36,10 @@ type NewAudioForPreviewProps = {
 	readonly name: string | undefined;
 	readonly showInTimeline: boolean;
 	readonly stack: string | null;
+	readonly disallowFallbackToHtml5Audio: boolean;
+	readonly toneFrequency: number | undefined;
+	readonly audioStreamIndex: number | undefined;
+	readonly fallbackHtml5AudioProps?: FallbackHtml5AudioProps;
```
JonnyBurger: If a component already has only mandatory props, please avoid adding any optional props

This means that I intentionally made all props mandatory to maximize type safety

packages/docs/docs/media/audio.mdx
JonnyBurger: But actually, we said we don't expose it as a top-level prop. can you remove the documentation of the top level prop instead?

packages/studio/src/helpers/use-max-media-duration.ts
```diff
@@ -51,16 +51,12 @@ export const useMaxMediaDuration = (s: TSequence, fps: number) => {
 			}
 
 			// In case of CORS errors, fall back to getVideoMetadata
-			getVideoMetadata(src)
-				.then((metadata) => {
-					const durationOrInfinity = metadata.durationInSeconds ?? Infinity;
+			return getVideoMetadata(src).then((metadata) => {
```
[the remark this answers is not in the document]
JonnyBurger: Vercel is right, we should not remove the `.catch()`!

packages/media/src/audio/audio-for-preview.tsx
samohovets: Thanks, fixed!

packages/docs/docs/media/audio.mdx
samohovets: Yep, removed

PR #5787, opened 2025-10-11
Integrate LLM side panel chat using CrawlChat

MehmetAdemi:


MehmetAdemi:
After discussing with @JonnyBurger, we decided to close this PR as we are going to get rid of the navigation bar provided by Docusaurus and come up with our own implementation.

PR #5800, opened 2025-10-14
`@remotion/renderer`: Fix crash with odd dimensions in combination with scale

JonnyBurger:

### DIFF
--- a/packages/renderer/src/render-media.ts
+++ b/packages/renderer/src/render-media.ts
@@ -740,8 +740,8 @@ const internalRenderMediaRaw = ({
 
 				const stitchStart = Date.now();
 				return internalStitchFramesToVideo({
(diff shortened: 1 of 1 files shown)


packages/renderer/src/render-media.ts
```diff
@@ -740,8 +740,8 @@ const internalRenderMediaRaw = ({
 
 				const stitchStart = Date.now();
 				return internalStitchFramesToVideo({
-					width: actualWidth,
-					height: actualHeight,
```
saimfd, line 744: Will this also ensure that width and height are even numbers and not odd?

JonnyBurger: It is already done in a previous step!

PR #5867, opened 2025-10-25
Add AI video template

webmonch:
This PR adds an AI video template.

Comes with a small CLI that generates the timeline and content for the video. Video is rendered based on that timeline.

End result looks like this (the images could be improved using a better model):

https://github.com/user-attachments/assets/fdded84e-97c8-4706-a06e-6e8542026569

### DIFF
--- a/packages/template-prompt-to-video/cli/timeline.ts
+++ b/packages/template-prompt-to-video/cli/timeline.ts
@@ -136,22 +136,6 @@ export const getBgAnimations = (durationMs: number, zoomIn: boolean) => {
     endMs,
   });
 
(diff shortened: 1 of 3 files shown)


JonnyBurger:
Huge! 🥳

Thanks a lot for sending this in.
As it's large, needing a few days to process this. This is a great addition to our list of templates!

webmonch:
> Huge! 🥳
> 
> Thanks a lot for sending this in. As it's large, needing a few days to process this. This is a great addition to our list of templates!

Glad you like it! 
I may push a few minor commits later today to fix failing tests.

PR #5920, opened 2025-11-06
Pad leading silence for negative trim start on audio

dg314:
Rather than throw an error when `trimStartInSeconds` is negative, gracefully fallback to padding the audio with a leading silence. We can get rid of the `trimStartInSeconds > -0.0000000001` check and handling, as we get equivalent behavior from `fixFloatingPoint` and the `silenceFrames > 0` check.

This addresses fatal errors of the following form, which are occurring over 10x per hour on our production site (app.videogen.io):

```
trimStartInSeconds is negative: -0.02660416666666667. {"timeInSeconds":0,"ts":0.02660416666666667,"d":0.021333333333333333,"isFirstSample":true,"isLastSample":true,"durationInSeconds":0.03333333333333333,"i":0,"st":[0.02660416666666667]}
    at extractAudioInternal (https://remotionlambda-useast1-6078qr2mcj.s3.us-east-1.amazonaws.com/sites/videogen-prod/bundle.js:71501:15)
    at async Promise.all (index 1)
```

To confirm that the fix works, I created a patch of this change within our codebase, rebuilt and redeployed the bundle, and then confirmed that several previously affected videos now successfully export.

For example, any composition with the following clip would fail with that error on our production site:

https://github.com/user-attachments/assets/cbd1dc62-bf27-4be5-8e45-a5324a89edfe

### DIFF
--- a/packages/media/src/audio-extraction/extract-audio.ts
+++ b/packages/media/src/audio-extraction/extract-audio.ts
@@ -1,8 +1,15 @@
 import {type LogLevel} from 'remotion';
 import {audioManager} from '../caches';
 import {combineAudioDataAndClosePrevious} from '../convert-audiodata/combine-audiodata';
-import type {PcmS16AudioData} from '../convert-audiodata/convert-audiodata';
-import {convertAudioData} from '../convert-audiodata/convert-audiodata';
+import {
+	convertAudioData,
+	fixFloatingPoint,
+	type PcmS16AudioData,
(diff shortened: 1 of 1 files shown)


JonnyBurger:
Thanks for sending in a PR!
I tested and with #5923 and #5925 we already landed a fix.

BTW, if you enable "Allow contributors to change this branch" in GitHub, we can more easily merge your PR next time for glory and credits :)

PR #5928, opened 2025-11-10
Add custom Chrome browser option

JonnyBurger:
- Added new chromeVersion option to ensureBrowser() and related APIs
- Users can now specify a custom Chrome version directly via the chromeVersion option
- Updated OnBrowserDownload callback to receive chromeVersion parameter
- Added comprehensive documentation with examples and version sources
- Important: Users must delete node_modules/.remotion before changing versions

The chromeVersion option provides a simpler alternative to using onBrowserDownload for version control, while onBrowserDownload can still override it if needed.


JonnyBurger:
Oops, already possible: https://www.remotion.dev/docs/renderer/ensure-browser#onbrowserdownload

PR #5931, opened 2025-11-10
`@remotion/media`: fix choppy audio when unpausing playback

### LINKED ISSUE #5926: `@remotion/media`: audio chunks are played repeatedly after unpausing playback
# Bug Report 🐛

When unpausing playback, I hear at least 3 audio chunks that are played in the wrong order.

https://github.com/user-attachments/assets/24fe7664-a100-4502-a795-b77bb8ef2c06

Probably related #5872.

samohovets:
There was a drift correction mechanism in `seekTo()` that would reset the audioSyncAnchor when detecting >0.04s difference. This was causing a race 
condition when unpausing:

1. `play()` sets audioSyncAnchor for the current time
2. `seekTo()` gets called as frames advance (1-2 frames = 0.04-0.08s later)
3. Drift detection triggers and resets audioSyncAnchor again
4. This disrupts the already-scheduled audio chunks, causing glitches

The fix: Only apply drift correction when paused. During playback, the audioSyncAnchor set by `play()` should remain stable. The drift correction is still needed for scrubbing while paused, but not for normal frame 
advancement during playback

This fixes #5926

### DIFF
--- a/packages/media/src/media-player.ts
+++ b/packages/media/src/media-player.ts
@@ -339,19 +339,6 @@ export class MediaPlayer {
 			return;
 		}
 
-		if (!this.playing) {
-			const newAudioSyncAnchor =
-				this.sharedAudioContext.currentTime -
-				newTime / (this.playbackRate * this.globalPlaybackRate);
-			const diff = Math.abs(newAudioSyncAnchor - this.audioSyncAnchor);
-			if (diff > 0.04) {
-				this.setPlaybackTime(
-					newTime,
-					this.playbackRate * this.globalPlaybackRate,
-				);
-			}
-		}
-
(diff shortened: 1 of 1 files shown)


samohovets:
Now I even removed this check after some testing, we don't need it because:

1. Every time user is seeking during playback, the chain is:`pause()` => `seekTo() with is this.playing = false` => `play()` => `seekTo() with this.playing = true`
2. Every time `play()` gets called, audio anchor is being reset
3. `play()` is always called before any `seekTo()` where `this.playing = true` is expected
4. When video is paused, we don't care about audio playback (scheduling)

Seems like a too defensive check

JonnyBurger:
I tested it and could not find any new issues after removing this!

SGTM! 🤣

PR #5969, opened 2025-11-18
`@remotion/media`: Perfect loop for new video tag

### LINKED ISSUE #5918: `@remotion/media`: In preview, buffer does not anticipate a loop
The audio and video iterators should anticipate the videos jumping to the beginning of the loop to make looping seamless.

We know when the video starts using `trimBefore`, and we should prepare an iterator when the current time nears the end of the Sink, or the end duration of the video.

samohovets:
Fixes #5918


samohovets:
There are still perfect loop problems if video has b-frames. 

Mediabunny fetches timestamp `0.6s` when requesting time `0s` :( 

This should be fixable I think

PR #5985, opened 2025-11-24
`@remotion/lambda`: Automatically switch to `gl=swangle` if `gl=angle` is passed

### LINKED ISSUE #5982: Three.js template has `gl=angle` by default, but Lambda does not support it
Which leads to a painful experience. 
It would make sense to switch to `--gl=swangle` automatically on Lambda, if you specify `--gl=angle` (even if it is by template default)

https://gist.github.com/JonnyBurger/14f320e2d2102cfb81155ea5d26b9ff4

niladrix719:
Fixes #5982


JonnyBurger:
This wouldn't work because we had validation before - it wouldn't get past this.  
I changed it to now switch the renderer in the correct place!

Issue #5987, opened 2025-11-26
Support `crossOrigin` for `@remotion/media` <Audio> & <Video>


hnbnh:
# Feature Request 🛍️

- Support the `crossOrigin` attribute the same way as `<Html5Video>` and `<Html5Audio>`.

PR #5989, opened 2025-11-26
feat(media): add crossOrigin prop to `@remotion/media` Audio & Video components

hnbnh:
- Closes: https://github.com/remotion-dev/remotion/issues/5987

### DIFF
--- a/packages/docs/docs/media/audio.mdx
+++ b/packages/docs/docs/media/audio.mdx
@@ -172,7 +172,7 @@ export const MyComposition = () => {
 
 ### `crossOrigin?`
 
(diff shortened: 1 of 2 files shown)


JonnyBurger (changes requested):
Thanks for submitting a PR!

With using the new media tags, the URL will always be `fetch()`-ed and played with a custom implementation.

There is no underlying `<video>` or `<audio>` tag. Therefore the documentation and implementation in its current state is wrong.

Assets must be CORS-enabled to make use of the new `@remotion/media` tags.  
If they are not, they fall back to the HTML5 tags: https://www.remotion.dev/docs/media/fallback#when-a-fallback-is-attempted

The only way this PR can make sense if we add support for the `crossOrigin` prop in case of a fallback. This prop should then be added under https://www.remotion.dev/docs/media/video#fallbackoffthreadvideoprops and https://www.remotion.dev/docs/media/audio#fallbackhtml5audioprops .

JonnyBurger:
Sorry, but my comments were not addressed properly and `crossOrigin` is still a top level prop and it comes with a lot of changes which don't make sense :(

I am closing this in favor of #6005

PR #6016, opened 2025-12-03
Update performance.mdx

dherault:
I saw the typo, I fixed the typo.


JonnyBurger:
That's the way to go! 🙌

Thanks for that, as you can see I did not actually die mid sentence 🫠

PR #6035, opened 2025-12-08
`@remotion/lambda`: Handle `concurrency` being undefined

JonnyBurger:
Only merge if https://discord.com/channels/809501355504959528/844143007183667220/1447661945003184178 turns out to not be true


JonnyBurger:
I got no response on Discord, so I have to assume user just didn't update their Lambda SDK. Closing

PR #6075, opened 2025-12-13
`@remotion/media`: Fix rapid block/unblock cycles during video seek

samohovets:
## Summary

- Fix spinner flickering caused by overlapping block/unblock cycles when seeking video 
- Move playback blocking logic from `startVideoIterator` to the `seek` level, ensuring only one block/unblock 
cycle per seek operation

## Problem

When seeking through video, if `tryToSatisfySeek` returned `'not-satisfied'`, `startVideoIterator` was called fire-and-forget with its own internal block/unblock. Multiple rapid seeks caused overlapping block/unblock cycles, resulting in spinner flickering.

## Solution

- Split `startVideoIterator` into `startVideoIteratorWithoutBlocking` (core logic) and `startVideoIterator` (with blocking wrapper)
- `seek` now manages blocking at its level - only blocks when a new iterator is actually needed
- No blocking when frame is already available (`'satisfied'` path)
## Testing

Added `video-iterator-manager.test.ts` with tests verifying `maxConcurrentBlocks === 1` during rapid seeking.

### DIFF
--- a/packages/media/src/test/video-iterator-manager.test.ts
+++ b/packages/media/src/test/video-iterator-manager.test.ts
@@ -1,0 +1,102 @@
+import {ALL_FORMATS, Input, UrlSource} from 'mediabunny';
+import {expect, test} from 'vitest';
+import {makeNonceManager} from '../nonce-manager';
+import {videoIteratorManager} from '../video-iterator-manager';
+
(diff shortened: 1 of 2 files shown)


samohovets:
Closing in favor of #6078

PR #6076, opened 2025-12-14
`remotion`: Fix audio gaps in web renderer caused by stale React closure

samohovets:
Fixes audio gaps that occurred during client-side rendering with `@remotion/web-renderer`, particularly noticeable in React 18 and with expensive frames.

**The problem:** `collectAssets()` in `RenderAssetManager` used a `useImperativeHandle` closure that captured `renderAssets` state. When `registerRenderAsset()` updated the state, React would queue a re-render, but `collectAssets()` could be called before React updated the closure—returning an empty array and causing audio gaps.

**The fix:** Use a `useRef` alongside state for synchronous access. The ref is updated immediately when assets are registered, and `collectAssets()` reads from the ref instead of relying on the closure.

### DIFF
--- a/packages/core/src/RenderAssetManager.tsx
+++ b/packages/core/src/RenderAssetManager.tsx
@@ -4,9 +4,9 @@ import {
 	useImperativeHandle,
 	useLayoutEffect,
 	useMemo,
+	useRef,
 	useState,
 } from 'react';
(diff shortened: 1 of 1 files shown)


JonnyBurger:
I independently verified that this is the correct fix! 😊
It even saves a forced React state update, nice job!!

PR #6097, opened 2025-12-16
`@remotion/web-renderer`: Add telemetry using `@remotion/licensing`

pskd73:

### DIFF
--- a/packages/docs/docs/licensing/index.mdx
+++ b/packages/docs/docs/licensing/index.mdx
@@ -7,30 +7,35 @@ crumb: API
 
 _available from v4.0.237_
 
-This package allows holders of the [Company License](https://remotion.pro/license) to send events to Remotion to track the usage of cloud renders.
+This package allows holders of the [Company License](https://remotion.pro/license) to send events to Remotion to track the usage of renders.  
 Also, it offers an API to programmatically check the usage in order to implement spend controls.
 
 ### How do I use this package?
 
-On your company license dashboard on [remotion.pro](https://remotion.pro), you can find your public and secret API key.  
-You can use the public API key to register usage points and the secret API key to get the current usage.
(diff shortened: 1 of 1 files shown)


packages/web-renderer/src/render-still-on-web.tsx
```diff
@@ -134,6 +137,11 @@ async function internalRenderStillOnWeb<
 			await artifactsHandler.handle({imageData, frame, assets, onArtifact});
 		}
 
+		await sendUsageEvent({
```
JonnyBurger: This can be async, user result should not be blocked by it

packages/web-renderer/src/send-telemetry-event.ts
```diff
@@ -0,0 +1,34 @@
...
+	if (licenseKey === null) {
+		Internals.Log.warn(
+			{logLevel: 'warn', tag: 'web-renderer'},
+			'You need to provide a license key to use the web renderer going forward.',
+		);
+	}
+
+	await registerUsageEvent({
+		apiKey: licenseKey,
+		event: 'cloud-render',
```
JonnyBurger: Call it "webcodecs-conversion" (this is what's meant by client side render, we will in the backend soon allow an alias to be passed)

packages/web-renderer/src/render-still-on-web.tsx
```diff
@@ -135,7 +138,21 @@ async function internalRenderStillOnWeb<
...
 		return imageData;
+	} catch (err) {
+		sendUsageEvent({succeeded: false, licenseKey}).catch((err2) => {
+			Internals.Log.error(
+				{logLevel: 'error', tag: 'web-renderer'},
+				'Failed to send usage event',
+				err2,
+			);
+		});
+		throw err;
```
pskd73, line 160: I hope this is fine

PR #6157, opened 2025-12-29
`@remotion/studio`: Unified render button

JonnyBurger:
TODO: 

- [x] Better labels
- [x] Disable if command line flag is not specified
- [x] Clicking one should open the modal
- [x] In bundled mode, show the "Render on Web" button


samohovets:
@JonnyBurger there's also a bug that can be fixed in this PR: 

if the video was rendered using web renderer, the progress does not reset, it's stuck in "Rendering frame 301/300".

JonnyBurger:
@samohovets I will tackle this separetely! Goal is to add the render to the sidebar like server side renders

PR #6161, opened 2025-12-29
`@remotion/media-utils`: Optimize audio window loading priority and revert connection closing

samohovets:
- Prioritize loading the current playback window first in useWindowedAudioData(), then load surrounding windows. This improves performance on slow connections by ensuring the most critical audio data loads first.
- Revert PR #6121's connection closing changes since the UrlSource caching wasn't being utilized when creating new Input instances per request. Keeping the Input and track in state allows proper connection reuse across window fetches.

### DIFF
--- a/packages/media-utils/src/use-windowed-audio-data.ts
+++ b/packages/media-utils/src/use-windowed-audio-data.ts
@@ -317,9 +317,8 @@ export const useWindowedAudioData = ({
 
 		// Prioritize the current window where playback is at.
 		// On slow connections, this ensures the most important window loads first.
-		const currentWindowNeedsFetch = windowsToActuallyFetch.includes(
-			currentWindowIndex,
-		);
+		const currentWindowNeedsFetch =
+			windowsToActuallyFetch.includes(currentWindowIndex);
 		const otherWindowsToFetch = windowsToActuallyFetch.filter(
 			(w) => w !== currentWindowIndex,
 		);


packages/media-utils/src/get-partial-audio-data.ts
```diff
@@ -1,26 +1,27 @@
...
 
 export type GetPartialAudioDataProps = {
+	track: InputAudioTrack;
 	fromSeconds: number;
 	toSeconds: number;
 	channelIndex: number;
 	signal: AbortSignal;
-	src: string;
-	isMatroska: boolean;
+	isMatroska?: boolean;
```
JonnyBurger: Why are we making mandatory props optional?
This increases the chance for us to make mistakes

packages/media-utils/src/use-windowed-audio-data.ts
```diff
@@ -96,11 +103,9 @@ export const useWindowedAudioData = ({
 
 			signal.addEventListener('abort', cont, {once: true});
 
-			const source = new UrlSource(src);
-
-			using input = new Input({
+			const input = new Input({
```
JonnyBurger, line 106: Now we are not disposing the input anymore, it is being kept indefinitely.

While this makes sense in preview, won't this make it more likely for us to hit the Chrome connection limit during render?

We had this problem in the retreat, won't we just reintroduce it again?

Maybe it makes sense to differentiate between renders and preview as well!

packages/media-utils/src/get-partial-audio-data.ts
samohovets: Good catch! Thanks.

packages/media-utils/src/use-windowed-audio-data.ts
samohovets: I can confirm we're not re-introducing the same problem! 

The fix we introduced in the retreat:
- masked the main problem: Mediabunny's `UrlSource` was over-fetching + `input.dispose()` was not working, then seeking spawned new connections while initially opened ones were still in-flight.
- introduced a new problem: no cache from `UrlSource` was leveraged. By default it caches 64MB of data, which is pretty beneficial in seeking scenarios over a long file.

It's not incorrect to store `Input` since now all requests are correctly being aborted + no over-fetching!

PR #6170, opened 2025-12-31
`@remotion/web-renderer`: Add audioCodec and audioBitrate options

### LINKED ISSUE #6047: Web Renderer: Add option to specify the audio codec
- Add a new option `audioCodec` to `renderMediaOnWeb`, it can work similarly as `codec`
- Add the option to `packages/studio/src/components/RenderModal/WebRenderModal.tsx` in the `Audio` settings group. If we have not yet added that, introduce it (ServerRenderModal has it already)
- This one has a special consideration: Firefox does not have support for AAC encoding, so Opus will be used by default. Therefore, if the user customizes the audio codec, they need to think about this too. Make sure the user is aware of this.
- Check using Mediabunny if the specified audio codec can be put in the container. You can use [`outputContainer.getSupportedAudioCodec()`](https://github.com/remotion-dev/remotion/blob/c5e64946c9c458e7057bbd417b642a3d1ca98154/packages/convert/app/lib/can-transcode-or-copy.ts#L27)
- Check if the audio codec can be encoded by the browser using [`getEncodableAudioCodecs()`](https://github.com/remotion-dev/remotion/blob/c5e64946c9c458e7057bbd417b642a3d1ca98154/packages/convert/app/lib/can-transcode-or-copy.ts#L32)
- Don't show impossible options in `WebRenderModal.tsx`, either if audio codec is not encodable or cannot be put in the container. Or show it and handle it nicely by graying out the option or validating it otherwise

samohovets:
Fixes #6047 

- Add `audioCodec` option to `renderMediaOnWeb()` with support for `'aac'` and `'opus'` codecs
- Add `audioBitrate` option to control audio quality (`'very-low'` | `'low'` | `'medium'` | `'high'` | `'very-high'`)
- Add Audio Codec and Audio Quality selectors to the Studio's Web Render Modal
- Automatically detect browser-encodable codecs and disable options accordingly

### DIFF
--- a/packages/studio/src/components/RenderModal/WebRenderModal.tsx
+++ b/packages/studio/src/components/RenderModal/WebRenderModal.tsx
@@ -47,14 +47,14 @@ import {
 	ResolveCompositionBeforeModal,
 	ResolvedCompositionContext,
 } from './ResolveCompositionBeforeModal';
+import {useEncodableAudioCodecs} from './use-encodable-audio-codecs';
+import {useEncodableVideoCodecs} from './use-encodable-video-codecs';
 import {WebRendererExperimentalBadge} from './WebRendererExperimentalBadge';
 import {WebRenderModalAdvanced} from './WebRenderModalAdvanced';
 import {WebRenderModalAudio} from './WebRenderModalAudio';
 import {WebRenderModalBasic} from './WebRenderModalBasic';
 import {WebRenderModalLicense} from './WebRenderModalLicense';
 import {WebRenderModalPicture} from './WebRenderModalPicture';
-import {useEncodableAudioCodecs} from './use-encodable-audio-codecs';
-import {useEncodableVideoCodecs} from './use-encodable-video-codecs';
 
 type WebRenderModalProps = {
 	readonly compositionId: string;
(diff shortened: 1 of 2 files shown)


packages/studio/src/components/RenderModal/WebRenderModalAudio.tsx
```diff
@@ -1,23 +1,150 @@
...
+			const filtered = supported.filter((c) => encodable.includes(c));
+			setEncodableCodecs(filtered);
+		});
+	}, [videoContainer]);
+
+	useEffect(() => {
+		if (!encodableCodecs.includes(audioCodec) && encodableCodecs.length > 0) {
+			setAudioCodec(encodableCodecs[0]);
+		}
+	}, [encodableCodecs, audioCodec, setAudioCodec]);
```
JonnyBurger: 1. I agree with Copilot comment, this seems sketchy way to do it. Better would be to only store the user preference, and use a derived state to determine which codec is actually selected.
2. The useEffect() is anyway not working if the audio tab is not selected, e.g. if the user goes to main tab and changes video container this does not work as you intend to!
3. The fetching can be put in the parent component, then it gets prefetched for the audio modal and no "flicker" will be visible

samohovets: Thanks! Fixed 👍

JonnyBurger (dismissed):
Quality! Thanks a lot 🙌

PR #6189, opened 2026-01-01
docs(lambda/cli/functions/deploy): fix runtime preference parameter value to correct one

peterslany:
Fixing the docs for CLI command to deploy function to lambda `remotion lambda functions deploy --runtime-preference`. On the current version of the CLI (4.0.396), the `prefer-` prefix in the parameter value is no longer used and therefore the docs are incorrect and the command throws error. By deleting the `prefer-` prefix this issue gets resolved.

Adding error that led me to this bug for full context
```
An error occurred:
Error: Invalid runtime preference prefer-apple-emojis. Must be one of default, apple-emojis, cjk
    at validateRuntimePreference...
```

### DIFF
--- a/packages/docs/docs/lambda/deployfunction.mdx
+++ b/packages/docs/docs/lambda/deployfunction.mdx
@@ -71,7 +71,7 @@ Use a custom role for the function instead of the default (`arn:aws:iam::[aws-ac
 
 One of:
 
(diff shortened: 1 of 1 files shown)


JonnyBurger:
Thanks a lot!

PR #6198, opened 2026-01-02
`@remotion/web-renderer`: Support `background-clip` CSS property

JonnyBurger:


JonnyBurger:
@claude one test is failing but it is okay, just increase the tolerance to 0.03

PR #6200, opened 2026-01-02
fix space

### LINKED ISSUE #6188: Studio: Adding Tailwind removes the space between project name and "Open in"dialogue
Image

The space between these 2 elements is not there as it should be

```
cd packages/template-next-app-tailwind
npm run dev
```

hunxjunedo:
closes #6188


JonnyBurger (changes requested):
this is different from making the space work

- this changes the look of the focus outline
- this will duplicate the space when a template does not use tailwind

hunxjunedo:
@JonnyBurger this is not the case, I've tested this with both templates and the space is completely identical

JonnyBurger:
You are right! Sorry about this

PR #6203, opened 2026-01-03
`@remotion/licensing`: No unhandled rejection if usage event send fails

JonnyBurger:
Only a log message


JonnyBurger:
@claude fix this:

The response body can only be read once. Line 76 already calls res.json() to parse the response. Lines 89-90 attempt to call res.json() again, which will fail because the body stream has already been consumed. This code appears to be unreachable since all success paths return at line 79-82, and all error paths throw at line 86. Consider removing lines 89-90.

JonnyBurger:
@claude now we return nothing. throw an error instead

PR #6219, opened 2026-01-03
Update `WaitForFonts` component to resolve text issue

JonnyBurger:


JonnyBurger:
@claude oops in the documentation i forgot to import useDelayRender() instead of delayRender.

Plz fix the imports

PR #6240, opened 2026-01-05
`@remotion/web-renderer`: Proper cleanup of transforms using "using"

JonnyBurger:


samohovets:
Referencing https://github.com/Vanilagy/mediabunny/issues/270 here so `using` just works, without explicit `.close()` on media sources!

PR #6247, opened 2026-01-05
[DRAFT] `@remotion/web-renderer`: Add utility to check if render can be performed

### LINKED ISSUE #6103: Web renderer: add an utility for checking if render can be performed
> the highest level of DX here would be to have enough data to explain to the end user what they could do to render the video
> 
> switch browser? install necesarry codecs? something like this
> 
> i mean, not Remotion literally returning these messages, but have enough data to decide on your own as a dev how to communicate that to the user

- [ ] Would be nice if it was a single method
- [ ] We do not fail fast in `renderMediaOnWeb`, so it seems like this method is going to be re-used internally (we assume the overhead is low)
- [x] If possible, the API should be as close as possible to `renderMediaOnWeb`
- [ ] Can be reused in Studio

samohovets:
Add canRenderVideoCodec() and canRenderAudioCodec() functions to check if the browser supports encoding with specific codecs before attempting to render. These functions wrap mediabunny's canEncodeVideo and canEncodeAudio APIs with Remotion's type mappings.

Closes #6103

### DIFF
--- a/packages/web-renderer/src/can-render-video-codec.ts
+++ b/packages/web-renderer/src/can-render-video-codec.ts
@@ -1,5 +1,8 @@
 import {canEncodeVideo} from 'mediabunny';
-import type {WebRendererQuality, WebRendererVideoCodec} from './mediabunny-mappings';
+import type {
+	WebRendererQuality,
+	WebRendererVideoCodec,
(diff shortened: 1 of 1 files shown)


samohovets:
@claude fix monorepo build

PR #6259, opened 2026-01-07
`@remotion/web-renderer`: new APIs: `canRenderMediaOnWeb()`, `getEncodableAudioCodecs()`, `getEncodableVideoCodecs()`

### LINKED ISSUE #6103: Web renderer: add an utility for checking if render can be performed

samohovets:
Fixes #6103


packages/docs/docs/web-renderer/can-render-media-on-web.mdx
```diff
@@ -0,0 +1,165 @@
...
+
+### `canRender`
+
+_boolean_
+
+`true` if the render can proceed, `false` if there are blocking issues.
+
+### `issues`
+
+_CanRenderIssue[]_
```
JonnyBurger: Can we use TsType like elsewhere?

packages/docs/docs/web-renderer/can-render-media-on-web.mdx
```diff
@@ -0,0 +1,165 @@
...
+
+The video codec that will be used for rendering.
+
+### `resolvedAudioCodec`
+
+_string | null_ <TsType type="WebRendererAudioCodec" source="@remotion/web-renderer" />
+
+The audio codec that will be used for rendering. This may differ from the requested codec if a fallback was applied. `null` if `muted` is `true`.
+
+### `resolvedOutputTarget`
```
JonnyBurger, line 153: That's weird, we don't allow this as input, but it appears as output

packages/docs/docs/web-renderer/get-encodable-audio-codecs.mdx
```diff
@@ -0,0 +1,56 @@
...
+
+The container format: `"mp4"` or `"webm"`.
+
+### `options?`
+
+_object_ <TsType type="GetEncodableCodecsOptions" source="@remotion/web-renderer" />
+
+Optional configuration:
+
+- `bitrate?`: A number (bits per second) or quality preset (`"very-low"`, `"low"`, `"medium"`, `"high"`, `"very-high"`)
```
JonnyBurger: Usually we make all fields headings, so this should be a h4, `####`

packages/docs/docs/web-renderer/get-encodable-video-codecs.mdx
```diff
@@ -0,0 +1,52 @@
...
+
+The container format: `"mp4"` or `"webm"`.
+
+### `options?`
+
+_object_ <TsType type="GetEncodableCodecsOptions" source="@remotion/web-renderer" />
+
+Optional configuration:
+
+- `bitrate?`: A number (bits per second) or quality preset (`"very-low"`, `"low"`, `"medium"`, `"high"`, `"very-high"`)
```
JonnyBurger: Same as above, make this a h4 `####`

JonnyBurger: Just so people don't accidentally confuse themselves - can we call it `videoBitrate` instead of `bitrate`? Same for audio

samohovets: Thanks, fixed!

packages/docs/docs/web-renderer/can-render-media-on-web.mdx
samohovets: Great catch! Fixed, we have a fallback to `arraybuffer` in case `web-fs` is not available

PR #6273, opened 2026-01-09
`@remotion/licensing`: add retry with exponential backoff for `registerUsageEvent`

samohovets:


samohovets:
@claude can you add tests for retry behavior? Use https://mswjs.io/

PR #6280, opened 2026-01-12
`@remotion/web-renderer`: fix `onProgress` being called after render is finished

samohovets:
Related: fixes an issue in studio when renderer finishes rendering but onProgress is still reporting values

### DIFF
--- a/packages/web-renderer/src/test/render-media.test.tsx
+++ b/packages/web-renderer/src/test/render-media.test.tsx
@@ -99,7 +99,12 @@ test('should throttle onProgress callback to 250ms', {retry: 2}, async (t) => {
 	}
 });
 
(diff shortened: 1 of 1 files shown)


packages/web-renderer/src/throttle-progress.ts
```diff
@@ -2,14 +2,20 @@ import type {RenderMediaOnWebProgressCallback} from './render-media-on-web';
 
 const DEFAULT_THROTTLE_MS = 250;
 
+export type ThrottledProgressResult = {
+	throttled: RenderMediaOnWebProgressCallback;
+	cleanup: () => void;
```
JonnyBurger: `cleanup` is never used in the return value

samohovets: Fixed, thanks!

PR #6281, opened 2026-01-12
`@remotion/web-renderer`: Add `scale` option

clearlysid:
Similar to the server-side renderer, I added an (optional) `scale` prop to `renderMediaOnWeb` and `renderStillOnWeb`.

Notes:
- needed to pass the `scale` around and use it to change the transform matrix as well as position of the elements
- avoid apply the scale to precompose or clipping process (these don't benefit from scaling)
- added a new fixture and a test to check if scaling works on basic elements correctly
- added screenshots to demonstrate passing tests
- commented some decisions inline, can remove before merging if needed

I'm worried that having to pass the scale value around like this is not the best approach from a code quality standpoint. Maybe someone can suggest a better approach, though for now it seems to work okay.

### DIFF
--- a/packages/web-renderer/src/test/background-clip.test.tsx
+++ b/packages/web-renderer/src/test/background-clip.test.tsx
@@ -37,6 +37,6 @@ test('should render background-clip: text (scaled)', async () => {
 		blob,
 		testId: 'background-clip-text-scaled',
 		threshold: 0,
-		allowedMismatchedPixelRatio: 0.03,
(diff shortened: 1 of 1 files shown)


JonnyBurger:
WOW! That's brilliant, I thought it would not be possible for client-side rendering, but this is a great approach! 🙌

Nice surprise, will do some checks tomorrow and merge then!

JonnyBurger:
Yay!

PR #6286, opened 2026-01-14
About page: Update team picture

JonnyBurger:
- Replaced team.jpeg in both promo-pages and docs
- Cropped from 2998x2000 to 2600x1800 to better center the team
- Optimized from 1.1MB to 488KB (56% reduction)
- Improved composition with team members more prominent


JonnyBurger:
@claude 

1. I think we can make the team picture even smaller, 1500px wide at most
2. packages/promo-pages/src/components/TeamPicture.tsx should have the aspect ratio updated

PR #6288, opened 2026-01-14
`@remotion/web-renderer`: properly propagate errors during render

### LINKED ISSUE #6287: Web renderer: fix error propagation
The `@remotion/web-renderer` package has a bug where errors thrown during React component rendering (e.g., user code errors, delayRender timeouts) are not properly propagated to the caller. 

This blocks #6279

samohovets:
Fixes #6287

### DIFF
--- a/packages/web-renderer/src/test/error-propagation.test.tsx
+++ b/packages/web-renderer/src/test/error-propagation.test.tsx
@@ -1,5 +1,5 @@
-import {useEffect, useState} from 'react';
-import {useCurrentFrame, useDelayRender} from 'remotion';
+import {useEffect} from 'react';
+import {useCurrentFrame} from 'remotion';
 import {expect, test} from 'vitest';
 import {renderMediaOnWeb} from '../render-media-on-web';
 import {renderStillOnWeb} from '../render-still-on-web';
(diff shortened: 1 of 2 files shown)


packages/web-renderer/src/test/error-propagation.test.tsx
```diff
@@ -0,0 +1,168 @@
...
+				id: 'never-continues',
+				width: 100,
+				height: 100,
+				fps: 30,
+				durationInFrames: 10,
+			},
+			inputProps: {},
+			delayRenderTimeoutInMilliseconds: 3000,
+		}),
+	).rejects.toThrow(/delayRender.*Waiting forever/);
```
JonnyBurger: This could slow down the test suite significantly and from previous experiences timing based tests are also shaky.

We should consider just deleting this test..

packages/web-renderer/src/test/error-propagation.test.tsx
```diff
@@ -0,0 +1,168 @@
...
+class UseEffectError extends Error {}
+class FrameRenderError extends Error {}
+class UseEffectOnFrameError extends Error {}
+class InitialRenderError extends Error {}
+
+test('should propagate errors thrown in useEffect', async (t) => {
+	if (t.task.file.projectName === 'webkit') {
+		t.skip();
+		return;
+	}
```
JonnyBurger: Did we try to see if the test would also run fine in Safari?

samohovets: Weird, it was not working in the middle of my work.. now it works! Removed that.

packages/web-renderer/src/test/error-propagation.test.tsx
samohovets: Removed!

Issue #6176, opened 2025-12-31
Recorder: Cmd+I conflicts with new Ask AI modal


JonnyBurger:
It would be good to have a flag in the Studio to disable the "Ask AI" modal since it conflicts in certain Studios.

Something like `.setEnableAskAiFeature()`.

Here is a reference pull request for how a new flag in the Studio was implemented. 
https://github.com/remotion-dev/remotion/pull/5304/changes/ce82582e584625347e8265c5980d3d966b9f90c9

hunxjunedo:
Please assign

StanByriukov02:
/attempt #6176
Implement Config flag to disable Ask AI, removing Cmd/Ctrl+I binding + menu entry to avoid recorder conflict.

hunxjunedo:
@StanByriukov02 please do not attempt this, I am already working towards a solution.

PR #6296, opened 2026-01-15
@remotion/studio: Add Config flag to disable Ask AI 

StanByriukov02:
/claim #6176

Adds Config.setEnableAskAiFeature(boolean) (default: true) and wires it through bundler/studio so templates (like Recorder) can disable Ask AI and avoid Cmd/Ctrl+I conflicts.


JonnyBurger:
Thanks for the PR, however the issue was already assigned. Please do not do this, as we will end up with duplicate work which is a bummer for all of us :( We went with the PR of the original assigned developer now.

PR #6297, opened 2026-01-15
Send isStill & isProduction in usage event

### LINKED ISSUE #6214: `@remotion/licensing`: Support `isStill` and `isDevelopment` render
We now have these fields in the backend, but we have yet to send them in the frontend

pskd73:
fixes #6214 

| Function | isStill | isProduction |
| -- | -- | -- |
| `renderMediaOnWeb` |  `false` | NA - `host` being passed |
| `renderStillOnWeb` | `true` | NA - `host` being passed |
| `renderMedia` | `false` | Explicit |
| `renderStill` | `true` | Explicit |
| `renderMediaOnLambda` | `false` | Explicit |
| `renderStillOnLambda` | `true` | Explicit |

### DIFF
--- a/packages/renderer/src/render-still.ts
+++ b/packages/renderer/src/render-still.ts
@@ -1,4 +1,4 @@
-import {LicensingInternals, } from '@remotion/licensing';
+import {LicensingInternals} from '@remotion/licensing';
 import fs, {statSync} from 'node:fs';
 import path from 'node:path';
 import type {_InternalTypes} from 'remotion';
@@ -433,7 +433,6 @@ const internalRenderStillRaw = (
 					succeeded: true,
 					isStill: true,
 					isProduction: options.isProduction ?? true,
-				
 				})
 					.then(() => {
 						Log.verbose(options, 'Usage event sent successfully');


packages/renderer/src/options/is-production-key.tsx
```diff
@@ -0,0 +1,42 @@
...
+const cliFlag = 'is-production-key' as const;
+
+let currentIsProductionKey: boolean | null = null;
+
+export const isProductionKeyOption = {
+	name: 'Is Production Key',
+	cliFlag,
+	description: () => (
+		<>
+			Pass "true" if you are rendering in production. All the production renders
```
pskd73: This needs to be iterated

packages/lambda-client/src/render-still-on-lambda.ts
```diff
@@ -52,6 +52,7 @@ type OptionalParameters = {
 	forcePathStyle: boolean;
 	storageClass: StorageClass | null;
 	requestHandler: RequestHandler | null | undefined;
+	isProduction: boolean | null;
```
pskd73, line 55: It is being Partial<> so no undefined

packages/serverless/src/handlers/launch.ts
```diff
@@ -724,7 +725,12 @@ export const launchHandler = async <Provider extends CloudProvider>({
 		});
 		clearTimeout(webhookDueToTimeout);
 
-		sendTelemetryEvent(params.licenseKey ?? null, params.logLevel);
+		sendTelemetryEvent(
+			params.licenseKey ?? null,
+			params.logLevel,
+			null,
```
pskd73: Is this correct?

packages/lambda/src/cli/commands/render/render.ts
```diff
@@ -358,6 +358,9 @@ export const renderCommand = async ({
 			parsedLambdaCli[BrowserSafeApis.options.licenseKeyOption.cliFlag] ?? null,
 		storageClass: parsedLambdaCli['storage-class'] ?? null,
 		requestHandler: null,
+		isProduction:
+			parsedLambdaCli[BrowserSafeApis.options.isProductionKeyOption.cliFlag] ??
```
JonnyBurger: the flag is oddly named

`--is-production-key` 🤔

how about `--production` instead?

JonnyBurger: Also, we need to add it to `BooleanFlags`

packages/renderer/src/options/is-production-key.tsx
```diff
@@ -0,0 +1,42 @@
+import type {AnyRemotionOption} from './option';
+
+const cliFlag = 'is-production-key' as const;
+
+let currentIsProductionKey: boolean | null = null;
+
+export const isProductionKeyOption = {
+	name: 'Is Production Key',
```
JonnyBurger: There are no "production keys". This might be a confusion with "public keys" and "private keys".
There are only production renders.

packages/renderer/src/options/is-production-key.tsx
JonnyBurger: Yes, should also mention that `true` is the default

PR #6334, opened 2026-01-21
`@remotion/licensing`: Rename `apiKey` to `licenseKey`

### LINKED ISSUE #6125: Should rename all `apiKey` options to `licenseKey`
`apiKey` is a bad name because people think that the render happens on our server. Therefore we decided to rename it to `@remotion/licensing`

- [ ] `@remotion/licensing` 
- [ ] `renderMedia`
- [ ] `renderStill`
- [ ] `renderMediaOnLambda`
- [ ] `renderStillOnLambda`
- [ ] `render_still_on_lambda_payload.rb`
- [ ] `render_media_on_lambda_payload.rb`
- [ ] `packages/lambda-python/remotion_lambda/models.py`
- [ ] `packages/lambda-php/src/RenderParams.php`
- [ ] `packages/lambda-go`

They don't have to be done all at once and it can be split up into multiple PRs. 

Use caution in refactoring rather than trying to quickly get it done.

Mark the old props as `@deprecated` and document the new name in JSDoc.

Also update the corresponding docs and mark the old name as deprecated, and link to the new name instead. Like in `OffthreadVideo`, where we changed the name from `startFrom` to `trimBefore`, get inspired by how we do it there https://github.com/remotion-dev/remotion/blob/main/packages/docs/docs/offthreadvideo.mdx

pskd73:
fixes #6125

### DIFF
--- a/packages/lambda-php/src/RenderParams.php
+++ b/packages/lambda-php/src/RenderParams.php
@@ -201,7 +201,7 @@ class RenderParams
             'webhook' => $this->getWebhook(),
             'forceHeight' => $this->getForceHeight(),
             'forceWidth' => $this->getForceWidth(),
-            'licenseKey' => $this->getLicenseKey() ?? $this->getLicenseKey(),
+            'licenseKey' => $this->getLicenseKey(),
             'offthreadVideoCacheSizeInBytes' => $this->getOffthreadVideoCacheSizeInBytes(),
             'mediaCacheSizeInBytes' => $this->getMediaCacheSizeInBytes(),
             'offthreadVideoThreads' => $this->getOffthreadVideoThreads(),


pskd73:
Things to verify

For each public function
- Removed `apiKey` from internal functions
- Marked `apiKey` deprecated in JSDoc
- Properly done `licenseKey: apiKey ?? licenseKey ?? null` while passing payload to internal functions
- Marked `apiKey` deprecated in the docs
- Added `licenseKey` to the docs

packages/docs/docs/lambda/cli/render.mdx
```diff
@@ -302,7 +302,7 @@ Lets you set a custom user agent that the headless Chrome browser assumes.
 
 Passes `forcePathStyle` to the AWS S3 client. If you don't know what this is, you probably don't need it.
 
-### `--api-Key`<AvailableFrom v="4.0.253"/>
+### `--api-key`<AvailableFrom v="4.0.253"/>
```
JonnyBurger: Here, api-key should be marked as deprecated and license-key should be documented!

packages/lambda-php/src/RenderParams.php
```diff
@@ -202,7 +201,7 @@ public function serializeParams()
             'webhook' => $this->getWebhook(),
             'forceHeight' => $this->getForceHeight(),
             'forceWidth' => $this->getForceWidth(),
-            'licenseKey' => $this->getLicenseKey(),
+            'licenseKey' => $this->getLicenseKey() ?? $this->getLicenseKey(),
```
JonnyBurger: Is it meant to be `this->getApiKey()`

PR #6335, opened 2026-01-21
`@remotion/web-renderer`: Fix low fps when capturing R3F canvas

### LINKED ISSUE #6333: Web renderer: resulting FPS is lower when using `@remotion/three`
When rendering the `packages/example/src/ThreeBasic/index.tsx` composition at 60fps using the web-renderer package, the output doesn't look like 60fps. Frame analysis reveals ~7-10% of frames are duplicated (same frame captured twice in a row), causing stuttery playback.

Reproduction:

1. cd packages/examples`, `bun dev`, http://localhost:3000/three-basic
2. Set FPS to 60 in `Root.tsx`
3. Render on web

samohovets:
Fixes #6333 

Multiple problems are present:

1. Even though remotion_renderReady was already true (because flushSync made React update synchronously, and there were no delayRender handles blocking), we were unnecessarily waiting for a RAF (~16.6ms at 60fps) before even checking! `With check()` first, we at least avoid one unnecessary RAF cycle at the start of each frame's wait.
3. We're still getting duplicate frames at the start of the video in the client side render (investigating.

## Root Causes & Fixes
 1. `@remotion/web-renderer`: **Unnecessary RAF delay**. The `waitForReady` function was waiting for a requestAnimationFrame before checking if the frame was ready, adding ~16.6ms delay per frame even when already ready. The fix of #6268.

**Fix:** Check `remotion_renderReady` first before waiting for RAF.

 3. `@remotion/three`: **R3F reconciler timing**. R3F uses a separate React reconciler that batches context updates via microtasks. When Remotion updates frames rapidly with `flushSync`, R3F's reconciler doesn't immediately see the new frame value, causing it to render the previous frame.

**Fix:** Changed from `frameloop="demand"` + `invalidate()` to `frameloop="never"` + `advance()` with explicit `delayRender`/`continueRender` synchronization:
- `useLayoutEffect` in ThreeCanvas creates a delay handle when frame changes
- `ManualFrameRenderer` inside R3F's reconciler calls `advance()`, then `continueRender()`

Users can now also override the `frameloop` prop for preview mode (e.g., `frameloop="demand"`). 
During rendering, it's always `"never"`.

## Test
Added test that renders 300 frames at 60fps and verifies 0 duplicate consecutive frames.

### DIFF
--- a/packages/web-renderer/src/test/three-duplicate-frames.test.tsx
+++ b/packages/web-renderer/src/test/three-duplicate-frames.test.tsx
@@ -74,7 +74,7 @@ test(
 				width: 400,
 				height: 400,
 				fps: 60,
-				durationInFrames: 300,
+				durationInFrames: 30,
 			},
 			inputProps: {},
 			licenseKey: 'free-license',


packages/web-renderer/src/test/three-duplicate-frames.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+				component: ThreeTestComponent,
+				id: 'three-duplicate-test',
+				width: 400,
+				height: 400,
+				fps: 60,
+				durationInFrames: 300,
+			},
+			inputProps: {},
+			licenseKey: 'free-license',
+			onFrame: async (videoFrame) => {
```
JonnyBurger, line 81: This seems like a very slow test (9sec), this could slow us down

Can we not render less frames and still verify that it works?

samohovets: Yep, just tested. If 300, then amount of duplicate frames was around 100 (without the fix), with 30 frames it's around 10. 

Thanks, made it 30 frames!

PR #6344, opened 2026-01-22
Add headerpad linker flag for MacOS compositor Builds

webmonch:
Adds **-headerpad_max_install_names** linker flag when building compositor for macOS.

Enables macOS compositor to be bundled in Electron apps by adding headerpad linker flag for **install_name_tool** compatibility. Without enough header padding path rewriting fails when new path is longer than the original.

### DIFF
--- a/packages/compositor/build.ts
+++ b/packages/compositor/build.ts
@@ -249,9 +249,13 @@ for (const arch of archs) {
 		? `-C link-args=-Wl,-rpath,'$ORIGIN'`
 		: '';
 
-	const optimizations = all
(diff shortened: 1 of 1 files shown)


jmtimmel2:
I am new to this. How come you chose to add this linker flag?

webmonch:
I was making an Electron build for my latest project and ran into issue when bundling remotion binary into electron build that basically said binary's header doesn't have enough space for a new path (original one had to be rewritten with correct one in post  build script). This flag fixes it.

JonnyBurger:
Thank you! 😃

PR #6367, opened 2026-01-24
Skills: Map skill

JonnyBurger:


SasukeUchiha404:
I am pretty new to this, but it seems the way you merged this way pretty clean and uneventful with the number of changes that needed to be made.

PR #6371, opened 2026-01-25
Fix command to start Claude in terminal

dqhieu:


JonnyBurger:
Oops, thanks a lot for catching that!

PR #6391, opened 2026-01-27
Prompt-to-Motion-Graphics template: Compiler improvements and Improve Lambda error UX

ASchwad:
- Fix import stripping to handle multi-line imports, type imports, and combined imports
- Add Img component support to the compiler 
(These two combined should resolve the bug description that you shared)
- Catch and show error message when Lambda is not configured instead of raw error strings


ASchwad:
image
if the Lambda Render call fails, this is the error message we are surfacing now

PR #6392, opened 2026-01-27
feat(studio): preserve web renderer options on retry

### LINKED ISSUE #6327: Web Renderer: Add more items to the retry payload
When retrying, we should take over as many options as possible. All of these seem missing:

- Container
- Image format (for stills)
- Codec
- Download name (may be out name if we implement #6321)
- Scale
- Quality
- Keyframe interval
- Transparency
- Muted
- Audio quality
- Audio codec
- delayRender timeoout
- @remotion/media cache size
- Hardware acceleration
- Licensing + isProduction flag

romeoscript:
closes #6327

### DIFF
--- a/packages/studio/src/helpers/retry-payload.ts
+++ b/packages/studio/src/helpers/retry-payload.ts
@@ -5,8 +5,8 @@ import type {
 	PixelFormat,
 	X264Preset,
 } from '@remotion/renderer';
-import type {HardwareAccelerationOption} from '@remotion/renderer/client';
 import type {RenderJob} from '@remotion/studio-shared';
+import type {WebRendererHardwareAcceleration} from '@remotion/web-renderer';
 import {NoReactInternals} from 'remotion/no-react';
 import type {ClientRenderJob} from '../components/RenderQueue/client-side-render-types';
 import type {RenderModalState, WebRenderModalState} from '../state/modals';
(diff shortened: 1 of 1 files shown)


packages/studio/src/components/RenderModal/WebRenderModal.tsx
```diff
@@ -1,4 +1,11 @@
-import type {LogLevel} from '@remotion/renderer';
+import type {
+	AudioCodec,
+	Codec,
+	LogLevel,
+	StillImageFormat,
+	VideoImageFormat,
```
JonnyBurger: This uses the types from the server renderer - it should use the types from the web renderer instead:

https://www.remotion.dev/docs/web-renderer/types

packages/studio/src/components/RenderModal/WebRenderModal.tsx
```diff
@@ -182,37 +218,58 @@ const WebRenderModal: React.FC<WebRenderModalProps> = ({
...
 	const [logLevel, setLogLevel] = useState(() => initialLogLevel);
 	const [inputProps, setInputProps] = useState(() => defaultProps);
-	const [delayRenderTimeout, setDelayRenderTimeout] = useState(30000);
+	const [delayRenderTimeout, setDelayRenderTimeout] = useState(
+		initialDelayRenderTimeout ?? 30000,
+	);
 	const [mediaCacheSizeInBytes, setMediaCacheSizeInBytes] = useState<
 		number | null
-	>(null);
+	>(initialMediaCacheSizeInBytes ?? null);
```
JonnyBurger: When we expect the `?? null` case to never happen, can we remove it?

romeoscript: yh sure the  ?? null comparison  is redundant anyways

romeoscript:
need me to do something here ? @JonnyBurger

JonnyBurger:
Nope, just slow to react at the moment 😅

Thanks a lot, this is great! 😃 🎉

PR #6396, opened 2026-01-27
 Replace add-skills with skills

tiwariaayu:
This PR replaces the deprecated add-skill package with its successor, the skills package, in create-video.

Changes
1-Updated install-skills.ts to execute npx skills add remotion-dev/skills instead of using add-skill.
2-Updated error logs to suggest the correct manual command: npx skills add remotion-dev/skills


JonnyBurger:
Thanks a lot! 🙌

PR #6398, opened 2026-01-27
Prompt-to-Motion-Graphics template: Add follow-up edits via chat

ASchwad:
- Follow-up edits via quick line-based edits
- Propagate API, Runtime and Compilation errors to LLM and allow self correction
- Show used skills and edit tool use within chat


ASchwad:
Example for self-correction loop if the Models runs into API, Runtime or Compilation issues: 
image

JonnyBurger:
Siick 🔥

Issue #6453, opened 2026-02-04
Responsiveness issue in our new pricing component


JonnyBurger:
<img width="1179" height="1342" alt="Image" src="https://github.com/user-attachments/assets/8998e3d0-d404-43d9-b9ef-389c2a69d2d6" />

Vaibhav-Vanjani:
@JonnyBurger assign it to me. I'm working in it

<img width="1172" height="812" alt="Image" src="https://github.com/user-attachments/assets/f7937488-d58a-4ced-ac3b-c8e5cebc55d8" />

PR #6459, opened 2026-02-04
fix responsiveness issue

Vaibhav-Vanjani:
fixed Responsiveness issue in our new pricing component #6453

### DIFF
--- a/packages/design/src/Switch.tsx
+++ b/packages/design/src/Switch.tsx
@@ -7,7 +7,7 @@ export const Switch: React.FC<{
 	return (
 		<div
 			data-active={active}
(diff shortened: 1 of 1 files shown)


JonnyBurger:
Wow, thanks a lot for taking a look at a task for our business 🙈

image

Left: After, right: before. Hmm, it feels like we still have a rather cluttery layout even after the change

Vaibhav-Vanjani:
Hi @JonnyBurger, Can you check the layout and let me know if it looks okay or still cluttered? If it’s not quite right, I’d appreciate your input on how you’d like it to be.

updated-remotion-ui

JonnyBurger:
Definitely! 

Thanks a lot for the willingness to help! :))

PR #6464, opened 2026-02-05
Update media playback error documentation

tt0mmy:
Removed mention of unsupported codec '.m3u8' in playback error documentation since Chrome does support this.

https://caniuse.com/?search=M3U8

image


JonnyBurger:
Fair enough!

PR #6475, opened 2026-02-07
Update homepage and pricing table

MehmetAdemi:
I made changes to the pricing table an homepage in general.

Slider for automators now also gives an info message when 50'000 renders per month are surpassed, indicating that at this point the user is eligible for the enterprise license.

Fix contrast issues with the slider in dark theme.

Community Stats have been updated.


MehmetAdemi:
Yes, indeed! 

I'll try to add that logic as well.

MehmetAdemi:
@JonnyBurger can you please check it out now?

Also, please pay attention for the to the tooltip message for the Creator Seat.

JonnyBurger:
Thanks. The only change I've made is `Creator Seat` -> `Seat` as this is the terminology we use everywhere else.

PR #6487, opened 2026-02-09
Add voiceover skill using ElevenLabs TTS

JonnyBurger:
## Summary

- Ports the voiceover skill from https://github.com/remotion-dev/skills/pull/3 by @rohailaltaf into this monorepo
- Adds `rules/voiceover.md` with guidance for generating AI voiceovers using ElevenLabs TTS
- Covers scene config, audio generation, dynamic duration via `calculateMetadata`, and rendering audio in components

## Test plan

- [ ] Verify `voiceover.md` renders correctly
- [ ] Verify SKILL.md index link works

### DIFF
--- a/packages/skills/skills/remotion/rules/voiceover.md
+++ b/packages/skills/skills/remotion/rules/voiceover.md
@@ -37,7 +37,7 @@ const response = await fetch(
   {
     method: "POST",
     headers: {
-      "xi-api-key": process.env.ELEVENLABS_API_KEY,
+      "xi-api-key": process.env.ELEVENLABS_API_KEY!,
       "Content-Type": "application/json",
       Accept: "audio/mpeg",
     },


JonnyBurger:
Thanks a lot!

Tested it:

> use remotion best practices.
> create two scenes. use stock images of snow / winter and add a slight zoom in while they are showing.
> create a voiceover for approximately each 10 seconds, describing the pictures.
> add a light leak transition between them

and worked well! 

https://github.com/user-attachments/assets/1ea152d6-0907-4db5-8caf-766b044192d9

I condensed the skill and added more cross-references so it uses less tokens.

devagul93:
The MUST NOT fall back to other TTS tools instruction is too restrictive — many users
  have existing voiceover workflows on other platforms or tools they prefer. The skill
  could be more useful by keeping ElevenLabs as the recommended example while removing the
   hard prohibition, so users can adapt the pattern to whatever TTS they're already using.
  Happy to contribute a follow-up PR making it provider-agnostic if there's interest.

JonnyBurger:
> The MUST NOT fall back to other TTS tools instruction is too restrictive — many users have existing voiceover workflows on other platforms or tools they prefer. The skill could be more useful by keeping ElevenLabs as the recommended example while removing the hard prohibition, so users can adapt the pattern to whatever TTS they're already using. Happy to contribute a follow-up PR making it provider-agnostic if there's interest.

Good point!
I would merge a better version, wanna send a PR?

devagul93:
I'll send a PR 👍

Issue #6322, opened 2026-01-20
Web Renderer: Persist the client-side rendering queue


JonnyBurger:
To be done after #6321 

As mentioned there, we lose all state on a page refresh, even on a bad hot reload.
The server queue feels much better because it survives page reloads. We should also persist - either on the client or on the server.

samohovets:
Fixed in #6498

PR #6498, opened 2026-02-10
`@remotion/studio`: persist and upload client-side rendered videos into the output directory

### LINKED ISSUE #6321: Web Renderer: Upload client-side rendered videos into `out/` directory
After consideration, I am with you @samohovets!

It makes sense to upload the videos rendered with the Studio to the `out/` directory, and when you click the item in the render panel, it serves it from the HTTP server.

- If you accidentially close the tab, you lose all your renders
- Same as if you hot reload and encounter a syntax error, or change React hooks, the page needs to reload -> all renders are gone
- When calling `renderMediaOnWeb()`, we clear previous renders from WebFS! This can break the preview of renders, fixed if we put it in the out/ directory
- Align the behavior with SSR

### LINKED ISSUE #6515: `@remotion/studio`: Render modal does not respect `Config.setOutputLocation()`
## Description

The Studio render modal (both server-side and client-side) ignores `Config.setOutputLocation()` set in `remotion.config.ts`. The output name always defaults to `out/{compositionName}.{ext}` regardless of the configured output location.

## Steps to reproduce

1. Set `Config.setOutputLocation('renders')` in `remotion.config.ts`
2. Open Studio and click Render
3. Observe the output name defaults to `out/MyComp.mp4` instead of `renders/MyComp.mp4`

## Expected behavior

The render modal should use the configured output location as the default. If the value is a full file path (e.g., `renders/video.mp4`), use it as-is. If it's a directory (e.g., `renders`), use it as the output directory prefix instead of `out/`.

## Root cause

`Config.setOutputLocation()` is evaluated when the Studio loads `remotion.config.ts`, but the value is stored in `ConfigInternals` and never passed to the frontend via `RenderDefaults`. Only CLI renders (`npx remotion render`) respect it.

This has been the case since `Config.setOutputLocation()` was introduced in v3.1.6.

samohovets:
Fixes #6321 and #6322 in one go.

Also fixes #6515 — the Studio render modal now respects `Config.setOutputLocation()` from `remotion.config.ts` for both server-side and client-side renders. If the value is a full file path (e.g., `renders/video.mp4`), it's used as-is. If it's a directory (e.g., `renders`), it replaces the default `out/` prefix.

### DIFF
--- a/packages/studio-server/src/routes.ts
+++ b/packages/studio-server/src/routes.ts
@@ -20,13 +20,13 @@ import {
 	removeCompletedClientRender,
 } from './client-render-queue';
 import {getFileSource} from './helpers/get-file-source';
-import {resolveOutputPath} from './helpers/resolve-output-path';
 import {getInstalledInstallablePackages} from './helpers/get-installed-installable-packages';
 import {
 	getDisplayNameForEditor,
 	guessEditor,
 	launchEditor,
 } from './helpers/open-in-editor';
+import {resolveOutputPath} from './helpers/resolve-output-path';
 import {allApiRoutes} from './preview-server/api-routes';
 import type {ApiHandler, QueueMethods} from './preview-server/api-types';
 import {getPackageManager} from './preview-server/get-package-manager';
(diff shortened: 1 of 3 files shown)


JonnyBurger:
Very nice! Thanks a lot!

I noticed #6528, but that was already a problem before, I will not stop the merge.

Issue #6493, opened 2026-02-10
Firefox bad audio encoding


JonnyBurger:
[From message posted on Discord by **london_fed**](https://discord.com/channels/809501355504959528/1437812025605754900/1470445426359537878)
----
Thanks Jonny....According to CoPilot:

```
What you're giving Remotion:

✅ Original MP3/WAV/M4A file (whatever user uploaded)
✅ Original sample rate
✅ Original bitrate
✅ Original encoding
✅ Completely untouched
The only place audio might be transformed is inside Remotion itself when it:

Decodes your audio
Re-encodes it into the final video's audio track (AAC/Opus depending on container)
```

I am using firefox `140.7.0esr (64-bit)` if that helps

JonnyBurger:
**Root cause**: Firefox's `AudioDecoder` outputs audio in `f32`/`f32-planar` format, while Chrome outputs `s16-planar`. When `convertAudioData()` called `audioData.copyTo(dest, {format: 's16'})` directly, Firefox's native f32→s16 conversion produced different results than expected, leading to bad-sounding audio.

**Fix**: When the decoded `AudioData.format` is `f32` or `f32-planar` (Firefox), we now normalize through an explicit `f32-planar` intermediate `AudioData` before converting to s16. This ensures the final s16 samples are consistent regardless of browser. When the format is already an s16 variant (Chrome), we keep the direct `copyTo` fast path to avoid unnecessary allocations.

PR #6505, opened 2026-02-11
#6493 Block AAC on Firefox

tiwariaayu:
image
-->Root Cause
Firefox has a known bug where its AAC encoder produces corrupted/glitchy audio output, even though it claims to support AAC via `MediaRecorder` and WebCodecs API. The current fix (added in #6493) detects this combination and fails fast to prevent wasted renders.

However, this "fail fast" approach is too aggressive - we should automatically fall back to Opus (which works) instead of forcing the user to manually change their code.

--> Solution
Changed the severity from `'error'` to `'warning'` and automatically return `'opus'` as the fallback codec when Firefox + AAC is detected.

### DIFF
--- a/packages/web-renderer/src/resolve-audio-codec.ts
+++ b/packages/web-renderer/src/resolve-audio-codec.ts
@@ -40,32 +40,32 @@ export const resolveAudioCodec = async (options: {
 	const mediabunnyAudioCodec = audioCodecToMediabunnyAudioCodec(audioCodec);
 	const canEncode = await canEncodeAudio(mediabunnyAudioCodec, {bitrate});
 
-	if (canEncode) {
-		return {codec: audioCodec, issues};
-	}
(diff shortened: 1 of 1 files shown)


JonnyBurger:
I cannot confirm this, the bug already happens when encoding with Opus - AI hallucination?

tiwariaayu:
Root cause can be --> The platform encoder problem or sample rate mismatch
The Opus encoder in Firefox is built-in (not OS-dependent) and is much more robust because it's the engine powering Google Meet, Zoom (web), and Discord. It handles these sample-rate irregularities much more gracefully than the fragile AAC bridge does.

JonnyBurger:
This is not the issue at all.
Firefox already does use Opus over AAC! The audio is still garbled.

Will actually solve this hopefully in #6507

PR #6506, opened 2026-02-11
Remotion on Vercel updates to docs and template

anthonyshew:
- Updates documentation about Remotion on Vercel for clarity
- Updates template for faster first-render experience by prebuilding the snapshot


package.json
```diff
@@ -76,6 +76,7 @@
 			"!packages/template-tiktok/whisper.cpp/**",
 			"!packages/template-audiogram/whisper.cpp",
 			"!packages/template-audiogram/whisper.cpp/**",
+			"!packages/template-vercel/**",
```
anthonyshew, line 79: This template is now a nested Turborepo, which doesn't work great. It's a little bit of an anti-pattern, but needed for demos and templates.

This is how we handle it in the Turborepo repository and looks like some of the templates here follow this pattern as well.

PR #6530, opened 2026-02-16
`@remotion/studio`: Show render queue in read-only mode

### LINKED ISSUE #6528: Web Renderer: Studio does not show render queue in read-only mode
```
cd packages/example
bunx remotion bundle
bunx serve build
```

Now if we render a video client-side, we correctly download the video instead of trying to upload, but the render queue is not shown at all.

samohovets:
## Summary

Fixes #6528

When the Studio runs in read-only mode (bundled via `remotion bundle` + static serve, or after the dev server is stopped), the render queue panel is completely hidden — even though client-side browser rendering works correctly.

## Changes

- **`OptionsPanel.tsx`**: Show the "Renders" tab and `<RenderQueue />` component in read-only mode when `SHOW_BROWSER_RENDERING` is enabled
- **`RenderButton.tsx`**: Auto-fallback from `server-render` to `client-render` when the server disconnects and browser rendering is available
- **`RenderQueue/index.tsx`**: Don't show "The studio server has disconnected" when browser rendering is enabled — show the queue instead

## Testing

**Scenario 1 — Bundled mode:**
```
cd packages/example
bunx remotion bundle
bunx serve build
```
Open in browser → "Renders" tab is visible → client-side renders appear in the queue.

**Scenario 2 — Dev server disconnected:**
```
cd packages/example
bun run dev
```
Open Studio → Ctrl+C to stop server → render button auto-switches to "Render on web" → renders appear in queue instead of the disconnected message.

Both scenarios require `Config.setExperimentalClientSideRenderingEnabled(true)` in `remotion.config.ts` (already set in `packages/example`).


packages/studio/src/components/RenderButton.tsx
```diff
@@ -178,9 +185,21 @@ export const RenderButton: React.FC<{readonly readOnlyStudio: boolean}> = ({
...
 		.previewServerState.type;
+
+	// auto-fallback to client-render when server disconnects and browser rendering is available
+	useEffect(() => {
+		if (
+			connectionStatus === 'disconnected' &&
+			SHOW_BROWSER_RENDERING &&
+			renderType === 'server-render'
+		) {
+			setRenderType('client-render');
```
JonnyBurger: No need for a useEffect() here - this should be a derived state

JonnyBurger (changes requested):
Also the `if` statements are crazy now, we have

```ts
connectionStatus === 'disconnected' && !SHOW_BROWSER_RENDERING
readOnlyStudio && !SHOW_BROWSER_RENDERING (x3)
connectionStatus === 'disconnected' && SHOW_BROWSER_RENDERING
connectionStatus === 'connected' || SHOW_BROWSER_RENDERING
connectionStatus !== 'connected' && !SHOW_BROWSER_RENDERING
```

Surely we can find some way to simplify this?

packages/studio/src/components/RenderButton.tsx
samohovets: Good catch! I indeed can derive here

Issue #6558, opened 2026-02-17
Studio: Better handling missing module


JonnyBurger:
Let's say we have code in our Studio:

```ts
import {MyComp} from "./does-not-exist"
```

We get an overlay immediately, but I think it does not recover.
When we reload the page, we get a white page!

PR #6559, opened 2026-02-17
Studio: Better handling missing module #6558

tiwariaayu:
The error overlay and hot middleware are now initialized before any user code executes. This ensures that even if the bundle fails early (for example due to a missing module), the error handling layer is already active and can display the build error instead of showing a blank page.
The global guard prevents duplicate initialization during hot reloads, and the defensive try/catch guarantees that failures in dev tooling do not crash the Studio itself.


JonnyBurger:
This does seem to help! Thank you!

Issue #6563, opened 2026-02-17
Studio: Cannot hot reload Root.tsx after invalid module anymore


JonnyBurger:
After triggering a bug like in https://github.com/remotion-dev/remotion/issues/6558, making changes in Root.tsx does not update the compositions anymore until we reload it

PR #6582, opened 2026-02-18
#6563 Cannot hot reload Root.tsx after invalid module anymore 

tiwariaayu:


JonnyBurger:
Tested and does not work.
Please test this case yourself before you submit a PR...

PR #6597, opened 2026-02-19
`@remotion/renderer`: Surface signal when FFmpeg process is killed

JonnyBurger:
## Summary
- When FFmpeg is killed by a signal (e.g. SIGKILL, SIGTERM), the `exit` handler in `prespawn-ffmpeg.ts` now correctly treats it as an error instead of `quit-successfully`
- Added `signal` field to the `quit-with-error` variant of `RunningStatus`
- Error messages in `render-media.ts` now include the signal name when present

## Test plan
- [ ] Verify that a signal-killed FFmpeg process during parallel encoding surfaces the signal in the error message
- [ ] Verify that normal FFmpeg exit (code 0, no signal) still reports `quit-successfully`

### DIFF
--- a/packages/renderer/src/render-media.ts
+++ b/packages/renderer/src/render-media.ts
@@ -696,13 +696,13 @@ const internalRenderMediaRaw = ({
 								const exitStatus = preStitcher?.getExitStatus();
 								if (exitStatus?.type === 'quit-successfully') {
 									throw new Error(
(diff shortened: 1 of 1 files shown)


greeenboi:
thank you for the assist 🫡 @JonnyBurger  (frm discord)

lmk if i can help with testing this pr

JonnyBurger:
Yes totally, please test the next version and post the new error message

then we can figure out the actual error and fix that!

PR #6618, opened 2026-02-20
`@remotion/studio`: When duplicating composition, jump to new one

### LINKED ISSUE #5111: Studio: Once a composition is "duplicated", it should jump to it
That's the only part about this flow that doesn't make sense yet

JonnyBurger:
## Summary
- After duplicating a composition, the studio now automatically navigates to the new composition
- Adds an optional `onSuccess` callback to `CodemodFooter`, called after the codemod is successfully applied
- `DuplicateComposition` uses this callback to push the URL and set the canvas content to the new composition ID

Previously, after duplicating a composition the user stayed on the original composition, which was confusing.

Closes #5111

## Test plan
- [ ] Right-click a composition → Duplicate → confirm duplication
- [ ] Verify the studio navigates to the newly created composition after duplication
- [ ] Verify rename and delete codemods still work as before (no regression from the optional `onSuccess` prop)

### DIFF
--- a/packages/studio/src/components/InitialCompositionLoader.tsx
+++ b/packages/studio/src/components/InitialCompositionLoader.tsx
@@ -94,6 +94,9 @@ export const InitialCompositionLoader: React.FC = () => {
 		const canvasContentFromUrl = deriveCanvasContentFromUrl();
 
 		if (canvasContent) {
+			// If the URL points to a different composition than the one currently
+			// displayed, switch to it. This handles the case where the URL is
+			// updated externally (e.g. after duplicating a composition).
 			if (
 				canvasContentFromUrl &&
 				canvasContentFromUrl.type === 'composition' &&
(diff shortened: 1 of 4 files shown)


packages/studio/src/components/InitialCompositionLoader.tsx
```diff
@@ -91,11 +91,26 @@ export const InitialCompositionLoader: React.FC = () => {
 	const staticFiles = useStaticFiles();
 
 	useEffect(() => {
+		const canvasContentFromUrl = deriveCanvasContentFromUrl();
+
 		if (canvasContent) {
+			if (
```
JonnyBurger, line 100: A comment explaining what this does would be good

packages/studio/src/components/NewComposition/CodemodFooter.tsx
```diff
@@ -18,6 +18,7 @@ export const CodemodFooter: React.FC<{
 	readonly errorNotification: string;
 	readonly genericSubmitLabel: string;
 	readonly submitLabel: (options: {relativeRootPath: string}) => string;
+	readonly onSuccess?: () => void;
```
JonnyBurger: Let's make it mandatory, but nullable

JonnyBurger (commented):
@pullfrog can you

PR #6662, opened 2026-02-24
`@remotion/web-renderer`: Adds font-style support

### LINKED ISSUE #6657: add support for font-style in web-renderer
# Feature Request 🛍️

Currently web renderer only supports fontWeight, fontSize and fontFamily properties. Would love the support for fontStyle property as defined [here](https://github.com/remotion-dev/remotion/blob/b53f3fd06453c8144f63ea50f3acb2a1d6236821/packages/web-renderer/src/drawing/text/draw-text.ts#L48C3-L48C70)

## Use Case

To support italic font styles

## Possible Solution

ctx.font supports it so we just need to pass the property down from computedStyles.

saimfd:
- Adds support for `font-style` css property in web renderer
- Updated limitations doc file

Closes #6657


JonnyBurger:
Super nice! Thanks a lot for adding this! 🙌

PR #6689, opened 2026-02-25
`@remotion/design`: prevent Button from stretching to full width

### LINKED ISSUE #6684: `@remotion/design`: New accessible button leads to full width
e.g. here:

https://www.remotion.dev/templates

Image

nielskaspers:
## Summary
- Changed `display: flex` to `display: inline-flex` on the `<Button>` component so it sizes to its content instead of stretching to fill the parent container
- Removed a duplicate `flex` Tailwind class in the shared styles

Callers that need full-width buttons already pass `w-full` via `className` (e.g. `templates.tsx`, `GetStartedStrip.tsx`), so existing usage is unaffected.

## Test plan
- [ ] Verify buttons on `/templates` page still fill their grid cells (they pass `w-full`)
- [ ] Verify buttons elsewhere (e.g. GetStartedStrip, MoreTemplatesButton) no longer stretch unexpectedly

Fixes #6684

### DIFF
--- a/packages/design/src/Button.tsx
+++ b/packages/design/src/Button.tsx
@@ -98,7 +98,7 @@ export const Button: React.FC<ButtonProps> = ({
 
 	const sharedClasses = cn(
 		'text-text',
-		'flex',
+		'inline-flex',
 		'justify-center',
 		'bg-button-bg',
 		'items-center',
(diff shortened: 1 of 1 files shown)


JonnyBurger:
Thanks a lot, I think that's it!

PR #6707, opened 2026-02-25
Replace Zod schema with plain objects in `useSchema` / `SequenceControls`

JonnyBurger:
## Summary
- Replaces Zod schema definitions (`z.object(...)`) with plain vanilla objects (`{ type: 'number', min: 0, ... }`) for the `SequenceControls` / `useSchema` system (timeline field controls)
- Defines new `SequenceSchema` and `SequenceFieldSchema` types in `packages/core/src/sequence-field-schema.ts`
- Removes zod dependency from `@remotion/media` and `@remotion/light-leaks` for timeline schemas
- Studio timeline components now read field properties (`min`, `max`, `step`, `default`, `description`) directly from plain objects instead of using zod introspection utilities

## Test plan
- [x] `bun run build` passes
- [x] `bun run stylecheck` passes
- [ ] Manual: open studio, add a `<Video>` or `<Audio>`, expand timeline track → verify number draggers and boolean checkboxes still work with correct min/max/step

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -6,7 +6,7 @@
       "name": "remotion-monorepo",
       "dependencies": {
         "p-limit": "7.2.0",
-        "turbo": "2.8.10",
+        "turbo": "2.8.11",
         "typescript": "5.9.3",
       },
       "devDependencies": {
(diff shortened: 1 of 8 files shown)


packages/studio-server/src/preview-server/routes/save-sequence-props.ts
```diff
@@ -49,7 +49,7 @@ export const saveSequencePropsHandler: ApiHandler<
 		RenderInternals.Log.info(
 			{indent: false, logLevel},
 			RenderInternals.chalk.blueBright(
-				`${fileLink} updated: ${key} ${oldValueString} \u2192 ${newValueString}`,
+				`${fileLink} updated: ${key} ${oldValueString.trim()} \u2192 ${newValueString}`,
```
JonnyBurger: @pullfrog revert this, it was not necessary

PR #6625, opened 2026-02-21
`@remotion/cli`: Migrate remaining flags to AnyRemotionOption for unknown flag detection

Vaibhav-Vanjani:
#6605 

## Summary

This PR migrates the remaining CLI flags to the `AnyRemotionOption` system as a prerequisite for enabling unknown flag detection in the CLI.

By ensuring all flags are registered via `AnyRemotionOption`, the CLI can reliably detect and warn (or error) when users pass unrecognized flags.

---

## Changes

**Modified files:**

- `packages/cli/src/initialize-cli.ts`
- `packages/cli/src/parsed-cli.ts`
- `packages/renderer/src/options/index.tsx`

**Newly added files:**

- `packages/renderer/src/options/browser-args.tsx`
- `packages/renderer/src/options/concurrencies.tsx`
- `packages/renderer/src/options/force.tsx`
- `packages/renderer/src/options/help.tsx`
- `packages/renderer/src/options/image-format.tsx`
- `packages/renderer/src/options/output.tsx`
- `packages/renderer/src/options/png.tsx`
- `packages/renderer/src/options/quality.tsx`
- `packages/renderer/src/options/quiet.tsx`

---

## Testing

Warns about unknown command line flags

6605_remotion_screenshot

---

## Follow-Up

Please let me know if any changes are required or if additional information is needed.

Issue #6605, opened 2026-02-20
Warn about unknown command line flags


JonnyBurger:
In order to warn about unknown CLI flags, all flags need to be migrated to the `AnyRemotionOption` system. Once all flags are migrated, the CLI can detect and warn about any unrecognized flags passed by the user.

## Prerequisites

The following flags still need to be refactored to the options system:

- [ ] `--output`
- [ ] `--frames`
- [ ] `--quiet` / `-q`
- [ ] `--help`
- [ ] `--force`
- [ ] `--timeout`
- [ ] `--concurrencies`
- [ ] `--browser-args`
- [ ] `--license-key`
- [ ] `--image-format` (deprecated alias)
- [ ] `--png` (deprecated, throws error)
- [ ] `--quality` (deprecated alias)

## Implementation

Once all flags are migrated, implement a check that compares `parsedCli` keys against the known set of `AnyRemotionOption` `cliFlag` values and warns (or errors) when an unknown flag is detected.

Vaibhav-Vanjani:
Hi @JonnyBurger, I’ve raised a PR for this issue. Just keeping you in the loop!

JonnyBurger:
Just a note: we are not accepting external PRs for this issue at the moment. We've now closed three PRs (#6625, #6721, #6846) because the implementation approach doesn't match what we have in mind — specifically, we need per-command flag detection, typesafe flag registration (so missing flags cause compile errors), proper alias handling, and config file + documentation updates for all migrated flags.

We'll handle this internally when we're ready. Thanks for the interest!

PR #6721, opened 2026-02-27
Per-command unknown flag detection for `render`

Vaibhav-Vanjani:
#6605: Unknown flag detection for `render` command

This PR implements unknown flag detection specifically for the `render` command, following the per-command approach discussed in [this comment](https://github.com/remotion-dev/remotion/pull/6625#issuecomment-3943733829). Once this approach is approved, I can refactor it for other commands.

## Summary
- Introduces `valid-flags/render.ts` listing all valid render flags from the [Remotion CLI documentation](https://www.remotion.dev/docs/cli/render).
- Implements `warnAboutUnknownFlag` to log mistyped or unsupported flags.
- Updates `render.tsx` to warn users about unknown flags before executing the render flow.

## Example behavior
Passing an unknown flag like `--typdksj` prints:

```text
Unknown flag "--typdksj" for render. Run "remotion render --help" to see valid options.

### DIFF
--- a/packages/cli/src/unknown-flag-detect.ts
+++ b/packages/cli/src/unknown-flag-detect.ts
@@ -14,9 +14,11 @@ export const warnAboutUnknownFlag = function (
 	});
 
 	for (const flag of unknownFlags) {
+	for (const flag of unknownFlags) {
 		Log.warn(
 			{indent: false, logLevel},
-			`Unknown flag "--${flag}" for ${command}. Run "remotion render --help" to see valid options.`,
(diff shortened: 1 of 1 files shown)


JonnyBurger:
Thanks, but I think this will not do it -

- Some flags are missing
- The code does not actually compile
- This requires us to keep a manual, non-typesafe list of supported flags which I fear if we forget can lead to errors.

So, I'll try to take this into my own hands and try to make something which satisfies these criteria.. thanks nonetheless for the PR!

PR #6726, opened 2026-02-27
`@remotion/media`: Share single audioSyncAnchor across all MediaPlayers

### LINKED ISSUE #6727: Premounting does not work as indicator would lead you to believe
```ts
import {Video} from '@remotion/media';
import React from 'react';
import {Sequence, staticFile, useVideoConfig} from 'remotion';

const SLICE_DURATION_FRAMES = 3; // 0.1sec at 30fps
const NUM_SLICES = 10;
const PREMOUNT_SEC = 0.5;

export const SlicedVideo: React.FC = () => {
	const src = staticFile('sliced-video.mp4');

	return (
		<Sequence from={15} premountFor={15}>
			{new Array(NUM_SLICES).fill(0).map((_, i) => {
				const from = i * SLICE_DURATION_FRAMES;
				return (
					<Sequence
						key={i}
						from={from}
						durationInFrames={SLICE_DURATION_FRAMES}
						premountFor={PREMOUNT_SEC * fps}
					>
						<Video src={src} trimBefore={from} />
					</Sequence>
				);
			})}
		</Sequence>
	);
};
```

Image

However, 6 videos are premounted on frame 0. This is because the premount is cascading?

JonnyBurger:
## Summary
- All `MediaPlayer` instances now share a single `audioSyncAnchor` (mutable `{value: number}` object) created in `SharedAudioContextProvider`, instead of each maintaining their own
- The shared anchor maps `AudioContext.currentTime` → global Remotion timeline time; each player derives its own media time using its sequence offset and playback rate
- Removes `calculate-playbacktime.ts` (formula is now inline and different)

Closes #6727

## Test plan
- [ ] Play a composition with overlapping `<Video>`/`<Audio>` at different `<Sequence from={...}>` offsets — verify seamless audio timing
- [ ] Verify single video/audio playback still works correctly
- [ ] Verify playback rate changes (both per-player and global) work
- [ ] Verify trim changes work correctly
- [ ] `tsc --noEmit` passes for both `packages/media` and `packages/core`

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,8 +1,0 @@
-# By adding 1 dependency, we remove the defaults
-# The default is: ['*types*', '*eslint*', '@prettier/plugin-*', '*prettier-plugin-*']
-# Is a devDependency of `three` and a dependency of docs. Need to hoist
-public-hoist-pattern[]=@react-three/fiber
-auto-install-peers=true
-strict-peer-dependencies=false
-publish-branch=main
-update-notifier=false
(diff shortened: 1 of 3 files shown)


JonnyBurger:
Closing in favor of #6736

PR #6737, opened 2026-02-28
Fix mediabunny version mismatch in templates

JonnyBurger:
## Summary
- Templates (`template-music-visualization`, `template-recorder`, `template-three`) had mediabunny pinned at `1.34.4` while the catalog specified `1.35.1` — updated them to match
- Added a test in `validate-templates.test.ts` that checks template mediabunny versions match the catalog, preventing future drift
- Improved the `upgrade-mediabunny` skill to clarify that templates use explicit versions (not `catalog:`) and must be updated manually, and added `@mediabunny/ac3` to the upgrade checklist

## Test plan
- [x] `bun test src/templates/validate-templates.test.ts` passes (251/251)
- [x] `bun run stylecheck` passes

### DIFF
--- a/packages/renderer/src/test/render-still.test.ts
+++ b/packages/renderer/src/test/render-still.test.ts
@@ -38,7 +38,7 @@ test(
 	},
 );
 
-test.skip(
+test(
 	'Returns buffer in promise result',
 	async () => {
 		await ensureBrowser();


JonnyBurger:
@pullfrog

revert skipping the test. i have fixed the underlying issue in the meanhwile

PR #6740, opened 2026-02-28
`@remotion/web-renderer`: Fix scaffold scroll regression

samohovets:
### Fix: `renderMediaOnWeb()` scroll regression

This PR addresses a regression where `renderMediaOnWeb()` could make the host page scrollable due to the scaffold node's dimensions.

**Why this change?**
Previously, the scaffold's fixed element with explicit composition `width`/`height` could expand the `document.body` or `documentElement` scroll area. This change prevents unintended scrollability of the host page during rendering.

**What changed?**
1.  **Scaffold mounting strategy:**
    *   A new `position: fixed; inset: 0; overflow: hidden; visibility: hidden; pointer-events: none; z-index: -9999` wrapper is now appended to `document.body`.
    *   The composition root (the actual rendering target) is mounted as an `position: absolute; top: 0; left: 0;` child inside this wrapper, with its `width`/`height` set from the composition.
2.  **Cleanup:** The cleanup process now correctly removes both the outer wrapper and the inner composition root, in addition to unmounting the React root and cleaning up CSS.
3.  **Regression Test:** A new test in `render-media.test.tsx` asserts that `document.documentElement` and `document.body` scroll dimensions do not increase during or after a `renderMediaOnWeb()` call.

**How it works:**
The new fullscreen fixed wrapper effectively isolates the composition's layout from the main document's scroll area. By setting `overflow: hidden` on the wrapper, any content within (including large compositions) is clipped, preventing it from influencing the host page's scroll dimensions.

**Test Evidence:**
*   `packages/web-renderer/src/test/render-media.test.tsx` (including the new scroll regression test) passed in Chromium.
*   `packages/web-renderer/src/test/render-still.test.tsx` also passed, verifying shared scaffold path behavior.

**Edge Cases/Tradeoffs:**
*   Preserves existing APIs and backward compatibility.
*   Handles compositions larger than the viewport without affecting page scroll.
*   Ensures robust cleanup of all DOM nodes.

### DIFF
--- a/packages/web-renderer/src/create-scaffold.tsx
+++ b/packages/web-renderer/src/create-scaffold.tsx
@@ -75,7 +75,6 @@ export function createScaffold<Props extends Record<string, unknown>>({
 
 	const div = document.createElement('div');
 
-	// Match same behavior as in portal-node.ts
 	div.style.position = 'absolute';
 	div.style.top = '0';
 	div.style.left = '0';


samohovets:
@pullfrog

PR #6741, opened 2026-03-01
Fix asset selection highlighting all assets with the same name

samildeli:
The selected state compared only the filename, ignoring the parent folder path. Extract relativePath into a shared useMemo and use it in both the selection check and onClick handler.


JonnyBurger:
Thank you!! 🙌

Issue #3995, opened 2024-06-13
Creating Node API to launch the studio similar to `remotion studio` in Node.js.


JonnyBurger:
[From message posted on Discord by **kitty.sparkles**](https://discord.com/channels/809501355504959528/817306238811111454/1250810382013038697)
----
Hellooooo. 👋  Is there a Node API to start the studio? Basically a Node.js equivalent to `remotion studio`?

alexfernandez803:
@JonnyBurger  I like to take this one, but can you give some guidance where to implement this?

samohovets:
Closing this until we find a viable use-case.

For users who want to start the studio programmatically, you have options:

- [Child Process](https://nodejs.org/api/child_process.html) (Node.js)
- [Execa](https://www.npmjs.com/package/execa)
- [Bun Shell](https://bun.com/docs/runtime/shell), if you use Bun

Use this command to start the Studio app: https://www.remotion.dev/docs/cli/studio

PR #6769, opened 2026-03-03
`@remotion/studio`: Add programmatic Studio startup API

samohovets:
This PR adds programmatic Studio startup as a public `@remotion/studio` API.

It addresses #3995 by making `startStudio(options?)` available from `@remotion/studio` while keeping `remotion studio` and the programmatic API on the same fully functional Studio startup path.

## Summary
- Add `startStudio(options?)` to `@remotion/studio`, with docs and an example script for starting Studio from Node.js / Bun.
- Refactor `@remotion/studio` and `@remotion/cli` into thin adapters over shared Studio startup in `@remotion/studio-server`.
- Move config loading, entry-point discovery, env/render-default resolution, launch-session orchestration, the Studio render queue, and shared render progress/download/artifact helpers into `@remotion/studio-server`.
- Keep `@remotion/cli/config` as the compatibility import path for `remotion.config.ts`, but back it with the shared config implementation.
- Thread Studio preview/render asset paths from callers so lower layers no longer import `@remotion/studio` directly.
- Delete the intermediate `@remotion/studio-startup-core` package and the superseded CLI Studio queue implementation.

## Architecture breakdown
- `@remotion/studio`: Exposes the public `startStudio(options?)` API, applies ergonomic defaults, and resolves package-local Studio asset paths.
- `@remotion/studio/node`: Keeps the internal mandatory-options launch path separate from the public optional API surface.
- `@remotion/cli`: Keeps CLI-only concerns such as argv parsing, warnings, exit behavior, restart handling, and live input-props / env-var updates before delegating to the shared startup layer.
- `@remotion/studio-server`: Is now the shared internal owner of Studio startup: config execution, `Config` / `ConfigInternals`, entry-point resolution, env/default resolution, launch-session orchestration, render queue management, and server lifecycle.
- `@remotion/cli/config`: Remains the user-facing compatibility surface for config imports, but delegates to the shared config core in `@remotion/studio-server`.
- `@remotion/bundler`: Accepts explicit Studio asset paths so the bundler does not need to import `@remotion/studio`, avoiding a `studio-server -> studio` dependency cycle.

## Why this structure
- The public API now lives in the package users should import: `@remotion/studio`.
- The shared implementation now lives in `@remotion/studio-server`, which is the package that already owns Studio backend lifecycle and server runtime concerns.
- CLI and programmatic startup now share one implementation for config loading, queue behavior, and session launch.
- Caller-injected Studio asset paths preserve package layering and avoid reintroducing a package cycle.

## Test plan
- `bun test packages/studio-server/src`
- `bun test packages/studio/src/test/start-studio-api.test.ts`
- `bun test packages/cli/src/test/studio-startup.test.ts`
- `bunx turbo run make --filter=@remotion/studio-server --filter=@remotion/studio --filter=@remotion/cli --filter=@remotion/studio-shared`
- `bunx turbo run lint --filter=@remotion/studio-server --filter=@remotion/studio --filter=@remotion/cli`
- `bun run stylecheck` (currently still reports formatting issues in touched `@remotion/studio-server` / `@remotion/cli` files)


samohovets:
@pullfrog

samohovets:
Probably the amount of hassle isn’t worth it. It would require a lot of splitting and moving logic around, unifying things across the codebase, etc. – or a use case that hasn’t even been properly verified! Closing this.

PR #6814, opened 2026-03-05
`@remotion/web-renderer`: Add CSS filter property support

saimfd:
- Adds support for `filter` css property in `web-renderer` for chrome and firefox.

_Note: Webkit does not support filter in `OffScreenCanvas`, will need to find some other way to implement it in webkit._

### DIFF
--- a/packages/web-renderer/src/test/fixtures/filter.tsx
+++ b/packages/web-renderer/src/test/fixtures/filter.tsx
@@ -1,8 +1,8 @@
 import {AbsoluteFill} from 'remotion';
 
-const FilterBox: React.FC<{label: string; filter: string}> = ({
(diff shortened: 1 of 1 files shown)


JonnyBurger:
😍😍

The Safari thing got me curious, MDN states it as supported, but that's actually a bug! Filed
https://github.com/mdn/browser-compat-data/issues/29193

saimfd:
> 😍😍
> 
> The Safari thing got me curious, MDN states it as supported, but that's actually a bug! Filed [mdn/browser-compat-data#29193](https://github.com/mdn/browser-compat-data/issues/29193)

Awesome, I was also confused when I saw Baseline there but didnt get a change to test it in Safari.

PR #6827, opened 2026-03-06
`@remotion/web-renderer`: add render progress estimations

samohovets:
This PR adds progress time estimation support to `renderMediaOnWeb()` and wires it into Studio’s client render UI.

## Why
`renderMediaOnWeb()` progress previously exposed only frame counts. This made client-side render progress less informative than server-side rendering and left the completion timing semantics unclear.

## What changed
- Extended `RenderMediaOnWebProgress` with:
  - `renderEstimatedTime` (ms remaining)
  - `progress` (overall weighted progress from `0` to `1`)
  - `doneIn` (total elapsed ms from render start until frames are fully encoded)
- Kept `renderedFrames` in the web-renderer progress payload, marked as deprecated.
- Updated ETA estimation to start timing at the frame loop (avoids setup-time skew in early estimates).
- Simplified redundant conditional logic in Studio client progress messaging.
- Updated Studio client render progress plumbing to consume `doneIn`.
- Updated docs and browser tests to match the final payload shape.

## Notes
- `doneIn` is intentionally named differently from server `encodedDoneIn`: web rendering does not have the same separate post-render stitch stage semantics as server rendering.

## Test plan
- `cd packages/web-renderer && bunx vitest src/test/render-media.test.tsx src/test/frame-range.test.tsx --browser --run`
- `cd packages/studio && bun test src/test/client-render-progress.test.ts`

### DIFF
--- a/packages/studio/src/components/RenderQueue/client-render-progress.ts
+++ b/packages/studio/src/components/RenderQueue/client-render-progress.ts
@@ -29,7 +29,7 @@ export const getClientRenderProgressMessage = (
 	if (progress.renderEstimatedTime > 0) {
 		const etaString = `, time remaining: ${formatEtaString(progress.renderEstimatedTime)}`;
 
-		return `Rendering ${Math.round(progress.progress * 100)}%${etaString}`;
+		return `Rendering ${progress.encodedFrames}/${progress.totalFrames}${etaString}`;
 	}
 
 	return `Encoded ${progress.encodedFrames}/${progress.totalFrames}`;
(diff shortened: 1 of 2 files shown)


samohovets:
@pullfrog

samohovets:
@pullfrog

samohovets:
Also had a thought to remove it completely, but

I think many people are already adopting the web renderer in their apps (including myself 😄 ). 

Could we remove it in the first stable web-renderer release instead? I'm concerned about a lot of apps breaking

PR #6845, opened 2026-03-16
feat: indicate development vs production render in client-side render modal

### LINKED ISSUE #6169: Client side rendering modal should indicate if this is considered a production render
Image

If on localhost, this is considered a development render and will not be charged, this should be indicated in the modal

dturakhia:
## Summary

Adds a visual indicator in the client-side render modal (`WebRenderModal`) to show whether the current render is a **development render** (localhost, not charged) or a **production render** (deployed host, billed).

## Changes

- Added `isLocalhostRender()` helper that checks `window.location.hostname` for `localhost` / `127.0.0.1`
- Added `RenderEnvironmentBadge` component displayed in the modal header bar alongside the existing experimental warning badge
- On **localhost**: shows a green badge `Development render · Not charged`
- On **production hosts**: shows an orange/yellow badge `Production render · Usage will be billed`
- Styles match the existing Studio design system (dark background, same font-family/size pattern as `WebRendererExperimentalBadge`)

## Before / After

**Before:** The modal had no indication of whether the render would be charged.

**After:** A colored badge appears in the info bar below the render controls:
- 🟢 `Development render · Not charged` (localhost)
- 🟡 `Production render · Usage will be billed` (deployed)

Fixes #6169

### DIFF
--- a/packages/studio/src/components/RenderModal/WebRenderModal.tsx
+++ b/packages/studio/src/components/RenderModal/WebRenderModal.tsx
@@ -1,5 +1,6 @@
 import type {LogLevel} from '@remotion/renderer';
 import {getDefaultOutLocation} from '@remotion/studio-shared';
+import {WARNING_COLOR} from '../../helpers/colors';
 import type {
 	RenderStillOnWebImageFormat,
 	WebRendererAudioCodec,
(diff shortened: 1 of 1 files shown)


JonnyBurger:
Thanks for the contribution\! We appreciate you taking the time to work on this. However, we'll need to handle this one ourselves since it needs to be integrated with our backend. Closing for now, but thanks again for your interest in improving Remotion\! 🙏

PR #6887, opened 2026-03-20
`@remotion/studio`: Consolidate SSE connections to fix multi-tab hanging

### LINKED ISSUE #6863: Studio: 2 tabs hit Chrome's 6-connection-per-origin limit
## Problem

When opening 2 Studio tabs on `localhost`, the server appears to hang. The event loop shows as idle — the issue is Chrome's HTTP/1.1 limit of **6 concurrent connections per origin**.

Each tab opens **2 long-lived SSE connections**:
1. `EventSource('/events')` — live events (client-id.tsx)
2. `EventSource('/__webpack_hmr')` — webpack hot middleware (hot-middleware-client/client.ts)

With 2 tabs, that's **4 persistent connections**, leaving only 2 slots for API calls, assets, and chunk loading. Any burst of concurrent requests queues behind the limit.

**Confirmed**: using the network URL (e.g. `192.168.x.x:3123`) instead of `localhost` for the second tab works around the issue, because Chrome treats them as different origins.

## Proposed fix

Merge the two SSE streams into a single `EventSource('/events')` connection per tab.

- **Server**: Have the HMR middleware publish through `liveEventsServer.sendEventToClient()` instead of maintaining its own SSE client list. Add a new `EventSourceEvent` variant: `{type: 'hmr', payload: HotMiddlewareMessage}`.
- **Client**: In `client-id.tsx`, dispatch `hmr` events to the HMR handler. In `hot-middleware-client/client.ts`, subscribe to the shared EventSource instead of opening a new one.
- The HMR "send latest stats on connect" behavior would move into the live-events router's connect handler (similar to how initial undo/redo state is sent on connect).
- The HMR heartbeat (10s interval) can be adopted by the shared stream.

This halves the long-lived connections (2 tabs = 2 SSE instead of 4), leaving 4 slots for regular requests.

## Complexity

Medium-low. The HMR middleware's own client management (`createEventStream`) gets removed. Its webpack hooks and `publishStats` logic stay. Main work is wiring publish through the shared event system and subscribing the HMR client to the shared EventSource.

bskang-v6x:
## Summary

- Route webpack HMR events through the existing `/events` SSE connection instead of a separate `/__webpack_hmr` endpoint
- Reduces per-tab SSE connections from 2 to 1, preventing Chrome's HTTP/1.1 6-connection-per-origin limit from being hit with 3+ Studio tabs open
- Net reduction of ~150 lines by removing the standalone HMR event stream infrastructure

Closes #6863

## Test plan

- [x] `cd packages/example && bun run dev` — start Studio
- [x] Open 3+ tabs on `localhost:3000` — all should load without hanging
- [x] Edit a source file → verify HMR updates appear in all tabs
- [x] Check browser DevTools Network tab → confirm only 1 SSE connection (`/events`) per tab, no `/__webpack_hmr`

### DIFF
--- a/packages/studio/src/hot-middleware-client/client.ts
+++ b/packages/studio/src/hot-middleware-client/client.ts
@@ -128,4 +128,23 @@ export const enableHotMiddleware = () => {
 	window.__remotion_processHmrEvent = (hmrEvent: HotMiddlewareMessage) => {
 		processMessage(hmrEvent);
 	};
+
+	// Create a standalone SSE listener for HMR events immediately.
+	// This is needed because lazy-compiled modules require HMR updates
+	// to deliver compiled code, but the React-managed /events SSE
+	// (in PreviewServerConnection) only connects after React mounts —
+	// which itself depends on lazy modules loading first.
+	if (typeof window !== 'undefined' && typeof EventSource !== 'undefined') {
+		const source = new EventSource('/events');
+		source.addEventListener('message', (event) => {
+			try {
+				const parsed = JSON.parse(event.data);
+				if (parsed.type === 'hmr') {
+					processMessage(parsed.hmrEvent);
+				}
+			} catch {
+				// Ignore parse errors
+			}
+		});
+	}
 };


JonnyBurger:
Thanks, will look at it after https://github.com/remotion-dev/remotion/pull/6826 is merged

bskang-v6x:
Sounds good, I'll rebase onto main once #6826 is merged. Let me know if you'd like any changes in the meantime.

bskang-v6x:
@JonnyBurger 

conflicts resolved. but ci failed, I'm going to check soon.

JonnyBurger:
Thank you!
Had to let Claude run over it to actually make it work, but other than that, nice!

PR #6899, opened 2026-03-23
`create-video`: Respect npm registry config for corporate proxy support

### LINKED ISSUE #6888: create-video hardcodes registry.npmjs.org and fails behind corporate npm proxies
# Bug Report 🐛

`npx create-video@latest` fails during project creation in an environment where:

- `npm` is configured to use a corporate npm proxy / mirror (for example JFrog Artifactory)
- direct access to `https://registry.npmjs.org` is blocked or intercepted by the corporate network / proxy

The surprising part is that regular npm operations still work:

- `npm view remotion version --json` succeeds
- `npm install` for Remotion packages succeeds

So the failure seems specific to `create-video`, not to Remotion installation in general.

## Reproduction

1. Configure npm to use a corporate registry / npm proxy in `~/.npmrc`
2. Make sure direct access to `https://registry.npmjs.org` is blocked or rewritten by the network
3. Run:

```bash
npx create-video@latest
```

4. Choose `Blank`
5. Accept the default project name

## Actual behavior

Project creation crashes with:

```txt
<anonymous_script>:1
<!-- ... -->
^

SyntaxError: Unexpected token '<', "<!-- ... " is not valid JSON
    at JSON.parse (<anonymous>)
    at getLatestRemotionVersion (.../node_modules/create-video/dist/latest-remotion-version.js:29:17)
```

In my case, the response body appears to be HTML / proxy output instead of npm metadata JSON.

## Expected behavior

`create-video` should work in environments where npm itself is configured correctly to use a registry mirror/proxy.

At minimum, it should:

- respect npm registry configuration instead of hardcoding `registry.npmjs.org`, or
- fail with a clearer error message if the latest version lookup cannot return valid JSON

## Why I think this is a bug in `create-video`

The current source appears to hardcode the public npm registry:

`packages/create-video/src/latest-remotion-version.ts`

```ts
const req = http.get(
  'https://registry.npmjs.org/remotion',
  {
    headers: {
      accept:
        'application/vnd.npm.install-v1+json; q=1.0, application/json; q=0.8, */*',
    },
  },
  ...
);
```

That means `create-video` bypasses:

- `npm config get registry`
- `NPM_CONFIG_REGISTRY`
- corporate npm mirrors / proxies

So even when `npm view remotion version` works, `create-video` can still fail.

## Additional validation

In the same environment:

```bash
npm view remotion version --json
npm view @remotion/cli version --json
```

both succeed, and a plain `npm install` for Remotion dependencies succeeds as well.

This suggests the issue is the direct HTTP call in `create-video`, not the package availability.

## Suggested fixes

Any of these would help:

1. Resolve the latest version via npm config / npm registry instead of hardcoding `https://registry.npmjs.org`
2. Allow a CLI flag or env var to override the Remotion version lookup source
3. Gracefully handle non-JSON responses and print a useful error message such as:

```txt
Failed to fetch the latest Remotion version from the registry.
Your network may block direct access to registry.npmjs.org.
Please configure a reachable registry or pass an explicit version.
```

4. Optionally allow passing an explicit Remotion version so project scaffolding can continue offline / behind restricted networks

## Environment

- macOS
- Node.js: `v22.19.0`
- npm: `10.9.3`
- `create-video`: latest at the time of testing
- Network: corporate environment with proxy / traffic inspection and an npm registry mirror

## Workaround

As a workaround, I can clone the template repo directly and then run `npm install`, which works:

```bash
git clone https://github.com/remotion-dev/template-empty.git my-video
cd my-video
npm install
```

So this looks fixable in `create-video` specifically.

bskang-v6x:
## Summary

- Read the user's configured npm registry via `npm config get registry` instead of hardcoding `registry.npmjs.org`
- Falls back to the default public registry if the config command fails
- Handles both `http` and `https` registry URLs
- Adds a clear error message when the registry returns non-JSON responses (e.g. corporate proxy HTML pages)

Closes #6888

## Test plan

- [ ] `npm config set registry https://registry.npmjs.org` → `npx create-video` works as before
- [ ] `npm config set registry http://localhost:4873` (Verdaccio) → uses the local registry
- [ ] With an unreachable registry → shows the new descriptive error message instead of `SyntaxError: Unexpected token`

### DIFF
--- a/packages/create-video/src/latest-remotion-version.ts
+++ b/packages/create-video/src/latest-remotion-version.ts
@@ -1,9 +1,35 @@
-import http from 'https';
+import {execSync} from 'node:child_process';
+import http from 'node:http';
+import https from 'node:https';
+
+const DEFAULT_REGISTRY = 'https://registry.npmjs.org';
+
+const getRegistry = (): string => {
+	try {
+		const registry = execSync('npm config get registry', {
+			encoding: 'utf-8',
+			timeout: 10_000,
+			stdio: ['pipe', 'pipe', 'pipe'],
+		}).trim();
+
+		if (registry && registry !== 'undefined') {
+			return registry.replace(/\/$/, '');
+		}
+	} catch {
+		// Fall through to default
+	}
+
+	return DEFAULT_REGISTRY;
+};
 
 const getPackageJsonForRemotion = (): Promise<string> => {
+	const registry = getRegistry();
+	const url = `${registry}/remotion`;
+	const client = url.startsWith('https') ? https : http;
+
 	return new Promise<string>((resolve, reject) => {
-		const req = http.get(
-			'https://registry.npmjs.org/remotion',
+		const req = client.get(
+			url,
 			{
 				headers: {
 					accept:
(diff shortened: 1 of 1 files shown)


JonnyBurger:
Thank you!

PR #6910, opened 2026-03-26
fix: handle InputDisposedError in useWindowedAudioData to ensure prop…

### LINKED ISSUE #6894: `InputDisposedError: Input has been disposed` in `useWindowedAudioData`
```
use-windowed-audio-data.js:88 Uncaught (in promise) InputDisposedError: Input has been disposed.
    at Reader11.requestSlice (mediabunny.cjs:18636:1)
    at Mp4InputFormat._getMajorBrand (mediabunny.cjs:17075:1)
    at Mp4InputFormat._canReadInput (mediabunny.cjs:17093:1)
    at mediabunny.cjs:18513:1
    at async Input.computeDuration (mediabunny.cjs:18544:1)
    at async use-windowed-audio-data.js:52:1
```

We should just be a bit more tolerant

akashcorex:
When the src changes, or the component unmounts, the metadata AbortController disposes the mediabunny Input while computeDuration() may still be running. That correctly throws InputDisposedError, but the catch path called cancelRender(err), which rethrows and produced an unhandled rejection because fetchMetadata is not awaited.

Treat InputDisposedError like the waveform path and useMaxMediaDuration: call continueRender(handle) and return; keep cancelRender for real failures.

Fixes #6894.

### DIFF
--- a/packages/media-utils/src/use-windowed-audio-data.ts
+++ b/packages/media-utils/src/use-windowed-audio-data.ts
@@ -155,6 +155,11 @@ export const useWindowedAudioData = ({
 
 				continueRender(handle);
 			} catch (err) {
+				if (err instanceof InputDisposedError) {
+					continueRender(handle);
+					return;
+				}
+
 				cancelRender(err);
 			} finally {
 				signal.removeEventListener('abort', cont);


JonnyBurger:
Thanks!

PR #6918, opened 2026-03-27
`@remotion/google-fonts`: Fix ignoreTooManyRequestsWarning option

ytakahashi:
## Summary

The `ignoreTooManyRequestsWarning` option in `@remotion/google-fonts` does not actually suppress the console warning, even when explicitly set to `true`. 
I updated the warning condition inside `packages/google-fonts/src/base.ts` to properly respect `options?.ignoreTooManyRequestsWarning` so the warning can actually be silenced when the user explicitly configures it.

### DIFF
--- a/packages/google-fonts/src/base.ts
+++ b/packages/google-fonts/src/base.ts
@@ -198,7 +198,7 @@ export const loadFonts = (
 			}
 		}
 
-		if (fontsLoaded > 20) {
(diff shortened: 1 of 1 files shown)


ytakahashi:
Hi! One of the tests failed but it looks like this failure is unrelated to my change.
Please let me know if I need to do anything else!

JonnyBurger:
Thank you for the fix! 🎉

PR #6933, opened 2026-03-30
`@remotion/studio`: Fix JSON editor sync during undo/redo

tiwariaayu:
Removed conditional update logic and ensured the JSON editor always reflects the latest value state, keeping visual and JSON views in sync.


JonnyBurger:
Thanks a lot!

PR #6934, opened 2026-03-30
`@remotion/studio`: Better loop indicator

tiwariaayu:
looped videos show repeating thumbnails across the entire bar and that icons ♻️ still correctly signify the loop points.


JonnyBurger:
Thanks! Can you address Pullfrog comments?

JonnyBurger:
image

Looks like this, still 2 layers, so it actually does not address what the original issue asked for.

Issue #6704, opened 2026-02-25
Can we provide a better error recovery in the studio?


JonnyBurger:
Let's say we are in NewVideo.tsx

We add `trimBefore={-20}` to the `<Video>` tag

We get a fullscreen error overlay.

When we fix the error in the code, then the Studio reloads.

I wonder if we can do better by calling onUncaughtError? 

https://react.dev/reference/react-dom/client/createRoot

PR #6937, opened 2026-03-31
better error recovery in the studio #6704

tiwariaayu:


JonnyBurger:
Does not solve the reload issue as described in the original issue.

PR #6984, opened 2026-04-04
Skills: clarify defaultOutName must omit file extension

JonnyBurger:
## Summary

The agent skill for `calculateMetadata` showed `defaultOutName` without an extension in the code sample, but the prose and return-value list did not explain that the renderer appends the codec extension (see `getDefaultOutLocation` in `@remotion/studio-shared`). Including `.mp4` or `.mov` in `defaultOutName` produces paths like `Caption-First.mov.mov`.

## Changes

- Document that the file extension must be omitted and why.
- Clarify the `defaultOutName` bullet in the return value section.


JonnyBurger:
Closing: upstream skill example already omitted the extension; the drift was in the vendored copy in the consuming repo, not a gap in remotion-dev/skills.

PR #6995, opened 2026-04-04
`@remotion/studio`: Scroll selected asset and composition into view in sidebar

JonnyBurger:
When navigating to an asset or composition URL, scroll the sidebar tree so the selected item is centered in the viewport without animation.

- `scrollIntoView({ block: 'center', behavior: 'auto', })` in `useLayoutEffect` on selected asset and composition rows.
- On `popstate` for asset URLs, call `selectAsset` so folders expand and the Assets tab is shown.
- Memoize `useSelectAsset` with `useCallback`.

### DIFF
--- a/packages/studio/src/components/AssetSelectorItem.tsx
+++ b/packages/studio/src/components/AssetSelectorItem.tsx
@@ -1,4 +1,11 @@
-import React, {useCallback, useContext, useMemo, useRef, useState} from 'react';
+import React, {
+	useCallback,
+	useContext,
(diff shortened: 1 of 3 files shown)


JonnyBurger:
Closing in favor of #6997, which includes scroll-into-view **only when not fully visible** (`sidebar-scroll-into-view.ts`), asset metadata errors in the canvas, and the `InitialCompositionLoader` URL/sync changes from this PR (merged in 368f1890fd).

PR #7010, opened 2026-04-06
`@remotion/renderer`: Fix unsafe optional chaining with array access

bhaktofmahakal:
## Summary

Fixed three critical runtime safety bugs in `@remotion/renderer` that could cause crashes in production:

1. **puppeteer-evaluate.ts** - Fixed unsafe array access after optional chaining on exception descriptions
2. **BrowserPage.ts** - Fixed unsafe array indexing in CDP log level and tag parsing  
3. **Connection.ts** - Added try-catch error handling around JSON.parse to prevent crashes from malformed CDP messages

### Issue 1: puppeteer-evaluate.ts (lines 144-158)
The pattern `exceptionDetails.exception?.description?.split('\n')[0] as string` would crash if `description` is undefined, because split returns undefined and then accessing `[0]` throws.

**Fix**: Extract description with null coalescing, split once, then safely access array elements.

```typescript
const description = exceptionDetails.exception?.description ?? '';
const descriptionLines = description.split('\n');
const err = new SymbolicateableError({
  stack: description,
  name: className,
  message: descriptionLines[0] ?? '',
  frame,
  stackFrame: parseStack(descriptionLines),
  chunk: null,
});

### DIFF
--- a/packages/renderer/src/browser/BrowserPage.ts
+++ b/packages/renderer/src/browser/BrowserPage.ts
@@ -105,15 +105,12 @@ const format = (
 
 	for (const a of args) {
 		if (a.type === 'symbol' && a.description?.includes(`__remotion_level_`)) {
-			const levelParts = a.description?.split('__remotion_level_');
-			const levelValue = levelParts?.[1]?.replace(')', '');
-			if (levelValue) {
-				logLevelFromRemotionLog = levelValue as LogLevel;
-			}
+			logLevelFromRemotionLog = a.description
+				?.split('__remotion_level_')?.[1]
+				?.replace(')', '') as LogLevel;
 		}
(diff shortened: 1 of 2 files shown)


JonnyBurger:
Did you submit this PR because you faced a bug?
What happened?

bhaktofmahakal:
> Did you submit this PR because you faced a bug? What happened?

No, I didn't encounter this in production. This came from a systematic code review looking for potential runtime safety issues in the codebase.

bhaktofmahakal:
After review feedback from @pullfrog (thank you!), I've updated the PR. Pls review again

JonnyBurger:
Thanks!
We don't wanna refactor just for the sake of it, it also brings risk. If there was no problem we don't need to address anything for now

PR #7049, opened 2026-04-10
`remotion`: Truncate very long data URLs to avoid polluting logs

acj:
We've been working with large images as `` sources in our compositions and have had >10MB base64'd data URLs appear in our logs when something goes wrong (usually from memory pressure), which makes debugging difficult. This truncates those blobs.

The threshold length I chose is fairly arbitrary, so I'm happy to change it to something else. My working assumption is that showing the first part of the data URL is enough to help track down which image is being shown.

Example:
```
 Tab 0, delayRender()  "Loading <Img> with src=data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAADwAAAAhwCAYAAA...[truncated 10635270 chars]" handle was cleared after 236ms
```

### DIFF
--- a/packages/core/src/Img.tsx
+++ b/packages/core/src/Img.tsx
@@ -23,6 +23,16 @@ function exponentialBackoff(errorCount: number): number {
 	return 1000 * 2 ** (errorCount - 1);
 }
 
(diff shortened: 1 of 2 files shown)


acj:
Rebased to hopefully fix the browser test failure.

JonnyBurger:
Thanks a lot! 🙌

PR #7059, opened 2026-04-11
fix(media-parser): bump stream-remote test timeout 15s → 30s

AleksejDix:
## Summary

The "Should stream" test in `packages/media-parser/src/test/stream-remote.test.ts` fetches a 1MB MP4 from `test-videos.co.uk` and parses it, with a hardcoded 15-second timeout. This is tight for a network-bound test against a third-party CDN with variable latency.

Bumping to 30 seconds gives generous headroom for the 1MB download + parse without making the test slower in the happy path (the test still exits as soon as parsing finishes).

## Observed flake

In #7058's "Browser tests" job, this test failed at **exactly 15002ms** — 2ms over the deadline. The PR it was running against was a 4-line CSS-only change in `@remotion/promo-pages`, which has zero causal pathway to a media-parser network test. Classic flake signature: hits the deadline by single-digit milliseconds.

```
1 tests failed:
✗ Should stream [15002.05ms]
   ^ this test timed out after 15000ms.
```

## Diff

```diff
-	{timeout: 15000},
+	// Bumped from 15000 — third-party CDN (test-videos.co.uk) has
+	// intermittent slow responses on CI runners; the test was
+	// timing out at exactly 15002ms in a known-flake pattern.
+	// 30s is generous headroom for the 1MB download + parse.
+	{timeout: 30000},
```

5+/1−. Pure timeout bump. No assertion changes, no test logic changes.

## Why not a different fix

Considered alternatives:

| Alternative | Why not |
|---|---|
| Use a local fixture instead of a remote CDN | The point of `stream-remote.test.ts` is to test the streaming-from-URL code path. Switching to local would be coverage loss. |
| Mock the fetch | Defeats the integration-test purpose of the file. |
| Switch the CDN URL to a Remotion-hosted asset | Real fix long-term, but a larger refactor and out of scope here. Worth a follow-up. |
| Re-run the failed job and hope | Doesn't fix the underlying tightness; flake will recur. |

Bumping the timeout is the smallest, scope-isolated fix that eliminates this specific flake.

## Long-term

The test would benefit from being moved off the third-party CDN entirely. A follow-up that hosts the 1MB Big Buck Bunny clip on a Remotion-controlled URL (or as a local fixture for the streaming case) would fully eliminate the network dependency. That's a separate, larger PR.


JonnyBurger:
@pullfrog delete the entire test instead

PR #7063, opened 2026-04-13
feat(renderer): add optional bitsPerMb for ProRes encoding

### LINKED ISSUE #7031: @remotion/renderer: Optional FFmpeg bits_per_mb for ProRes (prores_ks)
## Summary

Remotion currently maps video-bitrate style control to FFmpeg `-b:v` for most codecs, but **explicitly skips that for ProRes** with a warning (`"ProRes does not support videoBitrate. Ignoring."` in `validateQualitySettings`).

That matches common FFmpeg behavior: **`-b:v` is not a meaningful knob for Apple ProRes**. However, FFmpeg’s **`prores_ks`** (Kostya) encoder—the one Remotion uses for software ProRes—does expose **`bits_per_mb`** (`ffmpeg -h encoder=prores_ks`), which **does** influence output size on real clips.

This issue proposes **optional support** for passing **`bits_per_mb`** (and possibly related options like `alpha_bits` / `quant_mat`) through `renderMedia`, CLI, and config, defaulting to current behavior when unset.

## Background

- **Encoder:** `packages/renderer/src/get-codec-name.ts` selects `prores_ks` for software ProRes.
- **Quality path:** `packages/renderer/src/crf.ts` → `validateQualitySettings` drops `videoBitrate` when `codec === 'prores'`.
- **Docs:** `packages/docs/docs/encoding.mdx` documents ProRes profiles; there is no mention of `bits_per_mb`.

## Observations (local testbed, synthetic source)

**Setup:** `lavfi` `testsrc`, **1280×720**, **30 fps**, **2 s**, no audio; encoder **`prores_ks`**.

### Encoder minimum

On the FFmpeg build used for testing, **`bits_per_mb` below 128** fails initialization (`too few bits per MB, please set at least 128`). So any API should either **clamp** or **validate** to ≥ 128 (or defer to FFmpeg defaults when omitted).

### ProRes 422 HQ (`-profile:v 3`, `yuv422p10le`)

| Finding | Value |
|--------|--------|
| Smallest output | **2 507 384 B** at **`bits_per_mb = 128`** |
| Plateau size (matches `8192`) | **5 364 279 B** |
| Smallest `bits_per_mb` with **byte-identical** file to `8192` | **`1434`** |

So for this easy clip, any value **≥ ~1434** is equivalent to **8192** (identical bytes to `bits_per_mb = 8192`).

### ProRes 4444 + alpha (`-profile:v 4`, `yuva444p10le`)

| Finding | Value |
|--------|--------|
| Smallest output | **2 444 541 B** at **`bits_per_mb = 128`** |
| Plateau size (matches `8192`) | **6 908 088 B** |
| Smallest `bits_per_mb` with **byte-identical** file to `8192` | **`1861`** |

4444 shows **more distinct sizes** in the sweep (harder / alpha) but the same pattern: a **finite budget** then a **hard plateau**; **`8192` is not special** beyond being above the plateau for this content.

### General takeaways

1. **`bits_per_mb` is a real size/quality lever** for `prores_ks`, unlike `-b:v`.
2. **Plateau is content-dependent**; harder sources will likely need higher `bits_per_mb` before saturating, but values **above the plateau are wasted** (and identical to each other for a given frame sequence).
3. The legacy FFmpeg encoder **`prores`** (non-ks) exposes **no** `bits_per_mb`; **`prores_ks`-only** feature.
4. **Hardware** `prores_videotoolbox` (when used) may need a **separate code path** or explicit “unsupported” if it has no equivalent.

## Suggested implementation sketch

- New optional field, e.g. `proResBitsPerMacroblock?: number | null` (name bikeshedding welcome), plumbed like `proResProfile`.
- When set, append FFmpeg args **`-bits_per_mb`** for `prores_ks` only; document interaction with profile and pixel format.
- Validate range **128–8192** (per current `prores_ks` AVOption) or pass through and surface FFmpeg errors.
- **Mutually exclusive** with `videoBitrate` for ProRes remains reasonable; alternatively document that only `bits_per_mb` applies.
- Docs: encoding / transparent-videos / `renderMedia` reference.

## References

- FFmpeg: `ffmpeg -h encoder=prores_ks` — `-bits_per_mb`, `-alpha_bits`, `-quant_mat`, `-mbs_per_slice`, `-vendor`
- Apple nominal profile bitrates vs this knob are **orthogonal**: profile sets tier; `bits_per_mb` caps budget within the encoder’s model for that tier.

renezander030:
## Summary

- Adds optional `bitsPerMb` parameter to `renderMedia()` and `stitchFramesToVideo()` that maps to FFmpeg's `-bits_per_mb` flag for the `prores_ks` encoder
- Validates range 128-8192 (matching `prores_ks` encoder limits) and ensures it's only used with the `prores` codec
- Plumbed through all encoding paths: `generateFfmpegArgs`, `prespawnFfmpeg`, `internalStitchFramesToVideo`, and `internalRenderMedia`

## Test plan

- [ ] Render a ProRes video with `bitsPerMb: 128` and verify smaller output than default
- [ ] Render a ProRes video with `bitsPerMb: 8192` and verify output matches default quality
- [ ] Verify `bitsPerMb` with non-ProRes codec throws a clear error
- [ ] Verify `bitsPerMb` below 128 or above 8192 throws a clear error
- [ ] Verify omitting `bitsPerMb` preserves existing default behavior

Fixes #7031

### DIFF
--- a/packages/renderer/src/ffmpeg-args.ts
+++ b/packages/renderer/src/ffmpeg-args.ts
@@ -20,6 +20,7 @@ const firstEncodingStepOnly = ({
 	encodingMaxRate,
 	encodingBufferSize,
 	hardwareAcceleration,
+	bitsPerMb,
 }: {
 	hasPreencoded: boolean;
 	proResProfileName: string | null;
@@ -31,6 +32,7 @@ const firstEncodingStepOnly = ({
 	encodingMaxRate: string | null;
 	encodingBufferSize: string | null;
 	hardwareAcceleration: HardwareAccelerationOption;
+	bitsPerMb: number | undefined;
 }): string[][] => {
 	if (hasPreencoded || codec === 'gif') {
 		return [];
@@ -55,6 +57,7 @@ const firstEncodingStepOnly = ({
 			encodingBufferSize,
 			hardwareAcceleration,
 		}),
+		bitsPerMb !== undefined ? ['-bits_per_mb', String(bitsPerMb)] : null,
 	].filter(truthy);
 };
 
@@ -72,6 +75,7 @@ export const generateFfmpegArgs = ({
 	hardwareAcceleration,
 	indent,
 	logLevel,
+	bitsPerMb,
 }: {
 	hasPreencoded: boolean;
 	proResProfileName: string | null;
@@ -86,6 +90,7 @@ export const generateFfmpegArgs = ({
 	hardwareAcceleration: HardwareAccelerationOption;
 	indent: boolean;
 	logLevel: LogLevel;
+	bitsPerMb?: number | undefined;
 }): string[][] => {
 	const encoderSettings = getCodecName({
 		codec,
@@ -164,6 +169,7 @@ export const generateFfmpegArgs = ({
 			encodingBufferSize,
 			x264Preset,
 			hardwareAcceleration,
+			bitsPerMb,
 		}),
 	].filter(truthy);
 };
--- a/packages/renderer/src/prespawn-ffmpeg.ts
+++ b/packages/renderer/src/prespawn-ffmpeg.ts
@@ -58,6 +58,7 @@ type PreStitcherOptions = {
 	colorSpace: ColorSpace | null;
 	binariesDirectory: string | null;
 	hardwareAcceleration: HardwareAccelerationOption;
+	bitsPerMb?: number | undefined;
 };
 
 export const prespawnFfmpeg = (options: PreStitcherOptions) => {
@@ -112,6 +113,7 @@ export const prespawnFfmpeg = (options: PreStitcherOptions) => {
 			hardwareAcceleration: options.hardwareAcceleration,
 			indent: options.indent,
 			logLevel: options.logLevel,
+			bitsPerMb: options.bitsPerMb,
 		}),
 
 		'-y',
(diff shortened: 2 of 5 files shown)


renezander030:
The remaining CI failure (`Build Node 16 on ubuntu-latest`) is a flaky file-watcher test unrelated to this PR:

```
✗ existenceOnly emits created with empty content when file appears [5000.05ms]
  ^ this test timed out after 5000ms
```

This test lives in `packages/studio-server/src/test/file-watcher.test.ts` and is not touched by this PR. It passes locally (57ms) and the same test also fails on recent `main` branch runs (e.g. runs from Apr 11). The timeout appears to be a race condition with `fs.watch` on slower CI runners.

JonnyBurger (changes requested):
This is missing adding the option in a of places, because you made it optional using `?`. Please remove the optionality and fix all the errors that occur because of it.

Also all the API docs need this option to be added, in the Remotion Studio the option needs to be exposed as well. This should also affect Lambda, Cloud Run, Vercel and all foreign language SDKs.

This would affect 100+ files probably, just as a heads-up if you are having second thoughts about implementing this!

See https://github.com/remotion-dev/remotion/blob/main/.claude/skills/add-cli-option/SKILL.md

renezander030:
Hey @JonnyBurger — fair point, and thanks for the heads-up on scope. Before I put in the work to plumb `bitsPerMb` through Studio, Lambda, Cloud Run, Vercel, and the foreign-language SDKs following `add-cli-option/SKILL.md`, do you still want this contribution from the outside, or have you already taken it on internally? Happy to commit to the full fan-out (non-optional, all surfaces, docs) if it's still useful — just don't want to duplicate work on a closed PR.

PR #7074, opened 2026-04-18
`@remotion/web-renderer`: Always preserve silent audio frames to match composition duration

### LINKED ISSUE #7073: web-renderer: leading/trailing silence is trimmed in mp3 output, breaking downstream timing
### Summary
When exporting audio-only mp3 via `renderMediaOnWeb()`, silence at the start and end of the composition is stripped from the resulting file. The output duration equals only the audible segment duration, not the composition duration.

### Why this is a real problem
The mp3 is fed into transcription (ASR) to get word-level timestamps that are then aligned back to the composition timeline. Because the leading silence is removed, every returned timestamp is shifted by the amount of stripped silence, so captions / subtitles / cue markers render at the wrong frame. Any pipeline that relies on `output mp3 time === composition time` breaks.

### Expected
`output mp3 duration === composition duration`. Silence at head/tail of the composition (frames with no `<Audio>` active) should be preserved as silent samples so that `t=0` in the mp3 maps to frame 0 of the composition.

### Current workaround
Placing a silent `<Audio>` that spans the entire composition (full `durationInFrames`, starting at frame 0) forces silence to be retained in the output and makes timing line up correctly. It works, but it's a hack — users shouldn't need to fabricate a dummy silent track for timing preservation.

### Request
Please expose this as a parameter on `renderMediaOnWeb()` so it can be controlled explicitly — e.g. an `enforceAudioTrack` / `preserveSilence` option, matching what `renderMedia()` (Node) already has via `enforceAudioTrack`. Current web-renderer audio options (`muted`, `audioCodec`, `audioBitrate`, `sampleRate`) don't cover this.

yamin-H:
### Summary
This PR adds the `preserveSilence` boolean option to `renderMediaOnWeb()`. When enabled, it ensures that silent segments at the head and tail of a composition (frames with no active `<Audio>` components) are preserved in the output file instead of being trimmed.

This matches the existing `enforceAudioTrack` functionality found in the Node.js `renderMedia()` API.

Closes #7073

### Problem
When exporting audio (especially for MP3) via `renderMediaOnWeb()`, silence at the start and end is currently stripped. This causes timing inaccuracies when feeding the output into transcription engines (ASR), as timestamped results no longer align with the original Remotion timeline.

### Solution
Updated the logic in `onlyInlineAudio` to return silent samples instead of `null` when the `preserveSilence` flag is active. This ensures the encoder receives consistent data for the full duration of the composition.

### Verification (Proof)
I've added a regression test in `packages/web-renderer/src/test/audio.test.tsx` that verifies a 5s composition produces a 5s output regardless of silent head/tail segments.

##Before fix (Silence trimmed to ~3s) 
Screenshot 2026-04-18 094456

##After fix (Silence preserved to 5s) 
Screenshot 2026-04-18 095400

**Test Output:**
```text
✓ chromium src/test/audio.test.tsx > output duration should match composition duration
Expected: 5s, Actual: 5.013333333333334s

### DIFF
--- a/packages/web-renderer/src/audio.ts
+++ b/packages/web-renderer/src/audio.ts
@@ -3,6 +3,10 @@ import type {TRenderAsset} from 'remotion';
 const TARGET_NUMBER_OF_CHANNELS = 2;
 
 function mixAudio(waves: Int16Array[], length: number) {
+	if (waves.length === 0) {
+		return new Int16Array(length);
+	}
+
 	if (waves.length === 1 && waves[0].length === length) {
 		return waves[0];
 	}
@@ -35,11 +39,8 @@ export const onlyInlineAudio = ({
 	fps: number;
 	timestamp: number;
 	sampleRate: number;
-}): AudioData | null => {
+}): AudioData => {
 	const inlineAudio = assets.filter((asset) => asset.type === 'inline-audio');
-	if (inlineAudio.length === 0) {
-		return null;
-	}
 
 	const expectedLength = Math.round(
 		(TARGET_NUMBER_OF_CHANNELS * sampleRate) / fps,


JonnyBurger:
Thanks! We should make this the default and only supported behavior actually.

PR #7150, opened 2026-04-25
`@remotion/example`: Add html-in-canvas demo composition

JonnyBurger:
## Summary

- Adds a new `html-in-canvas` composition under `packages/example/src/HtmlInCanvas/` that demonstrates how to use the WICG [html-in-canvas](https://github.com/WICG/html-in-canvas) `drawElementImage()` / `layoutSubtree` / `requestPaint()` APIs directly inside a Remotion composition, separately from the integrated path in `@remotion/web-renderer`.
- The composition renders an animated HTML scene (gradient background, big title, frame counter, pulsing circles) as the immediate child of a `layoutSubtree` canvas, captures it each frame via `drawElementImage()`, and applies a simple vertical sine-wave column-shift effect to a second output canvas.
- Uses `canvas.requestPaint()` + the `paint` event when available (Chrome Canary) for proper sync, and falls back to a double-`requestAnimationFrame` on regular Chrome where `requestPaint()` is not yet shipped.
- Shows a friendly fallback message instructing the user to enable `chrome://flags/#canvas-draw-element` when the API is unavailable.

## Test plan

- [ ] Open the Remotion Studio in `packages/example` (`bun run dev`) in Chrome with `chrome://flags/#canvas-draw-element` enabled and confirm the `html-in-canvas` composition shows the animated scene fading into the wave-distorted output.
- [ ] Verify the fallback message appears in browsers without the flag.
- [ ] Verify Studio scrubbing/playback updates the captured frame correctly.


JonnyBurger:
Replacing with a clean branch off main; see new PR.

PR #7157, opened 2026-04-26
`remotion`: Add canvas effects runtime, `<Solid>`, and `<Experimental.HtmlInCanvas>`

JonnyBurger:
## Summary

Adds a per-pixel canvas effect chain runtime to `remotion` core, plus two source components that consume it:

- **`<Solid>`** — flat-color procedural source.
- **`<Experimental.HtmlInCanvas>`** — rasterizes DOM children using the [WICG `html-in-canvas`](https://github.com/WICG/html-in-canvas) proposal (Chrome Canary).

The runtime mixes 2D, WebGL2, and (future) WebGPU effect passes in a single chain by ping-ponging between scratch canvases and bridging backend transitions with `createImageBitmap`. Effect authors declare effects with `defineEffect()`; the chain enforces a premultiplied-alpha, sRGB color contract end-to-end.

## What's in this PR

- `packages/core/src/canvas-effects/` — runtime types, `defineEffect`, `createDescriptor`, `CanvasPool`, `getGpuDevice`, `useEffectChain`, `<Solid>`, `<HtmlInCanvas>`.
- Re-exports from `packages/core/src/index.ts`: `defineEffect`, `createDescriptor`, `Solid`, `Backend`, `EffectDefinition`, `EffectDescriptor`, `EffectsProp`, `EffectApplyParams`. `HtmlInCanvas` is exposed under `Experimental.HtmlInCanvas`.
- 2D-only Bun + happy-dom unit tests in `packages/core/src/test/`.

This PR adds only the core runtime. Sample effects (`wave`, `tint`, `blur`), the `@remotion/canvas-effects` package, and example/docs updates ship in a follow-up.

## Test plan

- [x] `bun run build`
- [x] `bun run stylecheck`
- [x] `bun test` from `packages/core`
- [ ] Manual: validate `<Solid>` with effects in studio (covered in follow-up PR).

Made with [Cursor](https://cursor.com)

PR #7159, opened 2026-04-26
`@remotion/canvas-effects`: Add wave, tint, and blur effects

JonnyBurger:
## Summary

Builds on the canvas-effects runtime added to `remotion` core (#7157) by shipping a new package and the supporting example/docs updates.

- New package **`@remotion/canvas-effects`** with three ready-to-use effects:
  - `wave({amplitude, wavelength, speed, sliceWidth, background})` — 2D vertical-slice sine wave (ports the example's old `applyWaveEffect`).
  - `tint({color, amount})` — 2D flat-color tint that respects source alpha.
  - `blur({radius})` — separable Gaussian blur on WebGL2; returns a tuple of horizontal + vertical passes that the chain runtime flattens automatically.
- Migrates the example HTML-in-Canvas composition to consume `Experimental.HtmlInCanvas` from `remotion` and `wave` from `@remotion/canvas-effects`; deletes the now-redundant example-local files (`apply-wave-effect.ts`, `html-in-canvas.tsx`).
- Adds a new `solid-effects` composition exercising a cross-backend (WebGL2 + 2D) chain on a `<Solid>` source.
- Adds docs pages under `canvas-effects/` plus a new sidebar entry.

## Note on stacking

This branch is also the head of #7158, which targets the stacked branch `canvas-effects-core` (= the runtime PR #7157). This PR exists so reviewers can see the diff against `main` directly. Only one of the two should be merged once #7157 lands.

## Tradeoffs

- Sample effects intentionally live in their own package so the core stays small and to keep WebGL2/WebGPU specifics out of `remotion`.
- The first iteration of `blur` uses a fixed 9-tap kernel; sigma scales with `radius`. Larger radii will alias before they smooth — acceptable for a v1 effect but worth revisiting.
- Cross-backend transitions go through `createImageBitmap`. This is GPU-resident on supporting browsers but copies on others; effect chains that frequently switch backends will be slower on weaker platforms.

## Test plan

- [x] `bun run build`
- [x] `bun run stylecheck`
- [x] `bun test` from `packages/core`
- [ ] Manual smoke test: render `solid-effects` (no Chrome flag required) and `html-in-canvas` (Chrome Canary with `chrome://flags/#canvas-draw-element` enabled) compositions in studio.


JonnyBurger:
Closing in favor of the stacked pair #7157 + #7158.

PR #7160, opened 2026-04-26
remotion: Ramp up gain on AudioContext resume to avoid click/pop

### LINKED ISSUE #7140: Resuming an AudioContext should use a ramp up
# Avoiding Audio Artifacts with Gain Ramping in Web Audio API

When you abruptly start a sine wave at full volume, you get a **click or pop** because of the sudden discontinuity in the waveform. The solution is to ramp the gain from 0 to your target volume over a short period.

## Using `GainNode` with Linear/Exponential Ramp

```javascript
const ctx = new AudioContext();

// Create oscillator
const oscillator = ctx.createOscillator();
oscillator.type = 'sine';
oscillator.frequency.value = 440;

// Create gain node for envelope control
const gainNode = ctx.createGain();
gainNode.gain.setValueAtTime(0, ctx.currentTime); // start silent

// Ramp up over ~20ms (short enough to sound instant, long enough to avoid clicks)
gainNode.gain.linearRampToValueAtTime(1, ctx.currentTime + 0.02);

// Connect: oscillator → gain → output
oscillator.connect(gainNode);
gainNode.connect(ctx.destination);

oscillator.start();
```

## Ramp Down Before Stopping (equally important!)

Stopping abruptly causes the same artifact. Ramp down first:

```javascript
function stopOscillator(oscillator, gainNode, ctx) {
  const now = ctx.currentTime;
  gainNode.gain.setValueAtTime(gainNode.gain.value, now);
  gainNode.gain.linearRampToValueAtTime(0, now + 0.02);
  oscillator.stop(now + 0.02); // stop after ramp completes
}
```

## Key Methods on `AudioParam`

| Method | Behavior |
|---|---|
| `setValueAtTime(value, time)` | Sets value at exact time (use as anchor) |
| `linearRampToValueAtTime(value, time)` | Linear ramp to value by given time |
| `exponentialRampToValueAtTime(value, time)` | Exponential ramp (value must be > 0) |
| `setTargetAtTime(target, startTime, timeConstant)` | Asymptotic approach to target |

## Tips

* **~10–30 ms** is the sweet spot for ramps — imperceptible to the ear but eliminates clicks
* Always call `setValueAtTime()` **before** a ramp to anchor the starting point, otherwise the ramp may behave unexpectedly
* `exponentialRampToValueAtTime` cannot ramp to or from **0** — use a very small value like `0.0001` instead
* For more natural-sounding envelopes, `setTargetAtTime` gives you an exponential decay curve controlled by a time constant

Mothilal-M:
Closes #7140.

## Summary

Resuming an `AudioContext` after a user-initiated pause currently produces an audible click/pop, because audio output starts abruptly at full volume the moment the context transitions back to `running`.

This PR introduces a master `GainNode` between every audio source and `audioContext.destination`, and ramps that gain from `0` → `1` over 20 ms whenever the context is resumed — short enough to be imperceptible, long enough to eliminate the discontinuity.

## Changes

**New helper**
- `packages/core/src/audio/audio-ramp.ts` — `resumeAudioContextWithRamp(audioContext, masterGain)` schedules `setValueAtTime(0)` + `linearRampToValueAtTime(1, +0.02s)` on the master gain, then calls `audioContext.resume()`.

**Routing**
- `useSingletonAudioContext` now also creates a `masterGain` node and connects it to `destination`. Returned as `{audioContext, masterGain}`.
- `SharedAudioContext` (core) and `SharedAudioContextForMediaPlayer` (media) expose `masterGain`.
- `useAmplification` (core) and `audioIteratorManager` (media) connect each per-source `GainNode` to `masterGain` instead of `audioContext.destination`.

**Resume call sites**
- `playAllAudios()` in `shared-audio-tags.tsx` and `play()` in `@remotion/player`'s `use-player.ts` now use `resumeAudioContextWithRamp(...)` instead of `audioContext.resume()`.

**Tests**
- Updated 4 test files in `@remotion/media` that construct `SharedAudioContextForMediaPlayer` to include the new `masterGain` field.

## Test plan

- [ ] `bun run build` passes
- [ ] `bun run stylecheck` passes
- [ ] `bun test` in `packages/core` and `packages/media` pass
- [ ] Manual: in `packages/player-example`, pause and resume playback of a composition with audio — verify the click/pop is gone and audio still sounds at the user's set volume
- [ ] Manual: verify per-track volume control (`volume` prop on `<Audio>` / `<Video>`) still works (per-source gain is preserved; only master is ramped)

## Notes

- The master gain stays at `1` outside of resume ramps, so existing per-source volume behavior is unchanged.
- 20 ms is the documented sweet spot in the issue; not exposed as a tunable to keep surface area minimal.
- Ramp-down on `suspend()` is a logical follow-up but intentionally out of scope — the issue title focuses on resume, and pause-time clicks are less common in practice.

### DIFF
--- a/packages/core/src/audio/audio-ramp.ts
+++ b/packages/core/src/audio/audio-ramp.ts
@@ -1,0 +1,14 @@
+// Short enough to be imperceptible to the ear, long enough to avoid a click.
+// See https://github.com/remotion-dev/remotion/issues/7140.
+const RAMP_DURATION_SECONDS = 0.02;
+
+export const resumeAudioContextWithRamp = (
+	audioContext: AudioContext,
+	masterGain: GainNode,
+): Promise<void> => {
+	const now = audioContext.currentTime;
+	masterGain.gain.cancelScheduledValues(now);
+	masterGain.gain.setValueAtTime(0, now);
+	masterGain.gain.linearRampToValueAtTime(1, now + RAMP_DURATION_SECONDS);
+	return audioContext.resume();
+};
--- a/packages/core/src/audio/shared-audio-tags.tsx
+++ b/packages/core/src/audio/shared-audio-tags.tsx
@@ -12,6 +12,7 @@ import {useLogLevel, useMountTime} from '../log-level-context.js';
 import {Log} from '../log.js';
 import {playAndHandleNotAllowedError} from '../play-and-handle-not-allowed-error.js';
 import {useRemotionEnvironment} from '../use-remotion-environment.js';
+import {resumeAudioContextWithRamp} from './audio-ramp.js';
 import type {SharedElementSourceNode} from './shared-element-source-node.js';
 import {makeSharedElementSourceNode} from './shared-element-source-node.js';
 import {useSingletonAudioContext} from './use-audio-context.js';
@@ -80,6 +81,7 @@ type SharedContext = {
 	playAllAudios: () => void;
 	numberOfAudioTags: number;
 	audioContext: AudioContext | null;
+	masterGain: GainNode | null;
 	audioSyncAnchor: {value: number};
 	scheduleAudioNode: (
 		options: ScheduleAudioNodeOptions,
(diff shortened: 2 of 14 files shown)


JonnyBurger:
Thank you, but this is already in #7143 :)

PR #7164, opened 2026-04-26
`@remotion/renderer`: Fix `everyNthFrame` to always include the first frame

saismrutiranjan18:
### What

Fixes `everyNthFrame` frame selection to offset from the start of the frame range instead of using absolute frame numbers. This ensures the first frame is always included.

### Why

`getFramesToRender` filtered frames using `frame % everyNthFrame === 0`, which meant:
- `frameRange: [5, 5]` with `everyNthFrame: 2` produced an **empty frame list** (since `5 % 2 !== 0`), leading to `totalFramesToRender = 0` and `NaN` progress.
- Starting at frame 0, `everyNthFrame: 2` would select frames `0, 2, 4` — but starting at frame 1, it would select `2, 4, 6` instead of `1, 3, 5`.

### How

Changed the filter in `getFramesToRender` from:

```ts
.filter((index) => index % everyNthFrame === 0)
```

to:

```ts
.filter((frame) => (frame - frameRange[0]) % everyNthFrame === 0)
```

This offsets from the start of the range, so the first frame is always selected, and every Nth frame after that.

Also reverted the `totalFramesToRender === 0` NaN guard from the original PR since the root cause is now fixed.

### Files changed

- `packages/renderer/src/get-duration-from-frame-range.ts` — fix frame selection logic
- `packages/renderer/src/render-media.ts` — revert NaN guard (no longer needed)
- `packages/renderer/src/test/get-frames-to-render.test.ts` — 9 unit tests for the fix
- `packages/docs/docs/render-as-gif.mdx` — update docs with correct frame sequences

### Checklist
- [x] Bug fix (non-breaking change)
- [x] No new dependencies added
- [x] Existing tests still pass
- [x] New unit tests added
- [x] Docs updated

### DIFF
--- a/packages/renderer/src/render-media.ts
+++ b/packages/renderer/src/render-media.ts
@@ -505,20 +505,24 @@ const internalRenderMediaRaw = ({
 		`Rendering frames ${realFrameRange.join('-')}`,
 	);
 
-	const callUpdate = () => {
-		const encoded = Math.round(0.8 * encodedFrames + 0.2 * muxedFrames);
-		onProgress?.({
-			encodedDoneIn,
-			encodedFrames: encoded,
-			renderedDoneIn,
-			renderedFrames,
-			renderEstimatedTime,
-			stitchStage,
-			progress:
-				Math.round((70 * renderedFrames + 30 * encoded) / totalFramesToRender) /
-				100,
-		});
-	};
+
+const callUpdate = () => {
+	const encoded = Math.round(0.8 * encodedFrames + 0.2 * muxedFrames);
+	onProgress?.({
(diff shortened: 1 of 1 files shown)


JonnyBurger:
Superseded by #7173, which fixes the root cause instead of guarding against NaN.

JonnyBurger:
Thank you for the contribution and for identifying this issue! Instead of guarding against NaN, we decided to fix the root cause — `everyNthFrame` now offsets from the start of the frame range so the first frame is always included. Since we couldn't push to your fork, we opened #7173 with the fix. Thanks again!

Issue #7183, opened 2026-04-27
Player: useElementSize returns wrong height under 3D-rotated / non-uniform-scale ancestors


DavidWiener84:
## Summary

`@remotion/player`'s internal `useElementSize` hook (`packages/player/src/utils/use-element-size.ts`) measures the Player's outer container with `getClientRects()` (the post-transform AABB), then tries to undo the parent CSS transform by dividing both width AND height by ONE scalar derived from the X-axis ratio:

```ts
const probableCssParentScale =
  contentRect.width === 0 ? 1 : newSize[0].width / contentRect.width;

const width  = options.shouldApplyCssTransforms || probableCssParentScale === 0
  ? newSize[0].width
  : newSize[0].width  * (1 / probableCssParentScale);

const height = options.shouldApplyCssTransforms || probableCssParentScale === 0
  ? newSize[0].height
  : newSize[0].height * (1 / probableCssParentScale);
```

This is correct for uniform 2D scale (`transform: scale(N)`) where `scaleX === scaleY`. **It produces a wrong height under any non-uniform CSS transform** — `scale(X, Y)` with different factors, `rotateX` / `rotateY` / `rotate3d`, `perspective`, `matrix3d` with non-trivial Z, or any combination — because the X-axis and Y-axis AABB grow at different rates.

The Player then computes a wrong `canvasSize`, feeds it into `calculateCanvasTransformation`, and emits a non-zero `centerY` that letterboxes the composition vertically.

## Reproduction

```tsx
import { Player } from "@remotion/player";

function App() {
  return (
    <div style={{
      width: 320,
      height: 180,
      transform: "rotateY(45deg) rotateX(-15deg)",
      transformStyle: "preserve-3d",
    }}>
      <Player
        component={MyComp}
        compositionWidth={1920}
        compositionHeight={1080}
        durationInFrames={300}
        fps={30}
        style={{ width: "100%", height: "100%" }}
      />
    </div>
  );
}
```

**Expected:** the composition fills the 320×180 (16:9) container; no letterbox; `centerY === 0`.

**Actual:** `useElementSize` reports `height ≈ 320` (square) instead of `180`. `calculateCanvasTransformation` emits a positive `centerY`, which `calculateOuter` writes as `top` on the composition outer. The composition is rendered at the right scale but **vertically offset** inside its container — empty top portion of the plate, composition bottom spilling below the plate.

## Numerical trace from a real reproduction

```
[...]

PR #7185, opened 2026-04-27
@remotion/player: Read offsetWidth/offsetHeight in useElementSize (fixes all 3 measurement paths)

DavidWiener84:
Fixes `useElementSize` so all three of its measurement paths return the element's CSS layout box, robust against all ancestor CSS transforms.

Refs #7183 (full reproduction, numerical trace, and analysis). This is the **broader** of two PRs offered for that issue — the **minimal-diff** alternative is #7184, which fixes only the `ResizeObserver` path with a per-axis scale compensation. Maintainers can pick whichever shape they prefer; the two PRs are mutually exclusive.

## Background

`useElementSize` reads its target with `getClientRects()` in three places:

1. The initial `useState` initializer.
2. The `ResizeObserver` callback.
3. The `updateSize` callback (window-resize + `updateAllElementsSizes`).

`getClientRects()` returns the **post-transform AABB**, which is not what consumers want — they want the layout box. Today, only path (2) tries to recover the layout box, by dividing the AABB by `newSize.width / contentRect.width` (the X-axis ratio) **on both axes**. That recovery is correct under uniform 2D scale (`transform: scale(N)`) but wrong under any non-uniform transform (`scale(X, Y)`, `rotateX/Y`, `rotate3d`, `perspective`, `matrix3d`) because the X- and Y-axis AABB grow at different rates. Paths (1) and (3) don't compensate at all — they return the raw AABB, which is wrong under every transform.

## What this PR does

Introduces a `measureElement(el, shouldApplyCssTransforms)` helper that returns `{width, height, left, top}`:

- `width` / `height` come from `el.offsetWidth` / `el.offsetHeight` — the element's pre-transform content + padding box, exactly what the existing recovery was trying to compute. Reading from the layout engine directly is robust against all ancestor transforms (uniform 2D scale, asymmetric 2D scale, 3D rotation, perspective, `matrix3d`) without depending on `contentRect.width / height` reliability.
- `left` / `top` continue to come from `getClientRects()[0]` because consumers want **viewport-relative** coordinates, and `offsetLeft` / `offsetTop` are relative to the offset parent (not the viewport).
- When the caller passes `shouldApplyCssTransforms: true` (used by `PlayerSeekBar` for the transformed scrubber), the helper returns `getClientRects()` width/height unchanged — preserving the post-transform behaviour for callers that explicitly want it.

All three callsites use the helper. Net change: `+68 / -42` lines, all in `packages/player/src/utils/use-element-size.ts`.

## Behaviour matrix

| Parent transform | Old code | This PR |
|---|---|---|
| None | ✅ correct | ✅ correct |
| `transform: scale(2)` (uniform 2D) | ✅ correct in `ResizeObserver`, ❌ wrong in initial `useState` and `updateSize` | ✅ correct everywhere |
| `transform: scaleX(2) scaleY(3)` | ❌ wrong height | ✅ correct |
| `transform: rotateY(45deg)` + `transform-style: preserve-3d` | ❌ wrong height | ✅ correct |
| `perspective: 1000px` + `transform: rotateX(15deg)` | ❌ wrong height | ✅ correct |
| `shouldApplyCssTransforms: true` (callers wanting AABB) | ✅ AABB returned | ✅ AABB returned (unchanged) |

## Why this approach

The PR description for the minimal-diff alternative (#7184) preserves the existing "compensate via division" intent and only fixes the per-axis bug in the `ResizeObserver` callback. That PR is correct but has two limitations:

1. It leaves paths (1) and (3) returning the raw AABB. Initial measurements before the first `ResizeObserver` callback fires, and any `updateSize` triggered by a window-resize, will still be wrong under any transform.
2. It depends on `contentRect.height` being reliable — the existing inline comment in `useElementSize` claims it is not (`The contentRect returns the width without any \`scale()\`'s being applied. The height is wrong`), and the PR effectively decides the comment is outdated. We've found `contentRect` to be reliable on modern Chrome / Firefox / Safari, but if there's a known-bad browser, this PR (which never reads `contentRect`) sidesteps the question entirely.

This PR removes the recovery round-trip and gets the layout box directly from the property the layout engine populates for exactly this purpose.

## Test plan

`useElementSize` is a hook depending on `ResizeObserver` and DOM measurement; the `bun:test` suite under `packages/player/src/test/` is currently pure-utility-only. I'd be happy to add a `jsdom` or Playwright regression that asserts:

1. `transform: scale(2)` parent → correct width AND height (this PR passes; current `ResizeObserver` passes; current `useState`/`updateSize` fails).
2. `transform: scaleX(2) scaleY(3)` parent → correct height (current code fails everywhere; this PR passes).
3. `transform: rotateY(45deg)` + `transform-style: preserve-3d` parent → correct height (current code fails everywhere; this PR passes).
4. Callers passing `shouldApplyCssTransforms: true` get post-transform AABB unchanged.

Let me know if you have a preferred test harness for hooks like this.

### DIFF
--- a/packages/player-example/pages/use-element-size-testbed.tsx
+++ b/packages/player-example/pages/use-element-size-testbed.tsx
@@ -1,0 +1,139 @@
+import {Player} from '@remotion/player';
+import React from 'react';
+import {UseElementSizeTestbedComposition} from '../src/UseElementSizeTestbedComposition';
+
+/**
+ * Manual testbed for GitHub issue #7183 and PRs #7184 / #7185 (useElementSize
+ * under transformed ancestors).
+ *
+ * Run from repo root: `bunx turbo run dev --filter=@remotion/player-example`
+ * Open: http://localhost:3000/use-element-size-testbed
+ *
+ * Compare before/after: checkout `main` vs the PR branch and reload. After the
+ * fix, the 3D plate should stay visually filled (TOP/BOTTOM bands flush with the
+ * plate); before, the composition shifts vertically inside the plate.
+ */
+const plate: React.CSSProperties = {
+	width: 320,
+	height: 180,
+	border: '2px solid #666',
+	boxSizing: 'border-box',
+	backgroundColor: '#222',
+};
+
+const playerCommon = {
+	component: UseElementSizeTestbedComposition,
+	compositionWidth: 1920,
+	compositionHeight: 1080,
+	durationInFrames: 300,
+	fps: 30,
+	acknowledgeRemotionLicense: true as const,
+	controls: true,
+	style: {width: '100%', height: '100%'} as const,
+};
+
+const UseElementSizeTestbedPage: React.FC = () => {
+	return (
+		<div
+			style={{
+				padding: 24,
+				fontFamily: 'system-ui, sans-serif',
+				maxWidth: 900,
+				lineHeight: 1.5,
+			}}
+		>
+			<h1 style={{marginTop: 0}}>useElementSize / #7183 testbed</h1>
+			<p>
+				Reproduction from{' '}
+				<a href="https://github.com/remotion-dev/remotion/issues/7183">
+					issue #7183
+				</a>
+				: Player inside a 320×180 plate with 3D rotation. Before the fix,
+				<code> useElementSize </code>
+				can report an inflated height (post-transform AABB), which drives a
+				non-zero <code>centerY</code> and vertical letterboxing. Red/green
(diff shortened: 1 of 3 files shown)


JonnyBurger:
I'm sorry, I cannot merge it because I don't understand exactly what's going on here.

There is too much going on here for me to comprehend, and previous logic for handling the Player being nested in a div that has a scale applied to it seems to have been removed. Too big of a risk that I want to merge it.

PR #7197, opened 2026-04-28
Fix scale issue with filter precomposite

saimfd:


packages/web-renderer/src/test/fixtures/issue-7050-transform-scale.tsx
```diff
@@ -0,0 +1,295 @@
...
+					touchAction: 'none',
+					border: 'none',
+					boxShadow: 'none',
+					zIndex: 10,
+					borderRadius: 4,
+					fontSize: 21,
+					overflowWrap: 'break-word',
+					wordBreak: 'break-word',
+					lineHeight: 0.9,
+					transform: 'translate3d(0, 0, 0) scale(0.67)',
```
saimfd, line 27: Its the same component but with scale defined on parent level

PR #7224, opened 2026-05-01
`@remotion/vercel`: Fallback URL for glibc download

### LINKED ISSUE #7221: createSandbox() fails on Vercel: patchCompositor times out connecting to launchpadlibrarian.net
createSandbox() fails at Vercel build time because patchCompositor attempts to download libc6_2.35 from launchpadlibrarian.net via curl, but that host is unreachable from the Vercel sandbox. The same build succeeded 2 days ago on the same package version.

```
Error: Failed to patch compositor: curl: (28) Failed to connect...
    at patchCompositor (...)
    at async createSandbox (...)
```

Steps to reproduce:

1. Use v4.0.452 with createSandbox() in a Vercel build step
2. Deploy to Vercel

Expected behavior:

createSandbox() completes successfully and produces a snapshot, as it did a few days ago on the same codebase.

Additional context:

- The glibc download URL is hardcoded in index.mjs as a launchpadlibrarian.net URL
- A previous successful build 2 days prior used the same package version with no issues

Suggestion:

- Adding a retry with an alternative URL?
- Mirroring the glibc .deb on Remotion-controlled infrastructure?

JonnyBurger:
## Summary
- Add a fallback URL (`remotion.media`) for the libc6 deb download in the Vercel Sandbox compositor patch, used if the launchpadlibrarian fetch fails.

## Test plan
- [ ] Verify compositor patching still works in Vercel Sandbox when the primary URL is reachable.
- [ ] Verify fallback is used when the primary URL fails.

### DIFF
--- a/packages/vercel/src/internals/patch-compositor.ts
+++ b/packages/vercel/src/internals/patch-compositor.ts
@@ -3,6 +3,7 @@ import type {Sandbox} from '@vercel/sandbox';
 const GLIBC_DIR = '/tmp/glibc235';
 const LIBC6_DEB_URL =
 	'https://launchpadlibrarian.net/612471225/libc6_2.35-0ubuntu3.1_amd64.deb';
+const FALLBACK_URL = 'https://remotion.media/libc6_2.35-0ubuntu3.1_amd64.deb';
 
 /**
  * Remotion does not officially support glibc 2.34, but it can be patched.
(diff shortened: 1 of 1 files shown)


umairnadeem:
@JonnyBurger thanks for the quick fix! any chance of a quick patch release? fix isn't in 4.0.455 and we're blocked on launchpad from vercel sandbox. thank you!

PR #7273, opened 2026-05-05
Player: fix seek timeline updates and timeupdate after seek

### LINKED ISSUE #7205: seekTo + timeupdate event
Seems like when calling seekTo on the PlayerRef, the currentTimestamp is not updated, seeing neither frameupdate nor timeupdate bting called. Do you think this PR should fix that?

JonnyBurger:
## Summary

Fixes [GitHub issue #7205](https://github.com/remotion-dev/remotion/issues/7205) where `frameupdate` / `timeupdate` did not fire reliably after `seekTo()` (notably when paused in Firefox) and where `timeupdate` often did not fire in Chrome after seeks during playback.

## Root cause

1. **`seek()` skipped updating timeline state** when `useVideo()` had no `id` yet. Playback continues to advance frames under `config.id` (`player-comp`), but `seek()` only called `setTimelinePosition` inside `if (video?.id)`. That left React timeline position stale while `frameRef` still updated—so hooks keyed off `frame` never ran and no `frameupdate` / `timeupdate` fired.

2. **`timeupdate` throttling reset on every frame**: The previous effect cleared its `setTimeout` on each `frame` dependency change. While frames advance quickly (playback), the timer was constantly cancelled, so `timeupdate` often never emitted until motion stopped—matching “stale timestamp while playing / after seek” reports.

## Changes

- **`packages/player/src/use-player.ts`**: Always derive `compositionId` as `video?.id ?? config?.id ?? PLAYER_COMP_ID` and update timeline state on every seek (same key as `Player` initial state and playback `setFrame`).
- **`packages/player/src/use-playback.ts`**: Throttle `timeupdate` to ~250ms without cancelling a pending trailing dispatch when more frames arrive; the trailing dispatch uses the latest frame from a ref.

## Testing

- `bun run makeplayer`
- `cd packages/player && bun test src`


JonnyBurger:
Not confident in it

PR #7277, opened 2026-05-05
`remotion`: support `preservesPitch` prop

### LINKED ISSUE #7070: `preservesPitch`-support for React `Player` component
# Feature Request 🛍️

I'm using the `@remotion/player` package to render a video (with audio) as a react component on our website.
I implemented a simple playback speed picker, so users can speed up or slow down the video.
I've noticed that the sound changes pitch when changing the speed. 
By default, browsers don't change the pitch (see https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/preservesPitch).

## Use Case

Add a flag to enforce/disable pitch preservation!

## Github help-bot response

Based on the Remotion documentation, there's no built-in prop in the `@remotion/player` component to control `preservesPitch` on the underlying HTML5 audio/video elements. 

The Player's `playbackRate` prop (between -4 and 4) controls playback speed, but doesn't expose the underlying `preservesPitch` attribute of the browser's media elements. 

The `toneFrequency` prop on `<Html5Audio />` allows pitch control, but the documentation explicitly states: **"Pitch correction is currently only applied during rendering."** This means it won't affect the Player in the browser. 

Unfortunately, this means you cannot currently enforce/disable pitch preservation in the Remotion Player through documented APIs. The behavior depends on the browser's default (`preservesPitch: true` by default).

**Workarounds to consider:**
1. Open a feature request on [GitHub](https://github.com/remotion-dev/remotion/issues) to add `preservesPitch` support
2. If using audio from an `<OffthreadVideo />`, the same limitation applies

For now, the pitch-changing behavior you're experiencing is expected since browsers default to preserving pitch when changing playback rate.

Sources:
https://www.remotion.dev/docs/player/player/
https://www.remotion.dev/docs/audio/pitch

mvanhorn:
## Summary

`<Player>` and `<Thumbnail>` now accept a `preservesPitch?: boolean` prop that controls whether audio pitch is preserved when `playbackRate` is changed.

Default is `true`, matching `HTMLMediaElement.preservesPitch`'s browser default. Existing consumers see no behavior change.

## Why this matters

[#7070](https://github.com/remotion-dev/remotion/issues/7070) (dvschuyl) describes a playback speed picker built on `<Player>` where audio keeps its original pitch when sped up or slowed down. Browsers default to pitch preservation, but `<Player>` never sets the property, so users can't opt out for tape-style "chipmunk" or "deep voice" effects. [@jtheis85](https://github.com/remotion-dev/remotion/issues/7070#issuecomment-3559247824) confirmed the same ask in the Remotion Discord.

`HTMLMediaElement.preservesPitch` is a [standard browser property](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/preservesPitch) supported in every modern browser. Surfacing it as a Player prop is the minimal change.

## Implementation

The prop flows through the existing `PlaybackRateContext` (extended with `preservesPitch` and `setPreservesPitch`) so both Player-managed and Studio-managed compositions get the same behavior. `use-media-playback.ts` applies it to the media element in the same `useLayoutEffect` that sets `playbackRate`, on the same dependency array, which keeps the assignments in lock-step.

## Testing

Validator unit tests cover:

- `preservesPitch={true}`, `{false}`, and undefined all render
- non-boolean values throw with the same shape as `playbackRate` validation

`HTMLMediaElement.preservesPitch` is a browser-native audio effect and does not have a visual representation, so no screenshot/video evidence applies. The existing `<Thumbnail>` mock and the test-only `wrap-sequence-context` helper were updated to satisfy the new context type.

Fixes #7070


mvanhorn:
Updated in e01085f to match the peer-validator convention from `Player.tsx:270-327`: `TypeError` and `must be a boolean or undefined but got 'X' instead`. Test regex updated.

Kept the two `useEffect` hooks split per your "feel free to keep the split" note. One-prop-per-effect reads cleaner here since each dependency array documents one prop's flow into local state.

Re: the non-blocking docs observations, agreed. `<Thumbnail>` doesn't expose the prop and the `@remotion/media` WebCodecs path doesn't honor it. Will tighten the PR description to clarify this applies only to native `HTMLMediaElement` playback (`<Audio>`/`<Video>` via `useMediaPlayback`).

JonnyBurger (changes requested):
Please re-implement this for `preservePitch` to be a prop for OffthreadVideo, Html5Audio, Html5Video, not for the Player. It doesn't belong there.

mvanhorn:
Re-implemented per your direction (force-pushed 29002570d7, single clean commit on top of base):

- `preservesPitch?: boolean` is now a per-component prop on `OffthreadVideo`, `Html5Video`/`Video`, and `Html5Audio`/`Audio`. It threads through to `HTMLMediaElement.preservesPitch` via the same effect that already syncs `playbackRate`, so when the prop is omitted the browser default applies and nothing is written.
- The Player-level prop, the `PlaybackRateContext` field, the player-level validator, and `Thumbnail.tsx`'s default are all reverted - Player, Thumbnail, TimelineContext, and the docs/api.mdx entry are back to their pre-PR shape.
- New per-component validator added in `validate-media-props.ts` matching the existing peer-validator message format (`'preservesPitch' must be a boolean or undefined but got 'X' instead`).
- Tests cover default, true, false, and invalid input on each component plus the runtime sync to `HTMLMediaElement.preservesPitch`. All 64 tests in @remotion/core pass locally.
- Docs moved from `player/api.mdx` to `html5-audio.mdx`, `html5-video.mdx`, and `offthreadvideo.mdx`.

packages/core/src/validate-media-props.ts
```diff
@@ -1,7 +1,11 @@
 import type {VolumeProp} from './volume-prop.js';
 
 export const validateMediaProps = (
-	props: {volume: VolumeProp | undefined; playbackRate: number | undefined},
+	props: {
+		volume: VolumeProp | undefined;
+		playbackRate: number | undefined;
+		preservesPitch?: boolean | undefined;
```
JonnyBurger: don't mix `?` and `| undefined`, just use `undefined`

packages/docs/docs/html5-audio.mdx
```diff
@@ -122,6 +122,12 @@ export const MyVideo = () => {
 Playing a video in reverse is not supported.
 :::
 
+### `preservesPitch?`<AvailableFrom v="4.0.458" />
+
+Controls whether the audio pitch is preserved when [`playbackRate`](#playbackrate) is not `1`.
+
+Accepts `true`, `false`, or `undefined`. Default: `undefined`, which uses the browser default for [`HTMLMediaElement.preservesPitch`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/preservesPitch).
```
JonnyBurger: Disclose that this does not work during render. `toneFrequency` should be used for that

packages/docs/docs/html5-video.mdx
```diff
@@ -180,6 +180,12 @@ export const MyComposition = () => {
 Playing a video in reverse is not supported.
 :::
 
+### `preservesPitch?`<AvailableFrom v="4.0.458" />
+
+Controls whether the audio pitch is preserved when [`playbackRate`](#playbackrate) is not `1`.
+
+Accepts `true`, `false`, or `undefined`. Default: `undefined`, which uses the browser default for [`HTMLMediaElement.preservesPitch`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/preservesPitch).
```
JonnyBurger: Disclose that this does not work during render. `toneFrequency` should be used for that

packages/docs/docs/offthreadvideo.mdx
```diff
@@ -192,6 +192,12 @@ export const MyComposition = () => {
 Playing a video in reverse is not supported.
 :::
 
+### `preservesPitch?`<AvailableFrom v="4.0.458" />
+
+Controls whether the audio pitch is preserved when [`playbackRate`](#playbackrate) is not `1`.
+
+Accepts `true`, `false`, or `undefined`. Default: `undefined`, which uses the browser default for [`HTMLMediaElement.preservesPitch`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/preservesPitch).
```
JonnyBurger: Disclose that this does not work during render. `toneFrequency` should be used for that

JonnyBurger (changes requested):

mvanhorn:
Thanks @JonnyBurger. Pushed 7b632408:

- Renamed `preservesPitch` to `preservePitch` (matching your spelling) across `Audio`, `Video`, `OffthreadVideo` and their internal `Html5Audio`/`Html5Video` variants. No `preservesPitch`/`preservePitch` references in `packages/player/` (the original PR title was misleading; the prop was already only on the media components).
- `AudioForRendering`/`VideoForRendering` no longer assign `preservesPitch` on the DOM element during render. Docs updated on `html5-audio.mdx`, `html5-video.mdx`, `offthreadvideo.mdx`: `preservePitch` is preview-only; `toneFrequency` remains the render-time control.
- Default is `true`, matching `HTMLMediaElement.preservesPitch`. Validator updated; tests updated.

Verification: my sandbox could not reach the registry to run `pnpm install`, so I am relying on upstream CI to confirm. If you want me to flip any specific behavior (e.g., keep the `undefined`-as-untouched semantic) just let me know.

JonnyBurger:
@mvanhorn Tests are failing and the next version is now 4.0.462.
Otherwise, looks good to me!

mvanhorn:
Rebased onto latest main (force-pushed `feat/7070-player-preserves-pitch` — just the 2 PR commits clean on top, versions now reflect 4.0.462 from the upstream bump).

On the tests: root cause is that `NODE_ENV=test` sets `getRemotionEnvironment().isRendering` true, so the test render path goes through `AudioForRendering.tsx` which strips `preservePitch` (`{preservePitch: _preservePitch, ...nativeProps}`) by design - the layoutEffect that sets `mediaRef.current.preservesPitch` lives in `AudioForPreview` / `useMediaPlayback`, consistent with the "preview-only" scope from the last commit message.

Two options for landing the tests:

1. Wrap each render in a `RemotionEnvironmentContext.Provider` that forces `isRendering: false`, so the preview path runs in tests.
2. Drop the in-test assertion of `ref.current.preservesPitch` and assert at a level higher up (e.g. that `useMediaPlayback` was called with the right prop, or move to a Playwright/integration test).

Happy to push (1) as a follow-up commit if you confirm that's the shape you want - it's a small test-context wrapper and a few rewrites of the asserts. Or just lmk if you'd prefer to land this without the test fixture and revisit it separately.

mvanhorn:
@JonnyBurger thanks for shipping this. Exposing preservesPitch through `<Player>` and `<Thumbnail>` with the HTMLMediaElement-matching default keeps the API ergonomic for the rare case where pitch shift is actually wanted.

mvanhorn:
Shipping the preservesPitch prop unlocks the playbackRate-without-chipmunk path - thanks @JonnyBurger for the merge.

JonnyBurger:
@mvanhorn Awesome, glad this helps that playbackRate use case — thanks for the feedback!

PR #7295, opened 2026-05-08
Add MCP server for Remotion documentation

tabreznakhwa-lgtm:
Adds a project-level `.mcp.json` so Claude Code users working in this repo can search the Remotion documentation via `@remotion/mcp`.

This mirrors the existing `.cursor/mcp.json` config already in the repo — it uses the same `npx @remotion/mcp@latest` command.

### DIFF
--- a/.mcp.json
+++ b/.mcp.json
@@ -1,0 +1,8 @@
+{
+	"mcpServers": {
+		"remotion-documentation": {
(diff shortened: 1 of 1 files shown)


JonnyBurger:
Thank you, but this is recursive - all that info is already in the repo because this is the source! 🙂

PR #7310, opened 2026-05-11
`@remotion/studio`: Persist expansion of tracks in visual mode

JonnyBurger:
TODO: 

- [x] More explicit name for `SequencePropsSubscriptionStatesSetters` -> should be explicitly an overrideId map
- [x] `jsxInMapCallback` can be the same for every `nodePath`
- [x] Make it so `sequenceNodePath` cannot be null
- [x] Watcher for nodePath does not have to be duplicated for every Sequence
- [x] If we have a cached version of `stack -> nodePath`, then use that to reduce flicker
- [ ] Instead of `jsxInMapCallback`, we should just count how many sequences in timeline have the same `nodePath`


JonnyBurger:
@pullfrog re-review

PR #7343, opened 2026-05-13
`remotion`: Fix easing overshoot when it doesn't make sense

### LINKED ISSUE #7340: Clarify and maybe fix `Easing` in interpolate()
Questions I ask myself:

- What is the behavior for Easing when we overshoot? Does it continue linearly or do we extrapolate the Easing as well? 
- When we have multiple keyframes, what happens to the easing? Does it apply inbetween keyframes, or over the full range?

JonnyBurger:
- `Ease.out(Ease.circle)` could lead to NaN if value below 0
- `Ease.bezier()` and `Easing.bounce` now is clamped to [0, 1] - these easings do not make sense in an "overshooting" context

### DIFF
--- a/packages/core/src/bezier.ts
+++ b/packages/core/src/bezier.ts
@@ -146,19 +146,21 @@ export function bezier(
 	}
 
 	return function (x: number): number {
+		const clampedX = Math.min(1, Math.max(0, x));
+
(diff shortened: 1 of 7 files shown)


JonnyBurger:
@pullfrog Fix the linting CI

PR #7347, opened 2026-05-13
`@remotion/web-renderer`: Fix `text-transform` regex to avoid unusual capitalize

saimfd:
## Problem:
The capitalize `text-transform` regex was using `\b\w` (word boundary + word character) which incorrectly capitalized letters after apostrophes. For example, "don't" became "Don'T" instead of "Don't".

## Fix:
- Added a negative lookbehind (?<!') to skip word boundaries preceded by an apostrophe.

## Examples:

| Input         | Before        | After         |
| ------------- | ------------- | ------------- |
| don't         | Don'T         | Don't         |
| it's a test   | It'S A Test   | It's A Test   |


JonnyBurger:
Thank you!

PR #7361, opened 2026-05-13
Allow premounting layout none sequences

alec-watts:
## Summary
- Allow `<Sequence layout=\"none\">` to use `premountFor` / `postmountFor` in preview by removing the premount exclusion.
- Keep layout-none sequences unwrapped during normal playback, while hiding the temporary premount/postmount render with a `display: contents` wrapper.
- Add regression coverage for layout-none premount, normal playback, and postmount behavior.

## Test plan
- `bun test packages/core/src/test/series.test.tsx`
- `cd packages/core && bun run lint` could not start locally because `@remotion/eslint-config-internal/dist/index.js` is missing in this checkout.


JonnyBurger (changes requested):
Premounting works by adding `opacity: 0` to the container.
with `layout="none"` we have no container. How can this work?

alec-watts:
Good point. Premounting needs a host for the hidden state. I think the invariant for layout="none" should be “no layout-affecting container during active playback,” not “no internal DOM host can ever exist.” We can render a stable internal wrapper: display: contents while active, and a hidden out-of-flow container during premount/postmount. That preserves child mount identity and lets media/effects warm up without adding a layout box during normal playback. If Remotion wants layout="none" to mean literally no wrapper DOM ever, then generic premounting is impossible without child opt-in.

JonnyBurger:
The terminology of `layout="none"` and `layout="absolute-fill"` is not entirely right, but how it works:

- `layout="none"` has NO container element and this would be a breaking change
- `layout="absolute-fill"` has a container element, but it can be styled as you like with `style` and `styleWhilePremounted`.

So what you want is actually `layout="absolute-fill"`, and apply your desired styles such as `display: 'contents'`!

PR #7379, opened 2026-05-14
`@remotion/studio`: Save effect props back to source

JonnyBurger:
TODO: 

- [x] (A) Logs are not good
- [x] (B) `remotion__WEBPACK_IMPORTED_MODULE_1__.Internals.useMemoizedEffects is not a function` - circular dependency error? **Fixed** in `c953af5e2e`: `use-memoized-effects` no longer imports `internals`; it imports `OverrideIdsToNodePathsGettersContext` from `sequence-node-path` to break the `↔` cycle.
- [x] (C) Separate contexts for `getCodeValues` and `getEffectCodeValues` to minimize amount of rerenders when one state changes but not the other. **Fixed** in `9f8ee8cd9f`: `VisualModeSequenceCodeValuesContext` + `VisualModeEffectCodeValuesContext` with separate version counters and a shared ref to latest `CodeValues`.
- [x] (D) If the optimistic update needs to be reverted, propagate the error in `TimelineFieldRow.tsx` and `TimelineEffectFieldRow.tsx` using showNotification. **Fixed** in `4957ff2ff7` (notification only; see follow-up commits for `onSave().finally()` in timeline inputs).
- [x] (E) Instead of rejecting with `"no-args-object"`, insert an args object when saving (e.g. `tint()` → `tint({ amount: 0.5 })`). **Fixed** in `5f659db637`: codemod inserts `{}` as first arg; non-object first args use `computed` / `Cannot update effect prop: computed`; `no-args-object` removed; studio-shared test for optimistic update with unset props.
- [x] Test effect that is not inlined - `const t = tint({color: 'red', amount: 10}); effects={[t]}` does not work well
- [x] Test effect that has the arguments not passed
- [x] Updating just `tint()` with no arguments is accepted but is then not saved back


JonnyBurger:
@pullfrog re-review

PR #7493, opened 2026-05-20
`@remotion/media`: Add requestInit prop

### LINKED ISSUE #7093: `@remotion/media`: Add prop to control fetch cache behavior on `<Video>` / `<Audio>`
## Problem

Chrome's HTTP cache has a known bug with partial content (206) responses when CDNs return `Vary` headers (e.g. `Vary: Origin`). When Chrome tries to revalidate cached partial content, it fails and falls back to a regular 200 request. mediabunny then throws because it expects 206 for range requests:

```
HTTP server did not respond with 206 Partial Content to a range request.
```

This affects users serving media from CDNs like CloudFront (Wistia) and Cloudflare R2. The root cause is a Chromium bug: https://issues.chromium.org/issues/436025873

mediabunny addressed this by adding `cache: 'no-store'` to its internal fetch calls ([PR](https://github.com/Vanilagy/mediabunny/pull/236)), but users still report the issue when using the `<Video>` component from `@remotion/media` with certain CDN configurations.

## Current state

`<Video>` and `<Audio>` currently only forward `credentials` to mediabunny's `UrlSource` via `requestInit`:

```ts
new UrlSource(src, credentials ? { requestInit: { credentials } } : undefined)
```

There is no way for users to control the `cache` property (or other `RequestInit` options) of the underlying fetch calls.

## Proposal

Add a prop to `<Video>` and `<Audio>` to control the fetch cache behavior, e.g.:

```tsx
<Video src="https://cdn.example.com/video.mp4" fetchCache="no-store" />
```

Or more generally, expose a `requestInit` prop to allow users to pass through additional fetch options:

```tsx
<Video src="https://cdn.example.com/video.mp4" requestInit={{ cache: 'no-store', credentials: 'include' }} />
```

This would let users affected by the Chrome cache bug opt into `cache: 'no-store'` without waiting for an upstream Chromium fix.

## Relevant files

- `packages/media/src/video/props.ts` — Video prop types
- `packages/media/src/audio/props.ts` — Audio prop types
- `packages/media/src/media-player.ts` — preview path, constructs `UrlSource`
- `packages/media/src/video-extraction/get-frames-since-keyframe.ts` — render path, constructs `UrlSource`

## Related Discord thread

https://discord.com/channels/809501355504959528/1445495265028079718

samohovets:
## Summary

- Closes #7093.
- Adds a `requestInit` prop to `@remotion/media` `<Video>` and `<Audio>` and passes it to mediabunny `UrlSource` fetch requests in preview and render paths.
- Deprecates the older `credentials` prop in favor of `requestInit={{credentials: ...}}`, while keeping it working for backwards compatibility.
- Documents `requestInit`, including `cache: 'no-store'` for CDN/browser cache range-request issues.

## Testing

- `bunx turbo run lint formatting --filter='@remotion/media'`
- `bun run build-docs`
- Previous revision passed `bunx turbo run make --filter='@remotion/media'` and `bun run build`.
- Previous revision ran `bun run stylecheck` and failed only at known `@remotion/lambda-go` Go 1.22.2 vs required Go >= 1.23.0 environment limitation.


packages/docs/docs/media/audio.mdx
```diff
@@ -267,7 +271,28 @@ import {Audio} from '@remotion/media';
...
+
+```tsx twoslash
+import {AbsoluteFill} from 'remotion';
+import {Audio} from '@remotion/media';
+
+// ---cut---
+export const MyComposition = () => {
+  return (
+    <AbsoluteFill>
+      <Audio requestInit={{cache: 'no-store'}} src="https://example.com/audio.mp3" />
```
JonnyBurger: Let's use https://remotion.media/audio.mp3 to make it actually work!

packages/docs/docs/media/video.mdx
```diff
@@ -360,7 +364,28 @@ import {Video} from '@remotion/media';
...
+
+```tsx twoslash
+import {AbsoluteFill} from 'remotion';
+import {Video} from '@remotion/media';
+
+// ---cut---
+export const MyComposition = () => {
+  return (
+    <AbsoluteFill>
+      <Video requestInit={{cache: 'no-store'}} src="https://example.com/video.mp4" />
```
JonnyBurger: Let's use https://remotion.media/audio.mp3 to actually make it work!

packages/media/src/get-sink.ts
```diff
@@ -5,12 +5,41 @@ import {getSinks} from './video-extraction/get-frames-since-keyframe';
...
+	src: string;
+	credentials: RequestCredentials | undefined;
+	requestInit: RequestInit | undefined;
+}) => JSON.stringify([src, credentials, getRequestInitKey(requestInit)]);
+
 export const getSink = (
 	src: string,
 	logLevel: LogLevel,
 	credentials: RequestCredentials | undefined,
+	requestInit?: RequestInit,
```
JonnyBurger: Let's keep the pattern of mandatory objects, always.

It should be `requestInit: RequestInit | undefined`

packages/media/src/media-player.ts
```diff
@@ -132,6 +134,7 @@ export class MediaPlayer {
 		playing: boolean;
 		sequenceOffset: number;
 		credentials: RequestCredentials | undefined;
+		requestInit?: RequestInit;
```
JonnyBurger: Should be mandatory, but undefined may be passed

JonnyBurger (changes requested):
@pullfrog

JonnyBurger:
@samohovets Can you do one last check, do you think this is good to merge now?

samohovets:
LGTM!

Issue #7489, opened 2026-05-20
@remotion/web-renderer: words touch each other because findWords measures whitespace tokens with zero width


aryanranderiya:
## Summary

When a composition renders text through `renderStillOnWeb` or `renderMediaOnWeb`, multi-word strings inside a single text node come out with the inter-word spaces collapsed. "ship it tomorrow?" rasterizes as "shipit tomorrow?". The live Remotion Player and the regular CLI `renderMedia` both render the same composition correctly. The bug is exclusive to the canvas-based capture path in `@remotion/web-renderer`.

Hit this while building [motion-studio](https://github.com/theexperiencecompany/motion-studio) — a chat-demo composition with normal `<span>{text}</span>` bubbles became unusable in browser exports.

Related: tracked under the client-side rendering masterplan #5913.

## Repro

A minimal composition is enough:

```tsx
import { AbsoluteFill } from "remotion";

export const Repro = () => (
  <AbsoluteFill
    style={{
      background: "white",
      fontFamily: "-apple-system, BlinkMacSystemFont, sans-serif",
      fontSize: 48,
      display: "flex",
      alignItems: "center",
      justifyContent: "center",
    }}
  >
    <span>ship it tomorrow?</span>
  </AbsoluteFill>
);
```

Render with `renderStillOnWeb({ composition: <Repro/>, frame: 0 })`. The PNG comes out with "shipit tomorrow?" — the space between "ship" and "it" is rendered at zero width. With longer text it's clearer:

**Before (broken):**

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/ba6084b4-07cf-4c6a-ba08-278a2fb49d3d" />

**After (the patch below):**

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/626c4c84-cb62-46e8-af75-6a9966a54682" />

Same composition. Same input props. Only the renderer changed.

Reproduced on @remotion/web-renderer 4.0.459, macOS Chrome (host browser running the studio), Node 22, Bun 1.3.6.

## Root cause

`findWords` in `dist/esm/index.mjs` (source: `src/drawing/text/find-line-breaks.text.ts`) chops the text into segments via `Intl.Segmenter`, then for each segment temporarily mutates the parent span:

```
<span (parent)>
  <text>{stuff before this segment}</text>
  <span>{this segment}</span>
  <text>{stuff after this segment}</text>
</span>
```

[...]

PR #7522, opened 2026-05-21
`@remotion/web-renderer`: Add issue-7489 whitespace-collapsing fixture

JonnyBurger:
## Summary

Adds a fixture for #7489 (`<span>ship it tomorrow?</span>`) so the bug stays reproducible in the studio testbed for future debugging. **Does not fix the underlying bug** — an attempted rewrite of `findWords` (one-token-per-line via the Range API) caused regressions and was reverted.

Note: the minimal repro from the issue does not actually reproduce the visible space-collapsing in the test harness — the existing `whitespace-collapsing.test.tsx` fixtures (multiple `<span>` siblings) are still the more reliable reproducers. We've left a comment on #7489 asking the reporter for a tighter repro.

## What's in the diff

- New fixture `packages/web-renderer/src/test/fixtures/issue-7489-minimal.tsx` mirroring the issue's minimal repro.
- Visual regression test `issue-7489-minimal.test.tsx` with chromium/firefox/webkit baselines.
- `Root.tsx` registration so the composition is selectable in `bun run studio`.

## Test plan

- [ ] CI passes (the new test is currently green; baselines were captured against the original `Intl.Segmenter` implementation).


JonnyBurger (commented):
This does not work, several tests are now broken and collapse too much whitespace, such as /packages/web-renderer/src/test/fixtures/whitespace-collapsing-2.tsx

PR #7596, opened 2026-05-24
`@remotion/studio-server`: Resolve re-exported composition editor links

### LINKED ISSUE #7590: `Open composition in [editorName]` should have some recursion
If we try this action on `html-in-canvas-effects-halftone-preview`, then it resolves to `packages/example/src/HtmlInCanvas/index.tsx` instead of `packages/example/src/HtmlInCanvas/effects-halftone-preview.tsx`

Nit: It is `Show composition` but `Open component`, would make more sense to have both as `Show`

JonnyBurger:
Needs manual testing, probably fine

### DIFF
--- a/packages/studio-server/src/helpers/resolve-composition-component.ts
+++ b/packages/studio-server/src/helpers/resolve-composition-component.ts
@@ -1,6 +1,9 @@
 import fs from 'node:fs';
 import path from 'node:path';
 import type {
+	ExportAllDeclaration,
+	ExportNamedDeclaration,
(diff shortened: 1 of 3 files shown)


JonnyBurger:
Addressed Pullfrog feedback: fixed recursion backtracking in re-export resolution by treating `visited` as a recursion-stack (with proper unwind in `finally`), and made unresolved re-export-only branches throw so sibling branches can be tried. Added a regression test for cycle + fallback branch resolution.

PR #7609, opened 2026-05-24
`@remotion/studio`: Open docs from timeline labels

JonnyBurger:
LGTM, waiting for pullfrog


packages/core/src/CompositionManager.tsx
```diff
@@ -106,6 +106,7 @@ export type TSequence = {
 	duration: number;
 	id: string;
 	displayName: string;
+	documentationLink?: string;
```
JonnyBurger: make this not optional, but instead accept string | null

this will lead to a lot of errors, fix all of them and whereever possible, find a documentation link

packages/studio/src/components/Timeline/TimelineStack/index.tsx
```diff
@@ -204,9 +213,11 @@ export const TimelineStack: React.FC<{
 				onPointerEnter={onTitlePointerEnter}
 				onPointerLeave={onTitlePointerLeave}
 				title={
-					originalLocation
-						? getOriginalSourceAttribution(originalLocation)
-						: text || '<Sequence>'
+					documentationLink
```
JonnyBurger, line 148: Let's simplify this now - this is clickable when documentation link is there.

We never make it open the source, that is only possible with the secondary stack link from now on.

PR #7610, opened 2026-05-24
`@remotion/renderer`: Add GOP size render option

JonnyBurger:
Looks good to me, just waiting for pullfrog comments


JonnyBurger (commented):
Consider pullfrog comments.

Test every encoder to see if it is supported everywhere.

Comment on this issue once you are done

PR #7661, opened 2026-05-25
`@remotion/effects`: Add `contrast()` effect

JonnyBurger:
Needs demo like the other effects


JonnyBurger:
needs demo like the other examples @pullfrog

PR #7674, opened 2026-05-25
`@remotion/media`: Support subframe audio rendering

### LINKED ISSUE #7666: Can subframe perfect audio be rendered?
packages/example/src/SubframeAudio/index.tsx

Questions:

- During rendering, what assets do we register at frame 23?
- Do we respect the 23.5 and on the 24th frame trim off half of the frame at the end?
- Do we respect the 0.5sec offset and trim off half a frame at the beginning?

Idea: We render a video of 1 frame using web renderer and analyse the waveform.
Sine and square waves have a distinct pattern that we can recognize.

JonnyBurger:
## Summary
- Render fractional audio starts/ends by trimming and padding inline audio chunks
- Queue web-renderer inline audio by target frame so backfilled subframe chunks are included
- Add a web-renderer WAV regression test for subframe audio start timing

Fixes #7666

## Tests
- bun run build
- bun run formatting
- cd packages/web-renderer && bunx vitest run src/test/audio.test.tsx

### DIFF
--- a/packages/docs/docs/client-side-rendering/limitations.mdx
+++ b/packages/docs/docs/client-side-rendering/limitations.mdx
@@ -114,7 +114,7 @@ Filters are not supported in Safari/WebKit. Use Chrome or Firefox for filter sup
 The following Remotion components are supported:
 
 The [`<Video>`](/docs/media/video) from `@remotion/media` is <a style={{color: 'green'}}>supported</a>.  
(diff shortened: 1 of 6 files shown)


JonnyBurger:
https://github.com/remotion-dev/remotion/issues/7666#issuecomment-4554237342

PR #7693, opened 2026-05-26
`remotion`: Add Img effects fallback

### LINKED ISSUE #7599: How can an effect be applied to an <Img>
`` is not canvas based, but I think applying effects will be a much demanded use case.
What about if you apply an effect, we mount a `<canvas>` and use `new Image()` as the source?

### LINKED ISSUE #7613: `@remotion/core`: Make `<Img>` fall back to `<CanvasImage>` when effects are passed
## Summary

Make `` automatically render through the new canvas-backed image component when `effects.length > 0`.

## Proposed design

Make the current "Img" tag the inner tag.
Create a new outer tag which will become the new Img which has the same props as Img, but add `effects?: EffectsProp` to the outer tag.

Behavior should be split as follows:

- If `effects` is omitted or resolves to an empty array, route to the inner `` implementation unchanged.
- If `effects.length > 0`, render the canvas-backed image implementation from `<CanvasImage>` instead of a native ``.

## Conflicting props

When `` uses the canvas, some native image props cannot be honored. If `effects.length > 0`, `` should throw when conflicting props are passed.

Examples of likely conflicting props:

- `ref`
- `srcSet`
- `sizes`
- `loading`
- `decoding`
- `fetchPriority`
- `useMap`

## Acceptance criteria

- `` keeps rendering an `IMG` element.
- `` renders a `CANVAS` element through the canvas-backed image path.
- Existing `` behavior remains unchanged when no effects are passed.
- Passing unsupported native image props together with non-empty `effects` throws a clear error.
- Timeline/effect controls continue to work with the fallback path.
- Docs mention that `` automatically uses `<CanvasImage>` when effects are passed, and explain the conflicting prop restrictions.

JonnyBurger:
## Summary
- Add `effects` to `` and route non-empty effect arrays through `<CanvasImage>`.
- Keep empty or omitted effects on the existing native `` path.
- Throw clear errors for native image props that cannot be honored on the canvas fallback.
- Document the new behavior and restrictions.

Fixes #7613

## Tests
- `cd packages/core && bun test src/test/Img.test.tsx src/test/canvas-image.test.tsx src/test/truncate-src-for-label.test.ts`
- `bunx turbo make --filter="remotion"`
- `bun run build-docs`
- `cd packages/docs && bun render-cards.ts`
- `cd packages/core && bun test src/test`


JonnyBurger:
@pullfrog /formatting

PR #7740, opened 2026-05-28
docs: fix interpolate easing heading anchor

### LINKED ISSUE #7708: Easing docs have a weird title
`Per-segment easing (array) {#per-segment-easing-array}`

We tried to link here but it does not work

https://www.remotion.dev/docs/interpolate#per-segment-easing-array

Image

subhwastaken:
## What changed

Moves the `AvailableFrom` badge out of the `Per-segment easing (array)` heading in the interpolate docs.

## Why

Fixes #7708. The MDX heading was rendering the explicit `{#per-segment-easing-array}` anchor as part of the visible title, which made the section title look wrong and the intended anchor link unreliable.

## Verification

Docs-only change; no local docs build was run.

### DIFF
--- a/packages/docs/docs/interpolate.mdx
+++ b/packages/docs/docs/interpolate.mdx
@@ -176,7 +176,9 @@ interpolate(frame, [0, 10, 40, 100], [0, 0.2, 0.6, 1], {
 });
 ```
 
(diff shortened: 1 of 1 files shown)


JonnyBurger:
Thanks! slightly too late, I already fixed this earlier today 😅

PR #7742, opened 2026-05-28
fix(renderer): kill remaining child processes to prevent zombies

Jefsky:
## Summary

Fixes #7065.

After rendering completes, multiple zombie chrome-headless-shell processes are left behind with PPID 1. This happens because:

1. The `setpriv --pdeathsig SIGKILL` fix (from #7207) requires `util-linux` which is not installed in the official Docker images
2. When the process group kill fails (e.g. due to permissions), only the main Chrome process is killed, leaving children as orphans

### 1. Docker images: install `util-linux`

Added `util-linux` to all Dockerfiles that use apt/yum to ensure `setpriv` is available. This is the primary defense — `setpriv --pdeathsig SIGKILL` makes child processes receive SIGKILL when their parent dies.

### 2. Renderer code: best-effort child process cleanup

After killing the main process group, use `pkill -9 -P <pid>` as a fallback to kill any remaining child processes. This handles edge cases where:
- `setpriv` is not available
- Process group kill fails due to permissions
- Chrome spawns children outside its process group

The `pkill` call is wrapped in a try/catch with a 2-second timeout to avoid blocking if `pkill` is not available or the process is already gone.

## Testing

- Verified that `setpriv` is now available in the Docker images
- The `pkill` fallback is a no-op when no children remain (exit code 1, caught by try/catch)

PR #7745, opened 2026-05-28
`@remotion/renderer`: kill remaining child processes after browser shutdown

JonnyBurger:
## Summary

Best-effort cleanup for any child processes left behind after the browser process group is killed.

Split from #7742.

Authored-by: Jefsky <jefsky@users.noreply.github.com>

### DIFF
--- a/packages/renderer/src/browser/BrowserRunner.ts
+++ b/packages/renderer/src/browser/BrowserRunner.ts
@@ -128,6 +128,19 @@ export const makeBrowserRunner = async ({
 						);
 						proc.kill('SIGKILL');
 					}
+
+					// Best-effort: kill any remaining child processes that may have been
+					// missed by the process group kill (e.g. if setpriv is not available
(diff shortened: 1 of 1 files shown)


JonnyBurger:
Closing because the approach doesn't reliably solve the stated problem.

After browser process group kill on Linux, child processes are reparented by the kernel before userspace gets a chance to run. By the time `pkill -P` executes, any surviving children have already been moved to PID 1 or a subreaper, so the PPID no longer matches `proc.pid`. This means the command almost always finds zero processes and does nothing.

Additionally, there are two real hazards:
1. **PID reuse**: if `proc.pid` gets reaped and reused before `pkill` runs, the command kills children of an unrelated process
2. **Blocking shutdown**: `execSync(..., {timeout: 2000})` in a process.exit handler can stall Node.js exit by up to 2 seconds in minimal containers

A reliable fix would need to either enumerate descendants before killing the leader, or use a more stable scope like process group/session instead of PPID-after-teardown.

PR #7754, opened 2026-05-28
fix(studio): truncate overflowing timeline labels

### LINKED ISSUE #7752: Overflow of timeline list label
Image

subhwastaken:
## What changed

Fixes timeline list labels overflowing their row in narrow Studio layouts by making the label container and its text children explicitly shrinkable.

## Why

Fixes #7752. The title/source labels already used `overflow: hidden` and `textOverflow: ellipsis`, but flex items keep `min-width: auto` by default. That can make long labels keep their intrinsic width and spill outside the timeline list instead of truncating.

This change adds `minWidth: 0` to the shrinkable label text nodes and hides overflow at the label-row container level.

## Verification

- `git diff --check`

I could not run the Studio test suite locally because this sparse checkout does not have `node_modules` installed.

### DIFF
--- a/packages/studio/src/components/Timeline/TimelineStack/index.tsx
+++ b/packages/studio/src/components/Timeline/TimelineStack/index.tsx
@@ -117,6 +117,7 @@ export const TimelineStack: React.FC<{
 			textOverflow: 'ellipsis',
 			overflow: 'hidden',
 			flexShrink: 100000,
+			minWidth: 0,
 			userSelect: 'none',
 			WebkitUserSelect: 'none',
 		};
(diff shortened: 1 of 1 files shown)


JonnyBurger:
Hmm, no the described problem is not in TimelineStack.tsx...

subhwastaken:
You were right, the screenshot is about the expanded field rows rather than TimelineStack. I moved the fix to the expanded timeline field label styles so labels like "Playback Rate" stay on one line and truncate inside narrow panels instead of wrapping into the value column.

GitHub would not let me reopen this closed PR, so I opened the corrected patch here: https://github.com/remotion-dev/remotion/pull/7757

PR #7757, opened 2026-05-28
Fix expanded timeline field label wrapping

subhwastaken:
## Summary
- Keep expanded timeline field labels on one line inside narrow label columns
- Truncate long field labels with ellipsis so value controls stay aligned
- Apply the same label behavior to sequence and effect field rows

## Validation
- npx --yes oxfmt packages/studio/src/components/Timeline/TimelineFieldRow.tsx packages/studio/src/components/Timeline/TimelineEffectFieldRow.tsx packages/studio/src/components/Timeline/timeline-field-row-layout.ts --check
- git diff --check


JonnyBurger:
Thanks a lot! The fix looks correct, but I already fixed it and merged it earlier 😐 Sorry!

PR #7817, opened 2026-05-30
`remotion`: Add OffthreadVideo frame callback

alec-watts:
## Summary
- Adds an `onVideoFrameCallback` prop to `<OffthreadVideo>` that mirrors `HTMLVideoElement.requestVideoFrameCallback()` during preview.
- Threads the callback through the preview video path and exports the new `OnVideoFrameCallback` type.
- Documents the callback and allows `@remotion/media` fallback OffthreadVideo props to pass it through.

## Test plan
- `bun run format`
- `bunx tsc -p packages/core/tsconfig.json --noEmit`

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -238,7 +238,7 @@ export {
 	Video,
 } from './video/index.js';
 export {MediaPlaybackError} from './video/MediaPlaybackError.js';
-export type {OnVideoFrame} from './video/props.js';
+export type {OnVideoFrame, OnVideoFramePresented} from './video/props.js';
 export type {VolumeProp} from './volume-prop.js';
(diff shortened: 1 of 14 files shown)


JonnyBurger:
@alec-watts thanks again for PR #7817. After merging, we had second thoughts about the API shape and decided the simpler long-term direction is to extend the existing `onVideoFrame` callback with optional `requestVideoFrameCallback()` metadata args instead of adding a separate `onVideoFrameCallback` prop.

I opened #8064 with that adjustment. It removes the new prop/type and makes `onVideoFrame(frame, now?, metadata?)` cover the same preview metadata use case while keeping one video-frame hook API.

alec-watts:
Sounds great. Thanks for hard work!

PR #7864, opened 2026-05-31
`@remotion/webcontainer-studio`: Add new package

JonnyBurger:
## Summary

- Add a new `@remotion/webcontainer-studio` package that runs Remotion Studio in a browser via WebContainers.

- Boot a WebContainer, mount template files, run `npm install`, and start Studio in an iframe.

- Add Vite config with COOP/COEP headers required for SharedArrayBuffer/WebContainers.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -2523,6 +2523,23 @@
         "vite": "6.4.2",
       },
     },
+    "packages/webcontainer-studio": {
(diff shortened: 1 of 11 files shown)


JonnyBurger:
Interesting, but not gonna pursue right now, too early

PR #7972, opened 2026-06-01
`@remotion/promo-pages`: Fix investors link

samohovets:
## Summary
- Point the team page investors link directly to `/docs/investors`
- Confirmed `/investors` already redirects to `/docs/investors`

## Testing
- `cd packages/promo-pages && find src -type f \( -name '*.ts' -o -name '*.tsx' \) -print0 | xargs -0 bunx oxfmt --write`
- `bun run build`
- `bun run formatting`


samohovets:
@JonnyBurger FYI: there's a redirect from `/investors` to `/docs/investors` but it's not following the redirects locally

I initially thought it's client-side routing but it's not! So this fix is for local dev only

PR #7980, opened 2026-06-01
`@remotion/docs`: Move whiteNoise to Stylize effects

### LINKED ISSUE #7957: noise should be in stylize category

subharup31-crypto:
Fixes #7957.

## Problem
The effects overview currently groups `whiteNoise()` under Generate. Since it overlays randomized grayscale grain on the source, it fits the Stylize group better than procedural source generation.

## Approach
Move the `whiteNoise()` table-of-contents entry from Generate to Stylize while preserving the existing link, preview, alt text, name, and description.

## Tradeoffs
This is a docs/navigation-only change and does not affect the effect implementation or public API.

### DIFF
--- a/packages/docs/docs/effects/table-of-contents.tsx
+++ b/packages/docs/docs/effects/table-of-contents.tsx
@@ -237,6 +237,13 @@ const categories: {
 				name: 'vignette()',
 				description: 'Edge darkening or transparency effect',
 			},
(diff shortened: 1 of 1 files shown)


JonnyBurger:
nope, noise != whitNoise

the fix was already merged, but thank you

PR #7996, opened 2026-06-02
`@remotion/web-renderer`: Scale text-shadow with the export scale option

UmungoBungo:
## Problem

`@remotion/web-renderer` rasterizes the DOM to a canvas and bakes the export `scale` (and any element transform) into the canvas transform matrix. Per the canvas drawing model, `shadowOffsetX`, `shadowOffsetY` and `shadowBlur` are measured in the **output bitmap** and are **not** affected by the current transform — unlike `fillText`, which goes through the CTM.

As a result, when exporting at a higher resolution (e.g. a 1280×720 composition at `scale: 3`), text glyphs scale correctly but their `text-shadow` offset/blur stay at the original, unscaled pixel size.

## Fix

In `draw-text.ts`, read the live CTM and apply its linear part to the shadow offset vector (and its magnitude to the blur) before assigning the canvas shadow properties:

```ts
contextToDraw.shadowOffsetX = shadow.offsetX * ctm.a + shadow.offsetY * ctm.c;
contextToDraw.shadowOffsetY = shadow.offsetX * ctm.b + shadow.offsetY * ctm.d;
contextToDraw.shadowBlur    = shadow.blurRadius * Math.hypot(ctm.a, ctm.b);
```

This is a no-op at `scale: 1` and also keeps shadows correct on rotated/scaled elements.

## Why `box-shadow` is not touched

`box-shadow` is handled differently: it bakes the shadow onto a fresh `OffscreenCanvas` (identity transform) and composites it with `ctx.drawImage(...)`, where `ctx` already carries the scale CTM. Since `drawImage` respects the CTM, the baked offset already scales — multiplying by the scale there would double-scale it. Verified empirically with a Chromium canvas probe (a directly-drawn shadow keeps its unscaled offset, while a `drawImage`-composited one scales).

## Test

Adds `text-shadow-scale.test.tsx` with a fixture rendering a glyph with a hard horizontal `text-shadow`. The test decodes the output PNG and asserts the shadow offset scales ~3× between `scale: 1` and `scale: 3` (rather than a screenshot snapshot, so it asserts the behavior directly and needs no per-browser baselines). Confirmed it fails without the fix (ratio ≈ 1.0) and passes with it (ratio ≈ 3).

### DIFF
--- a/packages/web-renderer/src/drawing/text/draw-text.ts
+++ b/packages/web-renderer/src/drawing/text/draw-text.ts
@@ -96,6 +96,10 @@ export const drawText = ({
 
 		const textShadows = parseTextShadow(textShadowValue);
 
+		// Shadow offset/blur ignore the CTM (unlike fillText), so apply it manually.
+		const ctm = contextToDraw.getTransform();
+		const blurScale = Math.hypot(ctm.a, ctm.b);
+
 		const {strokeFirst} = parsePaintOrder(paintOrder);
 
 		for (const token of tokens) {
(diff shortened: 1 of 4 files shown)


JonnyBurger:
Nice way to test, thank you Matt! 😃

PR #8046, opened 2026-06-03
`@remotion/studio`: Fix keyframe navigation button clicks

JonnyBurger:
## Summary
- Keep keyframe controls above the overlapping row chrome so previous/next buttons receive pointer events
- Preserve the existing timeline row layout

## Testing
- `bun test packages/studio/src/test/timeline-keyframe-navigation.test.ts`
- `bun run build`
- `bun run formatting`


JonnyBurger (commented):
I don't understand why we need zIndex for this. The diamond itself also doesn't need it? Why do we need it for the arrows? Are we not messing up with the fundamentals

Issue #7698, opened 2026-05-26
Elements masterplan


JonnyBurger:
## Purpose

This is the continuous masterplan for Remotion Elements: reusable, remixable motion graphics that users can preview, install as source code, and edit in Remotion Studio.

Simple graphic Elements are preferred over complex compositions so they remain understandable and composable.

## Current execution

- **Remotion Elements v1 release:** #9769
- **Remotion Elements v1 announcement:** #9770

The v1 release is intentionally separate from the larger announcement. Elements will first become officially available, complete a bounded soak period, and then receive the larger launch campaign.

## Planning model

- Finite releases and announcements get dedicated tracking sub-issues with explicit definitions of done.
- This masterplan remains open as the continuous roadmap.
- Direct sub-issues outside the active release trackers are post-v1 ideas and improvements.
- Once a release scope is agreed, new blockers are only added for broken core flows or significant security problems.
- New Elements should remain individually scoped and follow the standard Element contribution and publishing requirements.

## Product direction

The Elements library should make motion graphics easy to discover, understand, install, remix, and compose. Over time, the catalog and authoring workflow can expand without preventing finite releases from being completed.

JonnyBurger:
Maps:

- Mapbox line A to B
- 3D building
- World map with country picker that colors the picked country and zooms-in: https://bklit.com/studio?s=v1.NrAMBoCIGMAsHsBO8AOAbApgF1gWjgIaJaQC6pQA

Sport:
- Strava visualization
- Mapbox / Athletes Eye demo

Storytelling
- Paper.js highlight
- News headline
- Bar Chart
- Pie Chart
  - suggestion: parameterize the number of slices & colors, make order of slices configurable, show and hide values, select unit for the values, etc. 
- Number Wheel

Audio
- Music visualizer
- Audiogram
- Sound wave (inspiration: https://www.reactvideoeditor.com/remotion-templates/sound-wave)

Captions
- TikTok template Captions style
- Some lyrics animation from Jonnys Party

Motion design system
- Lower third
- Code example

Effects:
- Greenscreen
- Wiggle
- Magnifier
- Sticker Peel
- Corner Pin https://x.com/Remotion/status/2071138304070717829

WebGL:
- Chromatic abboration

3D:
- Spinning object. e.g Disco ball
- Video texture

Remotion Best Practices:
- Drag and drop
- Jumpcuts technique

Everybody may post their ideas!

MehmetAdemi:
[...]

PR #8080, opened 2026-06-03
Docs: Add examples section structure

samohovets:
## Summary

Adds a draft top-level `/examples` docs section for issue #7698, backed by a dedicated Docusaurus docs plugin/sidebar and kept out of the main navbar for now.

This now includes the initial example-page system:

- category pages and sidebar entries for planned examples
- individual placeholder pages for planned examples
- `Submit an example` instead of `Suggest an example`
- PR-based submission instructions
- a colocated example layout: `packages/docs/examples/<category>/<slug>/index.mdx` next to the example TSX source
- a reusable page template with Preview, Use it, and Source sections
- a reusable copyable template in `packages/docs/examples-template/`
- a first live lower-third example using `@remotion/player`
- `source:./file.tsx` code-fence support so the displayed source is loaded from the colocated TSX file before Twoslash/Shiki runs
- Twoslash prewarming and plugin wiring for the new examples docs section
- Shiki/Twoslash stylesheet inclusion for examples code blocks

## Notes

The examples section is still intentionally draft/WIP. The current real example is mainly there to validate the proposed live-preview + source-display workflow before adding more examples.

The Remotion skill for creating examples is tracked separately in sub-issue #8158, not in this PR.

## Follow-up checklist

- [ ] Add more real examples using the colocated template
- [ ] Decide which examples should use live Player previews vs. rendered fallback previews

## Testing

- `bunx turbo run make --filter=@remotion/docusaurus-plugin`
- `bun run build-docs`
- `bun run formatting`


packages/docs/src/components/Examples/ExamplePage.tsx
```diff
@@ -0,0 +1,44 @@
...
+				component={component}
+				durationInFrames={durationInFrames}
+				fps={fps}
+				height={height}
+				width={width}
+			/>
+
+			<h2>Use it</h2>
+			<p>
+				Copy the source code into your Remotion project. The file exports the
```
JonnyBurger, line 34: Is any example using this already or is this dead code?

I don't see any example actually being implemented

packages/docusaurus-plugin/src/source-code.ts
```diff
@@ -0,0 +1,47 @@
...
+	value: string;
+};
+
+type Options = {
+	root?: string;
+};
+
+const sourcePrefix = 'source:';
+
+export const remarkSourceCode = (options: Options = {}) => {
```
JonnyBurger: Is there any example of this code in action? I couldn't find it

JonnyBurger (changes requested):
Feels like we will need to delete a lot of examples and boilerplate before put it public.

Maybe only 1-2 examples for the start but actually implement them rather than 20x placeholder?

packages/docs/examples/3d/video-texture.mdx
```diff
@@ -0,0 +1,10 @@
+---
+title: Video texture
+description: Use video as a texture in a 3D scene.
+---
+
+import {PlannedExample} from '../../src/components/Examples/PlannedExample';
+
+# Video texture
+
+<PlannedExample category="3D" description="Use video as a texture in a 3D scene." />
```
JonnyBurger: Potential drawback of this approach:

For agents, we automatically serve Markdown, so if one pastes an example link to the agent, it seems like the agent cannot see the source code even though the user can in the browser

packages/docs/src/components/Examples/ExamplePage.tsx
samohovets: Yes, here: https://github.com/remotion-dev/remotion/pull/8080/changes#diff-db8331ddf89d4f9f322b637d61db6e4e7c346bfb1865be95083a3b4024fad392

packages/docusaurus-plugin/src/source-code.ts
samohovets: Yessir: https://github.com/remotion-dev/remotion/pull/8080/changes#diff-db8331ddf89d4f9f322b637d61db6e4e7c346bfb1865be95083a3b4024fad392

packages/docs/examples/3d/video-texture.mdx
samohovets: Right... at the same time, I want to enforce the structure. 

Maybe could be just a test file that checks whether specific strings are present across examples. The same approach we have in tests for templates (e.g package.json fields are present, etc).

samohovets:
> Feels like we will need to delete a lot of examples and boilerplate before put it public.
> 
> Maybe only 1-2 examples for the start but actually implement them rather than 20x placeholder?

The page is not in the public navbar yet and not discoverable, I thought it's better to add pages upfront to have a docs-level TODO list. Can remove them as well!

PR #8085, opened 2026-06-03
`@remotion/studio`: Forward error screens to server logs

### LINKED ISSUE #7202: Forward errors to dev server so that agent can see it
Some agents manage the dev server. it would be useful to surface it so that when there is an error, agents can see it and react to it!

samohovets:
## Summary

- Forward Studio error screens to the Studio server via `/api/log-studio-error`
- Log forwarded errors with the `studio-frontend` tag
- Dedupe repeated frontend reports and swallow logging failures so the reporter cannot trigger another overlay
- Add server-side coverage for the logging route

Fixes #7202.

## Testing

- `bun run build`
- `bun run formatting`
- `cd packages/studio-server && bun test src/test/log-studio-error.test.ts`
- `cd packages/example && bunx playwright test e2e/verify-studio-error-log.tmp.test.mts --reporter=line` (temporary local verification test removed after run)


JonnyBurger (changes requested):
- Errors during renders are more nicely formatted than during preview, can we make it more nice and also more consistent?:

Image

Image

- `studio-frontend` is not a term we use anywhere else. Maybe a full phrase "Error occured in Studio:"?
- Would it not look better if we had no indentation?

samohovets:
@JonnyBurger awesome feedback, thank you. Didn't think we can reuse that. Now better indeed!

- no indentation
- "Error occured in Studio"
- looks more like the errors users can see during rendering

Zed 2026-06-05 12 11 29

PR #8238, opened 2026-06-08
`remotion`: Support interpolating transform-origin values

### LINKED ISSUE #8237: `remotion`: Support transform-origin keywords in interpolate()
`style.transformOrigin` should be keyframable in the visual editor. The implementation should use `interpolate()` directly rather than a transform-origin-specific helper.

## Why

`interpolate()` already supports translate-like CSS strings such as `50% 50%` and `100px 40px`, which matches the numeric syntax used by `transform-origin`.

The missing piece is keyword support. CSS transform-origin aliases such as `left top`, `center`, and `right bottom` should normalize to their equivalent percentage positions before interpolation.

Once `interpolate()` supports those keyword aliases, #8070 can keyframe `style.transformOrigin` by routing it to `interpolate()`.

## Scope

- Extend `interpolate()` string interpolation to support `transform-origin` keyword aliases.
- Normalize `left`, `center`, `right`, `top`, and `bottom` according to CSS transform-origin semantics.
- Keep percentage and length interpolation using the existing translate-like string interpolation behavior.
- Preserve or explicitly handle the optional z component. The z component should be a length.
- Keep mixed-unit behavior explicit, for example keyword/percentage values should not silently interpolate with `px` values on the same axis.
- Route `style.transformOrigin` schema fields to `interpolate()` for keyframes.
- Update docs and add tests for percentage, keyword, length, mixed-unit, z-component, and invalid-value behavior.

This is a prerequisite for #8070.

JonnyBurger:
## Summary

- Add interpolateTransformOrigin() for keyword-aware CSS transform-origin interpolation.
- Add style.transformOrigin to visual style schemas and route keyframes to the new helper.
- Wire Studio status parsing, keyframe codegen, timeline editing, tests, and docs.

Fixes #8237

## Testing

- bun test packages/core/src/test/interpolate-transform-origin.test.ts packages/core/src/test/interpolate-keyframed-status.test.ts packages/studio-shared/src/test/keyframe-interpolation-function.test.ts packages/studio-shared/src/test/optimistic-add-keyframe.test.ts packages/studio-server/src/test/update-keyframes.test.ts packages/studio-server/src/test/update-keyframes-imports.test.ts packages/studio-server/src/test/compute-sequence-props-status.test.ts
- bun run build
- bun run stylecheck


JonnyBurger:
Closing in favor of implementing transform-origin keyword support directly in interpolate().

Issue #8198, opened 2026-06-05
`selectComposition` stalls during server-side rendering on node 24.16.0


DanTheManGude:
# Bug Report 🐛

Remotion version `4.0.447`
Docker container running `ubuntu-22.04`
Node version `24.16.0`

During server side rendering we invoke a `mjs` script in order to perform the rendering. In this script we `await` [`selectComposition`](https://www.remotion.dev/docs/renderer/select-composition) from `@remotion/renderer`. [Log level](https://www.remotion.dev/docs/renderer/select-composition#loglevel) `verbose` is used. Some log messages (below) are made but it perpetually hangs. 

When using the same Dockerfile with the same Remotion version, but specifically install node `24.15.0`. Then there is no issue.

(These logs appear during successful renders as well) 
```
Ensuring browser executable
Created directory for temporary files /tmp/remotion-v4.0.447-assets02nsveft7c
Downloading Chrome Headless Shell https://www.remotion.dev/chrome-headless-shell
Customize this behavior by adding a onBrowserDownload function to selectComposition().
 compositor  Starting Rust process. Max video cache size: 2509MB, max threads = 0
Downloading from: https://remotion.media/chromium-headless-shell-linux-arm64-144.0.7559.20.zip?clearcache
Downloading Chrome Headless Shell - 9.5 Mb/83.5 Mb
Could not fetch sourcemap for  {redacted serve url} SyntaxError: Unexpected token '<', "<!DOCTYPE "... is not valid JSON
Downloading Chrome Headless Shell - 19.1 Mb/83.5 Mb
Downloading Chrome Headless Shell - 28.6 Mb/83.5 Mb
Downloading Chrome Headless Shell - 38.2 Mb/83.5 Mb
Downloading Chrome Headless Shell - 47.7 Mb/83.5 Mb
Downloading Chrome Headless Shell - 57.3 Mb/83.5 Mb
Downloading Chrome Headless Shell - 66.8 Mb/83.5 Mb
Downloading Chrome Headless Shell - 76.4 Mb/83.5 Mb
Downloading Chrome Headless Shell - 83.5 Mb/83.5 Mb
```

JonnyBurger:
I can reproduce this with a minimal Docker SSR setup.

I made a standalone repro using Ubuntu 22.04, Node 24.16.0, Remotion 4.0.447, and a tiny bundled composition. It intentionally lets `selectComposition()` download Chrome Headless Shell during the run, matching the reported failure path.

Result:

[...]

PR #8243, opened 2026-06-08
`@remotion/renderer`: Add Node 24 selectComposition repro

JonnyBurger:
## Summary

- Add a standalone Docker repro for #8198 using Ubuntu 22.04, Node 24.16.0, and Remotion 4.0.447.
- Allow overriding NODE_VERSION and REMOTION_VERSION to compare the failing version against patched/current Remotion versions.
- Document the observed failure and the successful Remotion 4.0.474 control run.

## Testing

- `cd packages/dockerfiles/issue-8198-node-24-select-composition && REMOTION_VERSION=4.0.474 SELECT_COMPOSITION_TIMEOUT_MS=120000 ./run.sh`
- `cd packages/dockerfiles/issue-8198-node-24-select-composition && SELECT_COMPOSITION_TIMEOUT_MS=30000 ./run.sh` (expected timeout on Remotion 4.0.447)
- `bun run build`
- `bun run stylecheck`


JonnyBurger:
Closing this because it only contains the Docker reproduction. The actual yauzl/fd-slicer lifecycle fix is already present on main, so this is not the fix-only PR we want.

Issue #8070, opened 2026-06-03
Should allow editing the transform origin


JonnyBurger:
## Remaining tasks

- [ ] Add a transform-origin crosshair / pivot handle to the selected outline when the `style.transformOrigin` sequence prop or one of its keyframes is selected.
- [ ] Make the pivot handle draggable for a single selected sequence.
- [ ] Implement pan-behind compensation while dragging: update `style.transformOrigin` and apply the corresponding `style.translate` compensation in the same drag transaction / undo step.
- [ ] Preserve keyframed behavior: dragging `style.transformOrigin` should add or update a keyframe at the current source frame, and if `style.translate` is keyframed, the compensation should add or update its keyframe too.
- [ ] Serialize dragged origins as percentages, preserve an existing z component, allow origins outside `[0%, 100%]`, and disable the pivot handle for parser-unsupported values such as complex `calc(...)` expressions.
- [ ] Scope the first implementation to the supported individual transform props (`style.translate`, `style.rotate`, `style.scale`); arbitrary `style.transform` / matrix transforms can stay best-effort or out of scope.

JonnyBurger:
Recommendation: Add `style.transformOrigin` as a distinct, string-backed visual style field, not as `uv-coordinate` and not as a generic string. It should parse into a 2D coordinate for UI, but serialize back to CSS like `"50% 50%"`.

The important design choices:

1. **Schema shape**
   Add a semantic field type like `type: 'transform-origin'` with default `"50% 50%"` and description `"Transform Origin"`. This keeps the public API honest: the value is CSS, but the UI can treat it as a 2D point.

2. **Canvas knob**
   Reuse the existing quad/UV helpers from `SelectedOutlineOverlay`: `getUvHandlePosition()` and `getUvCoordinateForPoint()`. When the `style.transformOrigin` sequence prop is selected, draw a distinct pivot/crosshair handle for that selected sequence. Also show the crosshair when a `style.transformOrigin` keyframe is selected.

3. **Pan-behind drag behavior**
   The crosshair should be draggable directly; no separate “pan behind” mode is needed. While dragging, behave like After Effects’ Pan Behind tool: update `style.transformOrigin` and apply a compensating `style.translate` change so the rendered element stays visually in place while the pivot moves. This means the drag is not merely “set CSS transform-origin”; it is “move the local-space pivot while preserving visual placement”.

[...]

PR #8252, opened 2026-06-08
`@remotion/studio`: Add support for editing transform origin

### LINKED ISSUE #8237: `remotion`: Support transform-origin keywords in interpolate()

JonnyBurger:
## Summary

- Add `style.transformOrigin` to the visual style schema
- Route transform-origin keyframes through `interpolate()`
- Add X/Y `InputDragger` controls for editing transform origins in the Studio timeline
- Update schema tests and docs demo controls for the new field type

Stacked on #8248.
Fixes #8237.
Part of #8070.

## Testing

- `bun test packages/core/src/test/find-props-to-delete.test.ts packages/core/src/test/wrap-in-schema-helpers.test.ts packages/core/src/test/interpolate-keyframed-status.test.ts packages/studio-shared/src/test/keyframe-interpolation-function.test.ts packages/studio-server/src/test/discriminated-union-update.test.ts packages/studio-server/src/test/get-all-schema-keys.test.ts`
- `bun test packages/studio/src/test/timeline-transform-origin-utils.test.ts`
- `bunx turbo make --filter=@remotion/studio`
- `bun run build`
- `bun run stylecheck`


JonnyBurger:
Closing in favor of #8265, which is the current rebased PR for transform origin editing.

PR #8262, opened 2026-06-09
`@remotion/studio`: Polish timeline styling

### LINKED ISSUE #8036: Timeline ticks should be a different color than tracks after behavior change
Before, you could use both to seek. I expect people to mistakenly use the tracks surface to seek going forward.

We could change the color so the seek area and the tracks area have a different color

JonnyBurger:
## Summary

- Polish timeline background and tick styling
- Adjust timeline sequence label spacing and stack display
- Update example compositions with transformed keyframed props for visual testing

## Testing

- `cd packages/example && bunx oxfmt src --write`
- `bun run build`
- `bun run stylecheck`


packages/studio/src/components/Timeline/TimelineTimeIndicators.tsx
```diff
@@ -52,7 +56,7 @@ const timeValue: React.CSSProperties = {
 	height: TIMELINE_TIME_INDICATOR_HEIGHT,
 	position: 'absolute',
 	top: 0,
-	width: '100%',
+	width: 'calc(100% + 1.5px)',
```
JonnyBurger, line 58: This is to compensate for the splitter which is 3px wide. Maybe not needed

PR #8266, opened 2026-06-09
fix: lower docs build heap to avoid OOM

### LINKED ISSUE #8242: Building docs is extremely compute extensive
Building the docs currently only works on Vercel Build machines with 16GB of RAM. 

On machines with less compute, we get OOM.

There are 2 directions to investigate:

1. Twoslash type checking might have a memory leak.
2. We give Docusaurus too much energy, taking it away from Rspack.

Advice from Sebastien Lorber:

> FYI we have 1400 docs on our english website and can build with 200mb of nodejs memory
NODE_OPTIONS="--max-old-space-size=200" pnpm --filter website docusaurus build --locale en
However this doesn't include Rspack which is not really nodejs memory
maybe you should actually reduce the memory used by node so that the rust side has more memory available?
I'm not a memory expert and it probably depends on the system it runs on, but worth giving it a try

mvanhorn:
## Summary

Lowers the Node heap cap on the docs `build-docs` script from `--max-old-space-size=2048` to `512` so the Docusaurus/Rspack build no longer OOMs on lower-memory machines.

## Why this matters

Per #8242, the docs build only reliably succeeds on ~16GB machines today. The root cause was traced to the Rspack build phase, not a Twoslash leak. With the current 2048MB heap the build peaked at ~5.1GB RSS, while a run with a much smaller heap cap dropped peak RSS to ~3.1GB. A smaller Node heap forces earlier GC and leaves more system memory for the Rust/Rspack side (which isn't counted in Node's heap). 512MB is a conservative middle ground between the 200MB experiment and the prior 2048MB.

Fixes #8242

### DIFF
--- a/packages/docs/package.json
+++ b/packages/docs/package.json
@@ -10,7 +10,7 @@
 		"format": "oxfmt src",
 		"docusaurus": "docusaurus",
 		"start": "bun copy-raw-docs.ts && bun fetch-prompt-submissions.ts && bun update-prompt.ts && cd .. && bun run build && cd docs && docusaurus start --host 0.0.0.0",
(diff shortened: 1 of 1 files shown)


JonnyBurger:
No, I just increased it because that led to OOM, but thank you

PR #8327, opened 2026-06-11
`@remotion/studio`: Preselect filename stem in rename dialog

### LINKED ISSUE #8318: Asset rename dialog should pre-select part before dot

Aditya07771:
## What does this fix?

Fixes #8318

The rename dialog does not pre-select the filename correctly when it opens. The user has to manually position the cursor before editing.

## What did I change?

* RenameStaticFile: on mount, call `setSelectionRange(0, stemEnd)` where `stemEnd` is the index of the last dot (or the full length if there is no dot). This selects `my-video` and leaves `.mp4` unselected.
* RenameFolder / RenameComposition: call `input.select()` on mount since folder names and composition IDs carry no extension.

## How to verify

1. Right-click any asset in Remotion Studio → Rename.
2. The dialog opens with `my-video` highlighted (not `my-video.mp4`).
3. Start typing to replace only the stem.
4. `.mp4` is preserved unchanged.


JonnyBurger:
Thank you!

PR #8362, opened 2026-06-12
`@remotion/studio`: Align outline toggle icons

### LINKED ISSUE #8309: Toggle outline icons does not align
Probably triggering for a lot of people

Image

Aditya07771:
## What does this fix?

Fixes #8309

The outline toggle icon was not visually aligned with the other toolbar controls.

## What did I change?

- Centered the contents of the shared `ControlButton`
- Kept the fix minimal so it applies consistently to the toolbar toggles

## How to verify

1. Open Remotion Studio.
2. Compare the outline toggle with the nearby toolbar buttons.
3. Confirm the icon is vertically centered.

### DIFF
--- a/packages/studio/src/components/ControlButton.tsx
+++ b/packages/studio/src/components/ControlButton.tsx
@@ -15,6 +15,8 @@ export const ControlButton = (
 		return {
 			opacity: props.disabled ? 0.5 : 1,
 			display: 'inline-flex',
+			alignItems: 'center',
+			justifyContent: 'center',
 			background: 'none',
 			border: 'none',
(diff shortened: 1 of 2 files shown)


JonnyBurger:
Thanks for the fix! I pushed a small refinement on top: the control buttons now use a fixed 30x30 hit target so the toolbar items line up consistently regardless of the SVG intrinsic size. I also made the outline icon itself 18x18, which matches the neighboring fullscreen icon better optically. The original flex centering helped, but the underlying mismatch was that the 16px outline/checkerboard buttons and the 18px fullscreen button ended up with different overall button sizes.

PR #8378, opened 2026-06-13
`@remotion/studio`: Add selectable guides

### LINKED ISSUE #8343: Guides should become a new type of selectable element
In the Studio we have guides and we have selectable elements.
Clicking a guide should select it and unselect other guides.

No mixed selection for guides is allowed, and only a single guide may be selected.

The backspace handler should delete the guide. No entry into the undo stack.

When a guide is selected, it should turn blue.

JonnyBurger:
## Summary

- Add guides as a `TimelineSelection` variant so they share the existing Studio selection model
- Keep guide dragging as separate interaction state while preserving selected guide highlighting
- Delete selected guides with Backspace/Delete without going through timeline undo helpers
- Guard timeline-only helpers from guide selections and add selection tests

Closes #8343.

## Test plan

- `bun test packages/studio/src/test/timeline-selection.test.ts`
- `bun run build`
- `bun run stylecheck`


JonnyBurger:
Superseded by #8381, recreated from a clean main base to remove unrelated changes.

PR #8385, opened 2026-06-13
Docs: Update jump cuts snippet to recommend premounting pattern

### LINKED ISSUE #8012: Jump Cuts example is not state of the art anymore
http://remotion.dev/docs/miscellaneous/snippets/jumpcuts

We now recommend premounting more Video tags

mvanhorn:
## Summary

The jump cuts snippet now leads with the recommended premounting approach: using `<Series>` with `premountFor` and a separate `<OffthreadVideo>` per section. This is the pattern JonnyBurger flagged as current best practice in the issue. The original single-tag approach is kept as a secondary option for small jumps, with a note explaining when it applies. Added links to the premounting and "Playing videos in sequence" docs in the See also section.

Fixes #8012

### DIFF
--- a/packages/docs/docs/miscellaneous/snippets/jumpcuts.mdx
+++ b/packages/docs/docs/miscellaneous/snippets/jumpcuts.mdx
@@ -7,13 +7,14 @@ crumb: 'Snippets'
 
 Sometimes you want to implement a "jump cut" to skip parts of a video (for example to cut out the "uhm"s).
 
-## Recommended: Pre-mounting multiple video tags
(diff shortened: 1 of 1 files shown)


JonnyBurger:
Thanks a lot! 🙌

mvanhorn:
Appreciate the docs update, @JonnyBurger. Leading the jump-cuts snippet with the premounting pattern should steer people to the better approach.

JonnyBurger:
Thanks @mvanhorn! Agreed, leading with the premounting pattern should make the recommended approach much clearer for people running into jump cuts.

PR #8409, opened 2026-06-13
Fix typos on comapny and investors pages

MehmetAdemi:


MehmetAdemi:
## Vercel preview build failure

The preview deployment is failing because the **Docusaurus docs build** (`docs#build-docs`) crashes during static site generation — not because of the typo/content changes in this PR.

### What fails

`docusaurus build` pre-renders every docs page on the server. Two pages fail with the same error:

- `/docs/api`
- `/docs/shapes/`

```
TypeError: Cannot read properties of undefined (reading 'width')
```

The React SSR stack trace in the Vercel log (`renderNodeDestructive`, etc.) is a symptom; the real error is above it.

### Root cause

Both pages render the shapes table of contents from `packages/docs/docs/shapes/table-of-contents.tsx`. That component draws a thumbnail for each shape via `ShapeDragPreview`, which calls `makeDefaultShapeInfo(shape)` and then reads `shapeInfo.width`.

A new **`Callout`** shape was recently added to `shapeComponents` in `shapes-info.tsx` (commit `d0dafec429`: “Add callout shape”), but **`makeDefaultShapeInfo` was never updated** to handle it. When the TOC maps over all shapes (including `Callout`), the switch falls through without a matching case, the function returns `undefined`, and accessing `.width` throws during SSR.

`/docs/api` fails for the same reason — the API overview page includes `<ShapesTableOfContents />`, which uses the same component.

### Fix

Add a `Callout` case to `makeDefaultShapeInfo` in `table-of-contents.tsx` (import `makeCallout` and wire its props). Also add `Callout` to `shape-drag-preview.tsx` for consistency.

I reproduced this locally with `bun run build-docs` in `packages/docs`; after adding the missing `Callout` case, the build completes successfully.

MehmetAdemi:
## How this was solved

The failing Vercel preview was caused by the docs build crashing during Docusaurus static site generation. The affected pages were `/docs/api` and `/docs/shapes/`, both of which render the shapes table of contents.

The immediate issue was that `Callout` had been added to the shared `shapeComponents` list, but the thumbnail/preview generation code still had explicit shape switch statements that did not handle `Callout`. During SSR, `makeDefaultShapeInfo('Callout')` returned `undefined`, and the table of contents then tried to read `shapeInfo.width`, causing:

```text
TypeError: Cannot read properties of undefined (reading 'width')
```

The fix adds `Callout` support to the docs shape preview paths:

- `packages/docs/docs/shapes/table-of-contents.tsx` now imports `makeCallout` and handles `case 'Callout'` in `makeDefaultShapeInfo()`.
- `packages/docs/components/shapes/shape-drag-preview.tsx` now includes `Callout` in the supported shape names and generates the preview via `makeCallout()`.

I verified the fix locally by running `bun run build-docs` in `packages/docs`; the docs build now completes successfully.

## More robust follow-up

The current fix is pragmatic and correct for the preview failure, but the underlying fragility is that shape support is duplicated across multiple registries/switch statements. This allowed `Callout` to be added in one place while preview generation remained incomplete.

A more robust follow-up would be to centralize shape preview metadata/generation so adding a new shape only requires updating one source of truth. At minimum, these switches should be made exhaustive with a `never` check so TypeScript fails at compile time when a `ShapeName` is not handled. That would prevent this exact SSR-only failure from reaching Vercel again.

PR #8420, opened 2026-06-14
`@remotion/studio`: Detect APNG and import as AnimatedImage

### LINKED ISSUE #8388: `animated-png.png` is dropped as `<Img>` even though `<AnimatedImage>` would be more appropriate

Aditya07771:
Closes #8388

When an animated PNG (APNG) is imported or dropped into Studio, it is currently inserted as ``, which does not preserve animation. This change detects APNG files via the PNG `acTL` chunk and inserts them as `<AnimatedImage>` instead.

Plain PNG files continue to be inserted as `` and are unaffected.

## Changes

* Added APNG detection in file type resolution by checking for the `acTL` chunk before falling back to standard PNG handling.
* Added a new `animated-image` asset type for insertable composition elements.
* Mapped APNG imports to `<AnimatedImage>` from `remotion`.
* Updated Studio asset import labeling to display `<AnimatedImage>` for APNG assets.
* Updated JSX insertion logic to generate the correct import and component.
* Updated remote asset handling to classify APNG files as animated images.
* Added and updated tests covering APNG detection, asset insertion, and JSX generation.

## Notes

Extension-only paths (for example, dragging an existing `.png` asset from the Asset panel) still resolve to ``, since only the file extension is available in that flow and the file contents cannot be inspected to determine whether the PNG is animated.

The fix applies to import/drop flows where file bytes are available and APNG detection can be performed reliably.


JonnyBurger:
Thanks for the contribution!

After testing this again against the originally reported behavior, we could not verify that this PR actually fixes the bug. The changes improve APNG detection/mapping in some import paths, but the reported scenario was not shown to be solved by this PR as submitted. To avoid merging an unverified or incomplete fix, I’m closing this PR.

If there is a minimal reproduction that demonstrates the original issue being fixed end-to-end, we can revisit with a new PR targeted at that specific path.

PR #8527, opened 2026-06-16
`@remotion/studio`: Add inspector sequence rename

JonnyBurger:
## Summary

- Adds inline sequence name editing to the Sequence inspector header.
- Shows a pencil affordance beside the title only when the same rename conditions used by the timeline are met.
- Shares sequence display/editability helpers with the timeline rename path.

## Testing

- bun run test (packages/studio)
- bun run make (packages/studio)
- bun run build
- bunx turbo run lint formatting --filter=@remotion/studio --no-update-notifier
- Browser smoke test on packages/example Studio at localhost:3007: selected a sequence, opened inspector rename input, canceled with Escape.

## Note

- bun run stylecheck was also run, but it failed due to pre-existing unstaged edits in packages/example/src/Empty.tsx adding unused imports. Those files are not part of this PR.

### DIFF
--- a/packages/studio/src/components/InspectorPanel/SequenceSelectionInspector.tsx
+++ b/packages/studio/src/components/InspectorPanel/SequenceSelectionInspector.tsx
@@ -1,13 +1,26 @@
-import React, {useCallback, useContext, useMemo} from 'react';
+import React, {
+	useCallback,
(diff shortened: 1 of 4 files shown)


JonnyBurger:
Closing per request.

PR #8536, opened 2026-06-17
`@remotion/lambda`: allow custom s3 provider regions

suzunn:
## Problem

Custom S3-compatible output providers can require region values that are not in the built-in AWS region union. Tigris, for example, documents `auto`, but `s3OutputProvider.region` was typed as the Lambda provider region and forced users to suppress TypeScript even though the AWS SDK client accepts a plain string for custom endpoints.

## Solution

I widened the custom credential region type while keeping the provider region relationship intact, and adjusted the Lambda service client helper so custom S3 clients can be keyed with that string region. I also added coverage that preserves a custom provider region such as `auto` when resolving the expected output name and credentials.

## Testing

- `oxfmt packages/serverless-client/src/constants.ts packages/lambda-client/src/get-service-client.ts packages/serverless-client/src/test/expected-out-name.test.ts --check`
- `git diff --check`

I also attempted the targeted Bun test and a serverless-client typecheck, but the local monorepo install could not complete in this Windows environment. `bun install --frozen-lockfile` timed out after 15 minutes, leaving workspace packages unresolved, so those commands failed on missing local modules like `@remotion/renderer/pure` and `remotion/no-react` before exercising this change.

### DIFF
--- a/packages/lambda-client/src/get-service-client.ts
+++ b/packages/lambda-client/src/get-service-client.ts
@@ -29,7 +29,7 @@ const getCredentialsHash = ({
 	forcePathStyle,
 	requestHandler,
 }: {
-	region: AwsRegion;
+	region: string;
 	customCredentials: CustomCredentials<AwsProvider> | null;
 	service: keyof ServiceMapping;
 	forcePathStyle: boolean;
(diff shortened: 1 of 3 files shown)


JonnyBurger:
Thank you for the contribution!

PR #8589, opened 2026-06-20
`remotion`: Cancel in-progress prefetch downloads on `free()`

12ya:
## Summary

- Abort prefetch requests even after response headers have arrived
- Cancel the active response body reader to stop downloading
- Reject `waitUntilDone()` when `free()` cancels the prefetch
- Prevent creating a Blob URL from a partial response
- Make repeated cancellation idempotent

## Testing

Added regression tests covering cancellation:

- Before receiving the response
- While reading the response body

Tests were not run locally because dependencies are unavailable and the installed Bun version is 1.0.0 instead of the required 1.3.3.

## Documentation

No documentation change is required. The existing `free()` documentation already states that an ongoing prefetch is aborted; this PR makes the implementation match that behavior.

## Tradeoffs

Downloaded partial data is discarded. A subsequent `prefetch()` call starts a new request rather than resuming the previous download.

### DIFF
--- a/packages/core/src/prefetch.ts
+++ b/packages/core/src/prefetch.ts
@@ -145,16 +145,16 @@ export const prefetch = (
 		resolve = res;
 		reject = rej;
 	});
+	void waitUntilDone.catch(() => undefined);
 
 	const controller = new AbortController();
-	let canBeAborted = true;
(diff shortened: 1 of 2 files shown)


JonnyBurger:
Thank you! 🙌

PR #8592, opened 2026-06-20
`remotion`: Add resumable prefetch API

12ya:
## Summary

Adds `resumablePrefetch()`, a client-side prefetch API supporting:

- Pausing an active download while retaining downloaded chunks in memory
- Resuming with HTTP `Range` and `If-Range` headers
- Freeing partial chunks or the completed Blob URL
- Progress reporting through `onProgress`
- Automatic use of the completed asset by Remotion media components

## Safety and fallback behavior

- Uses strong `ETag` or `Last-Modified` validators to prevent combining different asset versions.
- Restarts from byte zero when validators are unavailable.
- Restarts when the server ignores the range request and responds with `200`.
- Validates `206 Partial Content` and `Content-Range` responses.
- Rejects malformed or inconsistent partial responses.

## Documentation

Adds a dedicated `resumablePrefetch()` documentation page covering usage, options, lifecycle methods, server requirements, CORS, and compatibility.

## Tests

Added unit coverage for:

- Resuming a validated partial download
- Servers that ignore range requests
- Assets without suitable validators

## Tradeoffs

Partial chunks are retained in browser memory, increasing memory usage until completion or `free()`.

Downloads cannot resume across page reloads. Servers must support range requests and expose the required headers through CORS; otherwise, downloads safely restart from the beginning.


JonnyBurger:
Thanks!

`prefetch()` is not best practice anymore, premounting works better (https://www.remotion.dev/docs/troubleshooting/player-flicker)

I don't wanna get more people to use it by adding a new API, I would rather that premounting gets used correctly more often. But I will get your other PR merged, thanks a lot!

PR #8631, opened 2026-06-23
`@remotion/media`: Fix scrubbing wedging the audio scheduler

exire8:
## Problem
When a task of the audio sample scheduler was in flight and its source went stale (e.g. the user scrubbed away before it resolved), `processNext` would bail out early at the `running >= CONCURRENCY` guard. Because the in-flight task could never resolve for work nobody needed anymore, the single slot was never released and every subsequent waiter queued behind it forever, leading to an infinite loading state.

Here's a recording reproducing the problem in Remotion Studio:

https://github.com/user-attachments/assets/137805c0-5624-4833-ae96-2b87a3c3cb82

This is the code used for setting up the above composition:

```Jsx
import React from 'react';
import { Video } from '@remotion/media';
import { AbsoluteFill, Composition, Sequence } from 'remotion';

const VIDEOS = [
	'https://res.cloudinary.com/demo/video/upload/dog.mp4',
	'https://res.cloudinary.com/demo/video/upload/outdoors.mp4'
];

const ScrubChurn: React.FC = () => (
	<AbsoluteFill>
		{VIDEOS.map((src, i) => (
			<Sequence key={i} from={i * 30} durationInFrames={30} premountFor={15}>
				<Video src={src} objectFit="cover" />
			</Sequence>
		))}
	</AbsoluteFill>
);

export const Root: React.FC = () => (
	<Composition
		id="ScrubChurn"
		component={ScrubChurn}
		durationInFrames={VIDEOS.length * 30}
		fps={30}
		width={1280}
		height={720}
	/>
);
```

## Fix
`processNext` now checks whether the currently running entry has gone stale. If so, it frees the slot immediately so a fresh waiter can run instead of deadlocking.

### DIFF
--- a/packages/media/src/audio/sort-by-priority.ts
+++ b/packages/media/src/audio/sort-by-priority.ts
@@ -16,14 +16,18 @@ const CONCURRENCY = 1;
 
 const waiters: Waiter[] = [];
 let running = 0;
-let runningEntry: {waiter: Waiter; settle: () => void} | null = null;
+let runningEntry: {
+	waiter: Waiter;
+	cancel: () => void;
+	settle: () => void;
+} | null = null;
 
 export const processNext = (): void => {
 	if (running >= CONCURRENCY) {
 		if (runningEntry?.waiter.getPriority() === null) {
 			// Running entry went stale: free its slot so a fresh waiter can run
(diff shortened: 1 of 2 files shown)


JonnyBurger:
Thanks! 🙌

PR #8639, opened 2026-06-23
feat: auto-merge Sequence + Video into one timeline row with CapCut-like trimming

yantan826:
## Summary

When a user-authored `<Sequence>` wraps exactly one `<Video>` or `<Audio>` child, the Studio now **merges them into a single timeline row** (the purple media row) — the blue container row is hidden. This makes trimming feel like CapCut: one row per clip, and dragging edges works even without explicit `trimBefore`/`trimAfter` props.

### Row merge (automatic)
- `calculateTimeline()` now returns `{tracks, mergedParentIds}`
- Merged parents are filtered out by `shouldShowTrackInTimeline` and don't contribute depth indentation
- The surviving media row carries `mergedParentInfo` linking to the parent Sequence's `TSequence` and `nodePathInfo`

### Inspector — Container section
- Clicking the merged row shows **two sections** in the inspector: "Video Clip" (media props) and "Container" (transform: scale/translate/rotate/from/durationInFrames)
- Edits in either section write to the correct source code element

### CapCut-like trimming
- `canUpdateTrimAfter` and `canUpdateTrimBefore` now return `true` when the prop isn't set — allowing the Studio to **insert** `trimAfter`/`trimBefore` as new props on drag
- Same change in `TimelineSequence.tsx` so drag handles render even without existing trim props
- Right-edge drag on merged rows also syncs the parent Sequence's `durationInFrames` when it's a static value
- Left-edge drag similarly extended

### Sync drag handles on merged rows
- `TimelineSequenceMediaRightEdgeDragTarget` gains optional `parentNodePath`/`parentInitialDuration` fields
- Drag overrides and commit changes handle both the media and parent nodePaths in a single operation
- If the parent's `durationInFrames` is computed/keyframed, the media trim still works — just without the container sync

## Testing
- All 325 existing tests pass
- 2 new tests: merge detection, non-merge for multi-child
- Updated 2 tests to reflect new CapCut-like behavior (drag works even when trim props aren't set)
- Manual testing confirmed with Playwright: right-edge drag updates `trimAfter` via API

## Files changed
```
20 files changed, 818 insertions(+), 103 deletions(-)
```

### DIFF
--- a/packages/example/src/VideoTesting/index.tsx
+++ b/packages/example/src/VideoTesting/index.tsx
@@ -26,7 +26,14 @@ export const VideoTesting: React.FC<{
 
 	return (
 		<div>
-			<Sequence durationInFrames={durationInFrames}>
+			<Sequence
+				durationInFrames={durationInFrames}
+				from={13}
+				style={{
+					rotate: '-37.7deg',
+					scale: 0.662,
+				}}
+			>
 				<Comp src={codec === 'mp4' ? videoMp4 : videoWebm} />
 			</Sequence>
 		</div>
--- a/packages/example/src/VideoTesting/trim.tsx
+++ b/packages/example/src/VideoTesting/trim.tsx
@@ -13,19 +13,30 @@ export const VideoTestingTrim: React.FC<{
 	const {durationInFrames} = useVideoConfig();
 	const videoMp4 = staticFile('framermp4withoutfileextension');
 
-	let Comp;
-	if (type === 'codec') {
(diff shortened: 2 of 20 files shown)


JonnyBurger:
That's too much change in complexity for this type of change for my taste, sorry.
There is already an easy way to do this: Add `showInTimeline={false}` to the parent Sequence!

Plus, `<Video>` is now a sequence anyway! So you can add props like `from`, `durationInFrames` now 😊

PR #8642, opened 2026-06-24
fix: calculateMetadata() encoding defaults now override remotion.config.ts in CLI

### LINKED ISSUE #8587: `@remotion/cli`: calculateMetadata() encoding defaults should override remotion.config.ts
## Summary

`calculateMetadata()` can return encoding-related defaults such as `defaultVideoImageFormat`, `defaultPixelFormat`, and `defaultProResProfile`, but these values are not consistently applied when rendering via the CLI.

In practice, `remotion.config.ts` (and hardcoded CLI fallbacks) take precedence over composition-level defaults returned from `calculateMetadata()`. This makes it easy to accidentally render without transparency even when a composition explicitly requests ProRes 4444 + PNG + alpha pixel format.

We hit this while rendering separate transparent layers for canvas capture promo footage in `packages/brand`.

## Expected behavior

When rendering with `npx remotion render`, encoding defaults from `calculateMetadata()` should override global config defaults, but still lose to explicit CLI flags.

Suggested priority order:

1. Explicit CLI flags (`--image-format`, `--pixel-format`, `--prores-profile`, etc.)
2. `remotion.config.ts`
3. `calculateMetadata()` defaults (`defaultVideoImageFormat`, `defaultPixelFormat`, `defaultProResProfile`, etc.)
4. Built-in hardcoded defaults

This would mirror how `defaultCodec` already works today via `compositionCodec` in `videoCodecOption.getValue()`.

## Actual behavior

A composition can return transparent ProRes defaults from `calculateMetadata()`:

```tsx
return {
  defaultCodec: 'prores',
  defaultProResProfile: '4444',
  defaultPixelFormat: 'yuva444p10le',
  defaultVideoImageFormat: 'png',
  // ...
};
```

But if `remotion.config.ts` contains:

```ts
Config.setVideoImageFormat('jpeg');
```

then `remotion render` still uses JPEG frames and renders with a black page background instead of transparency.

Observed output without explicit CLI flags:

- Codec: `prores` (this part works — likely via `defaultCodec`)
- Profile: `HQ`
- Pixel format: `yuv422p10le` (no alpha)

Explicit CLI flags were required to get transparency:

```bash
bunx remotion render canvas-capture-promo \
  --image-format=png \
  --pixel-format=yuva444p10le \
  --codec=prores \
  --prores-profile=4444
```

That produced ProRes 4444 with alpha (`yuva444p12le`).

## Why this seems to happen

`defaultCodec` is wired into option resolution (`videoCodecOption.getValue()` accepts `compositionCodec`), but the other calculateMetadata encoding defaults do not appear to participate in the CLI render path.

Relevant areas to inspect:

- `packages/cli/src/image-formats.ts` — `getVideoImageFormat()` checks UI/config and then hardcodes `'jpeg'` for ProRes; it does not consult composition defaults
- `packages/cli/src/render.tsx` — resolves `pixelFormat` and `proResProfile` before composition config is available, without composition fallback
- `packages/renderer/src/options/video-image-format.tsx`
- `packages/renderer/src/options/pixel-format.tsx`
- `packages/renderer/src/options/prores-profile.tsx`

Note: `render-media.ts` does contain fallback logic like:

```ts
compositionWithPossibleUnevenDimensions.defaultVideoImageFormat ??
DEFAULT_VIDEO_IMAGE_FORMAT
```

but the CLI appears to pass resolved config/default values in earlier, preventing those composition defaults from being used.

## Reproduction

1. Create a composition whose `calculateMetadata()` returns transparent ProRes defaults (`defaultCodec: 'prores'`, `defaultProResProfile: '4444'`, `defaultPixelFormat: 'yuva444p10le'`, `defaultVideoImageFormat: 'png'`)
2. Set `Config.setVideoImageFormat('jpeg')` in `remotion.config.ts`
3. Render via CLI without explicit encoding flags
4. Inspect output with `ffprobe`

Expected: ProRes 4444 with alpha pixel format.

Actual: ProRes HQ / `yuv422p10le`, and JPEG-based frame capture (black background outside content).

## Impact

- Compositions that declare transparent export settings in `calculateMetadata()` are silently ignored
- Docs/examples suggesting calculateMetadata defaults for transparent ProRes do not work reliably with global config present
- Users must discover and manually pass CLI flags

## Suggested investigation

- [ ] Audit option resolution priority for all calculateMetadata encoding defaults
- [ ] Align CLI behavior with `defaultCodec` precedence model
- [ ] Confirm Studio render modal behavior is consistent (it may already read some composition defaults)
- [ ] Add regression test: config jpeg + calculateMetadata png/prores4444/yuva → render uses composition defaults unless CLI overrides
- [ ] Update docs if current precedence is intentional

## References

- Transparent videos docs: https://www.remotion.dev/docs/transparent-videos/
- `CalcMetadataReturnType` in `packages/core/src/Composition.tsx`

aryansingh1718:
Fixes #8587

## Problem

When a composition's `calculateMetadata()` returns encoding defaults (`defaultPixelFormat`, `defaultProResProfile`, `defaultVideoImageFormat`), these values were being silently ignored by the CLI renderer. Instead, `remotion.config.ts` settings or hardcoded fallbacks would win, causing compositions requesting transparent ProRes 4444 to render without alpha channel.

## Investigation

Confirmed that **Remotion Studio's render modal already handles this correctly** — it seeds `pixelFormat`, `proResProfile`, and `videoImageFormat` state from `resolvedComposition.default*` values before falling back to config/defaults. The bug was CLI-specific.

### `packages/cli/src/render.tsx`
Instead of passing the resolved (possibly hardcoded default) value for `pixelFormat` and `proResProfile` into `renderVideoFlow`, now checks if the source is `'default'` and passes `null`/`undefined` in that case, allowing composition defaults to take effect downstream.

### `packages/cli/src/render-flows/render.ts`
After `config` is resolved from `getCompositionWithDimensionOverride()`, applies the correct priority chain:
- `pixelFormat ?? config.defaultPixelFormat ?? 'yuv420p'`
- `proResProfile ?? config.defaultProResProfile ?? 'hq'`
- `uiImageFormat ?? config.defaultVideoImageFormat ?? null` (passed into `getVideoImageFormat`)

### `packages/cli/src/test/pixel-format.test.ts`
Added tests verifying `source` tracking behavior — `'default'` when nothing is explicitly set, `'config'` when explicitly configured, `'cli'` when flag is passed.

## Verification

Before fix: `profile=HQ`, `pix_fmt=yuv422p10le` (no alpha)

After fix: `profile=4444`, `pix_fmt=yuva444p12le` (alpha present)


JonnyBurger (changes requested):
PR contains unrelated changes and merge conflicts, can you fix? Thank you!

PR #8653, opened 2026-06-25
`@remotion/web-renderer`: Fix SVG control character rendering

### LINKED ISSUE #8650: "Failed to convert SVG to image" on `<Lottie>` with animated text
# [web-renderer] `renderMediaOnWeb`/`renderStillOnWeb` fails with "Failed to convert SVG to image" on `<Lottie>` with animated text

**Package:** `@remotion/web-renderer`
**Remotion version:** 4.0.479 (client-side rendering, experimental)
**Renderer:** client-side (`renderMediaOnWeb` / `renderStillOnWeb`)

## Summary

When client-side rendering a composition that contains a `<Lottie>` (`@remotion/lottie`) animation with an **animated text layer**, the render fails with:

```
Failed to convert SVG to image
```

The same composition renders perfectly with server-side rendering (`@remotion/renderer` / Lambda), so this is specific to the web-renderer's SVG capture path.

## Root cause

`@remotion/lottie` renders via `lottie-web`'s **SVG renderer**. For animated text, `lottie-web` writes the text into an `aria-label` attribute using the **control character U+0003 (ETX)** as a separator between glyphs, e.g.:

```
aria-label="B␃R␃A␃V␃O"   // ␃ = U+0003
```

In `web-renderer`, `turnSvgIntoDrawable()` serializes the live SVG and loads it as an ``:

```js
// packages/web-renderer ... turnSvgIntoDrawable
const svgData = new XMLSerializer().serializeToString(svg);
const url = `data:image/svg+xml;charset=utf-8,${encodeURIComponent(svgData)}`;
const image = new Image();
image.onerror = () => reject(new Error("Failed to convert SVG to image"));
image.src = url;
```

`U+0003` (and other C0 control chars) are **not valid characters in XML 1.0**. The live DOM tolerates them, but once the SVG is serialized into a standalone `data:image/svg+xml` document and parsed by the image loader, parsing fails:

```
error on line 1 at column …: invalid character in attribute value
```

→ `image.onerror` fires → "Failed to convert SVG to image".

This only affects the web-renderer because it round-trips the SVG through serialization + an ``; server-side rendering screenshots the live DOM directly and is unaffected.

## Reproduction

1. Compose a `<Lottie>` containing an animated **text** layer (any Lottie with text works; the text becomes `<text aria-label="…">` with U+0003 separators).
2. Render it with `renderMediaOnWeb()` (or `renderStillOnWeb()`).
3. Render fails at the first frame where the text layer is visible with "Failed to convert SVG to image".

(Verified by frame-bisecting a real composition: frames before the text layer appears convert fine; from the frame the animated text appears onward, conversion fails. Scanning the serialized SVG shows `U+0003` inside `aria-label` values.)

## Proposed fix

Strip invalid XML 1.0 control characters from the serialized string before building the data URL, in `turnSvgIntoDrawable()`:

```diff
- const svgData = new XMLSerializer().serializeToString(svg);
+ const svgData = new XMLSerializer()
+   .serializeToString(svg)
+   .replace(/[\u0000-\u0008\u000B\u000C\u000E-\u001F]/g, "");
```

This keeps the only C0 characters allowed in XML (tab U+0009, LF U+000A, CR U+000D) and removes the rest. It fixes all Lottie animations containing text, with no visual change (the stripped chars are non-printing separators inside `aria-label`).

We've been running this exact change as a `patch-package` patch on 4.0.479 and it fully resolves the issue.

## Environment notes

- `@remotion/lottie` only supports the SVG renderer, so there's no canvas-renderer workaround.
- Happy to open a PR with the one-line change + a test fixture (a Lottie with a text layer) if useful.

JonnyBurger:
## Summary

- Strip XML-invalid control characters before rasterizing serialized SVGs in the web renderer
- Add a regression fixture for Lottie-like SVG text aria-labels containing U+0003 separators

Fixes #8650.

## Testing

- `bunx vitest src/test/issue-8650-svg-control-character.test.tsx --browser --run`
- `bun run build`
- `bun run stylecheck`


JonnyBurger:
Closing this duplicate since #8651 landed the same fix for #8650. Thanks!

PR #8659, opened 2026-06-25
fix: preserve collapsed effect rows on selection

### LINKED ISSUE #8548: Effect collapsed state is not preserved anymore

pupuking723:
Fixes #8548.

Selecting an effect prop now expands the sequence row and the Effects container, but no longer clears the collapsed state for the specific effect row. Regular nested sequence props keep the previous parent-expansion behavior.

Verification:
- `PATH=/Users/mac/.bun/bin:$PATH bun run build`
- `PATH=/Users/mac/.bun/bin:$PATH bun test packages/studio/src/test/get-expanded-track-parent-keys.test.ts packages/studio/src/test/timeline-selection.test.ts`
- `PATH=/Users/mac/.bun/bin:$PATH bun run stylecheck`


JonnyBurger:
Thank you!

PR #8672, opened 2026-06-25
`@remotion/gif`: Reuse cached GIF instead of re-decoding on every remount

nvolungis:
## Problem

`GifForDevelopment` reads the decoded GIF from `volatileGifCache` / `manuallyManagedGifCache` in its `useState` initializer, but the `useEffect` then **re-runs the decode worker unconditionally on every mount**, even when that GIF is already cached. Each parse rebuilds an `ImageData` per frame.

In the Player/Studio this bites during **seeking/scrubbing**: an overlay's `<Sequence>` mounts and unmounts as the playhead crosses its boundary, so `<Gif>` remounts repeatedly and re-decodes the whole GIF each time. The freshly-allocated frame sets churn memory, and over a long editing session with GIF overlays the heap can grow until the renderer is OOM-killed. (Found in a video editor built on Remotion: scrubbing over a GIF overlay would peg CPU on GC and crash the tab; a heap diff showed `ImageData` / `Uint8ClampedArray` / `ArrayBuffer` growth, and instrumentation confirmed the worker was re-spawned on every remount.)

## Fix

If the GIF is already in either cache, reuse it (mirroring the `useState` initializer) and skip the worker:

```ts
useEffect(() => {
  const parsedGif =
    volatileGifCache.get(cacheKey) ?? manuallyManagedGifCache.get(cacheKey);
  if (parsedGif !== undefined) {
    update(parsedGif as GifState);
    currentOnLoad.current?.(parsedGif as GifState);
    return;
  }
  // …existing parse path
}, [cacheKey, resolvedSrc]);
```

## Test

Adds a test that seeds the cache and asserts no `Worker` is constructed when `<Gif>` mounts with the GIF already cached. Verified locally (`bun run makecore` + `bun test src/test`): red without the guard (worker spawned), green with it; the existing test is unaffected; `oxfmt --check` and `eslint` pass.

## Tradeoff / open question

On a cache hit `onLoad` now fires synchronously inside the effect rather than asynchronously after a worker round-trip. If you'd rather preserve the async timing, the cache branch can resolve through a microtask (`Promise.resolve(parsed).then(...)` with an abort guard) — happy to switch. Also open to *not* firing `onLoad` on reuse if you consider it a one-time event.

### DIFF
--- a/packages/gif/src/GifForDevelopment.tsx
+++ b/packages/gif/src/GifForDevelopment.tsx
@@ -61,6 +61,14 @@ export const GifForDevelopment = forwardRef<
 		currentOnError.current = onError;
 
 		useEffect(() => {
+			const parsedGif =
+				volatileGifCache.get(cacheKey) ?? manuallyManagedGifCache.get(cacheKey);
+			if (parsedGif !== undefined) {
+				update(parsedGif as GifState);
+				currentOnLoad.current?.(parsedGif as GifState);
+				return;
+			}
+
 			let done = false;
 			let aborted = false;
 			const {prom, cancel} = parseWithWorker({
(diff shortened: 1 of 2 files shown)


JonnyBurger:
Makes sense! Thanks a lot!

PR #8673, opened 2026-06-26
`@remotion/studio`: Preserve collapsed effect sections on selection

### LINKED ISSUE #8548: Effect collapsed state is not preserved anymore

pupuking723:
Fixes #8548.

The previous fix kept individual effect rows collapsed, but selecting an effect prop could still expand the parent Effects section. This keeps the normal parent expansion behavior for regular prop rows, while effect rows and effect props only expand the parent sequence row so the user's collapsed Effects state is preserved.

Verification:
- bun test packages/studio/src/test/get-expanded-track-parent-keys.test.ts
- bun test packages/studio/src/test/timeline-selection.test.ts
- bun run formatting in packages/studio
- bun run lint in packages/studio
- git diff --check


JonnyBurger:
Thanks for the PR! I tested the change, but it does not fix the behavior for me yet, so I'm going to close this one. Appreciate you taking a look.

PR #8686, opened 2026-06-26
`@remotion/studio`: Support multiplication sequence props

### LINKED ISSUE #7378: Parse and save back simple multiplication values
Let's say

```tsx
<Sequence premountFor={2 * fps} />
```

is in the code. Right now when we call /api/subscribe-to-sequence-props, this will lead to a value of `computed`, I suppose.
It would be cool if we could:

- In the server, recognize simple multiplications
- In the client, when we subscribe, pass the actual runtime value (let's say `2 * fps` evaluates to 60)
- Make an educated guess that `fps = 30`
- Instead of `computed`, return something like `multiplication`
- Allow editing and save back

```tsx
<Sequence premountFor={2.5 * fps} />
```

## Acceptance criteria

- [ ] Interpolation input ranges using simple multiplication with a `useVideoConfig()` value must be editable in Studio, not shown as `computed`, for example:

```tsx
style={{scale: interpolate(frame, [0, 3.33 * fps], [2, 4])}}
```

JonnyBurger:
## Summary

- Parse simple multiplication expressions like `2 * fps` as editable sequence props when a numeric runtime value is available
- Pass runtime sequence values through Studio subscriptions and save edited multiplication props back as expressions
- Preserve multiplication status during optimistic Studio updates

Fixes #7378.

## Test plan

- `bun test packages/studio-shared/src/test/optimistic-update-for-prop-statuses.test.ts packages/studio-server/src/test/compute-sequence-props-status.test.ts packages/studio-server/src/test/update-sequence-props.test.ts packages/studio-server/src/test/save-sequence-props.test.ts`
- `bun run build`
- `bun run stylecheck`


JonnyBurger:
Closing this PR in favor of the standalone implementation plan posted on #7378.

PR #8713, opened 2026-06-27
`remotion-monorepo`: Upgrade to TypeScript 7 RC

JonnyBurger:
## Summary

- replace `@typescript/native-preview` with the TypeScript 7 RC compiler package
- run package builds through the `tsc` binary while keeping `typescript@5.9.3` available for legacy JS API consumers
- restore previous ambient type discovery with `types: ["*"]`
- adjust the bundler loader's TypeScript API typing for the RC package layout

## Test Plan

- `bunx oxfmt src --write` in `packages/core`
- `bunx oxfmt src --write` in `packages/bundler`
- `bun run build`
- `bunx turbo run lint formatting --filter='@remotion/bundler' --filter='remotion' --no-update-notifier`

`bun run stylecheck` was run, but this local machine is missing PHP `ext-iconv`, causing `@remotion/lambda-php#lint` to fail during `php composer.phar install`. The run reached the existing PHP environment blocker after passing the TypeScript/Prettier path that this PR changes.


JonnyBurger:
Just not necessary right now, don't wanna test it, since it was not straightforward

PR #8721, opened 2026-06-27
`@remotion/studio`: Fix context menu retriggering

### LINKED ISSUE #7755: Right clicking of layer when a context menu is already open is not good
https://github.com/user-attachments/assets/80fa5d09-2559-49d7-aa7c-3ee9e80c3d31

samohovets:
## Summary

Fixes #7755 

- Allow right-clicks to pass through an open Studio context menu overlay so another context menu target can handle them
- Prevent the native browser context menu while a Studio context menu is open
- Avoid treating secondary-button pointer downs as outside clicks, so retriggered context menus are not immediately closed on pointerup

## Testing
- `bun run build`
- `bun run stylecheck`


JonnyBurger:
@samohovets You can merge on your own when tested

PR #8723, opened 2026-06-27
`@remotion/studio`: Preserve inspector collapsed effects

### LINKED ISSUE #8548: Effect collapsed state is not preserved anymore

samohovets:
## Summary

- Preserve collapsed rows in the Studio inspector across inspector unmounts/reselection
- Store inspector collapse state separately from timeline expanded-track state
- Keep effect row identity based on the existing inspector node path info rather than effect labels

Fixes #8548.

## Verification

- `bun run build`
- `bun run stylecheck`

Note: The first `bun run stylecheck` failed because an untracked `packages/template-react-router/.react-router` generated directory contained stale types. I removed that untracked generated directory and reran `bun run stylecheck`, which passed.

### DIFF
--- a/packages/studio/src/components/InspectorSequenceSection.tsx
+++ b/packages/studio/src/components/InspectorSequenceSection.tsx
@@ -69,8 +69,50 @@ const isEffectsRoot = (
 	return auxiliaryKeys[auxiliaryKeys.length - 1] === 'effects';
 };
 
+const INSPECTOR_COLLAPSED_ROWS_SESSION_STORAGE_KEY =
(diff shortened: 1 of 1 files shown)


samohovets:
@JonnyBurger fixed! I hope I fixed the right thing? 😃  The issue description was a bit empty but I guessed it's in the inspector. I expect the effect collapsibles to preserve state.

https://github.com/user-attachments/assets/9cdc6814-191b-4040-b1ac-d7324072b39f

JonnyBurger:
yes exactly!!

PR #8725, opened 2026-06-28
`@remotion/media`: Fix `<Video>` reinitializing MediaPlayer when onError identity changes

### LINKED ISSUE #8722: @remotion/media <Video>: unstable onError prop re-initializes the MediaPlayer every frame during preview playback
# Bug Report 🐛

## Summary
Passing an inline (non-memoized) `onError` to `<Video>` from `@remotion/media`
causes the MediaPlayer to be re-initialized on **every frame** during preview
playback. Each re-init re-opens the source (`UrlSource`), producing a storm of
Range fetches + re-decodes → heavy preview stutter and audio drop-outs.
Memoizing `onError` makes it disappear.

## Expected behavior
The identity of the `onError` callback should not tear down and rebuild the
media pipeline. An unstable error handler should not affect playback.

## Current behavior
With an inline `onError`, during playback there is ~1 Range request per
displayed frame for the same file (all served from disk cache, so it's
CPU/decoder saturation, not bandwidth). The preview stutters and audio breaks up.

## Steps to reproduce
1. Render `<Video src={url} onError={(e) => report(e)} />` inside a composition
   (note: `onError` is a new function each render).
2. Play the composition in `@remotion/player`.
3. Open the Network panel: one Range request per frame for `url` (disk cache),
   plus visible stutter / audio glitches.
4. Wrap `onError` in `useCallback` → the storm and stutter disappear.

```tsx
// Reproduces (new function identity every render):
<Video src={signedUrl} onError={(e) => report(e)} />

// Fixed (stable identity):
const onError = useCallback((e: Error) => report(e), []);
<Video src={signedUrl} onError={onError} />
```
## Root cause
In the preview Video component, the MediaPlayer-initialization useEffect
includes onError in its dependency array (dist/esm/index.mjs — the effect
that calls callOnErrorAndResolve / setMediaPlayerReady). Since the
composition re-renders every frame during playback, an inline onError changes
identity each frame, so the init effect (which re-opens the source) re-runs each
frame.

## Suggested fix
Store handler-style callbacks (onError, onVideoFrame, …) in a ref and read
the latest value inside the effect, and remove them from the initialization
effect's dependency array. This keeps "always call the latest handler" while
avoiding a full media-pipeline re-init on identity change.

## Workaround
Memoize onError with useCallback on the consumer side.

## Environment
@remotion/media: 4.0.484 (remotion 4.0.484)
Path: @remotion/player preview (not rendering / not <OffthreadVideo>)
Browser: Chrome (Windows)

mvanhorn:
## Summary

Passing an inline `onError` callback to `<Video>` from `@remotion/media` no longer tears down and rebuilds the preview `MediaPlayer` on every frame. `onError` is now read through a ref instead of being a dependency of the player-initialization effect, so a changing callback identity does not restart playback.

## Why this matters

The `MediaPlayer`-init `useEffect` in `packages/media/src/video/video-for-preview.tsx` listed `onError` in its dependency array. When a caller passed a non-memoized `onError` (a new function each render), the effect re-ran every frame during preview playback: it disposed the player, recreated it, and re-opened the `UrlSource`. That produced one Range request per displayed frame plus repeated re-decodes, saturating the decoder and causing preview stutter and audio drop-outs. Memoizing `onError` with `useCallback` made the symptom disappear, which pointed at the dependency array as the root cause.

The fix mirrors the ref pattern already used in the same file for other per-render callbacks (for example `effectsRef`):

- `onErrorRef = useRef(onError)` kept current on every render.
- The three call sites inside the init effect now read `onErrorRef.current` instead of `onError`.
- `onError` is removed from the init effect dependency array.

Behavior for stable callbacks is unchanged, and the most recent `onError` is still the one invoked when an error fires. The scope is limited to the `<Video>` preview path named in the issue; the audio preview path is untouched.

## Testing

- Added `packages/media/src/test/video-onerror-stable.test.tsx`: it spies on `MediaPlayer.prototype.initialize`, renders `<Video>` with a fresh inline `onError` identity on every render across several forced re-renders, and asserts the player initializes exactly once.
- `oxfmt src --check`, `eslint src`, and a `tsgo` typecheck pass for `@remotion/media`.
- The package build (`turbo run make` for `@remotion/media`) passes.
- The vitest browser suite could not be launched in my environment (the WebdriverIO chromedriver download returned 404 for the local Chrome version), so the new test was not executed here; it follows the existing `player-muted-video.test.tsx` pattern.

Fixes #8722

### DIFF
--- a/packages/media/src/test/video-onerror-stable.test.tsx
+++ b/packages/media/src/test/video-onerror-stable.test.tsx
@@ -1,0 +1,83 @@
+import {Player} from '@remotion/player';
+import React, {useEffect, useState} from 'react';
+import {createRoot} from 'react-dom/client';
+import {expect, test, vi} from 'vitest';
+import {MediaPlayer} from '../media-player';
+import {Video} from '../video/video';
+
+const waitFor = async (predicate: () => boolean) => {
+	const started = Date.now();
+	while (Date.now() - started < 10000) {
+		if (predicate()) {
+			return;
+		}
+
+		await new Promise((resolve) => setTimeout(resolve, 100));
+	}
+
+	throw new Error('Timed out waiting for condition');
+};
+
+test('does not reinitialize MediaPlayer when onError identity changes', async () => {
+	const container = document.createElement('div');
+	document.body.appendChild(container);
+
+	const initSpy = vi.spyOn(MediaPlayer.prototype, 'initialize');
+	const root = createRoot(container);
+	let committedRerenders = 0;
+
+	const VideoComposition: React.FC = () => {
+		const [rerenders, setRerenders] = useState(0);
+
+		useEffect(() => {
+			const interval = window.setInterval(() => {
+				setRerenders((current) => {
+					if (current >= 4) {
+						window.clearInterval(interval);
+						return current;
+					}
+
+					return current + 1;
+				});
+			}, 50);
+
+			return () => window.clearInterval(interval);
+		}, []);
+
+		useEffect(() => {
+			committedRerenders = rerenders;
+		}, [rerenders]);
+
+		return <Video src="/bigbuckbunny.mp4" onError={() => {}} />;
+	};
+
+	root.render(
+		<Player
+			acknowledgeRemotionLicense
+			component={VideoComposition}
+			compositionHeight={720}
+			compositionWidth={1280}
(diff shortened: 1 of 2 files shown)


JonnyBurger:
Thanks!

PR #8727, opened 2026-06-29
`@remotion/studio`: Cut effects (Cmd+X)

### LINKED ISSUE #8581: Cmd+X should work for effects

pupuking723:
Fixes #8581.

This adds a narrow Cmd/Ctrl+X path for selected effect rows. It reuses the same effect clipboard payload used by copy, then calls the existing timeline delete flow so selection cleanup and delete behavior stay shared.

Verification:
- bun test packages/studio/src/test/timeline-selection.test.ts
- bun run lint (packages/studio; existing warnings only)
- bun run formatting (packages/studio)
- bun run make (packages/studio)

### DIFF
--- a/packages/studio/src/components/Timeline/TimelineClipboardKeybindings.tsx
+++ b/packages/studio/src/components/Timeline/TimelineClipboardKeybindings.tsx
@@ -313,6 +313,128 @@ export type EffectClipboardDataFromSelections =
 			readonly type: 'uncopyable';
 	  };
 
+export type PendingEffectCut = {
+	readonly payload: EffectClipboardData;
+	readonly sourceTargetKey: string | null;
+	readonly selectedEffects:
+		| readonly {
+				readonly effectIndex: number;
+				readonly effect: EffectClipboardSnapshot;
+		  }[]
+		| null;
+};
+
+const getPendingEffectCut = ({
+	selectedItems,
+	payload,
+}: {
+	readonly selectedItems: readonly TimelineSelection[];
+	readonly payload: EffectClipboardData;
+}): PendingEffectCut => {
+	const effectSelections = selectedItems.filter(
+		(
+			selection,
+		): selection is Extract<TimelineSelection, {type: 'sequence-effect'}> =>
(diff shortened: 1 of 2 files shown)


pupuking723:
Moved the timeline position ref sync out of render and into an effect in 72230e70.

Local verification:
- bun test packages/studio/src/test/timeline-selection.test.ts
- bun run lint (packages/studio; existing warnings only)
- bun run formatting (packages/studio)
- bun run make (packages/studio)

samohovets (commented):
I tested manually and found two issues:

   - Cmd+X then Cmd+V does not preserve effect order. In `effect-copy-source`, cutting `brightness()` and pasting it back onto the same `<Solid>` moves it after `blur()`. Since effect order can affect rendering, this
 seems incorrect.
   - Cmd+X puts the internal Remotion effects JSON onto the system clipboard, so pasting into a normal text field/chat/email produces raw JSON. See image below.

Image

I think these should be fixed before merging.

pupuking723:
Thanks for testing this. I pushed a narrower follow-up in 1995f6f1.

Cut effects are now kept in a local Studio cut buffer instead of being written to the system clipboard, so Cmd+X no longer leaves the internal JSON available to normal text fields. When pasting back into the same sequence, the cut effects are inserted at their original indices and sent as a replace payload, so effect order is preserved instead of appending to the end.

I also added a focused regression test for the same-sequence order case.

Verified locally:
- `bun test packages/studio/src/test/timeline-selection.test.ts`
- `bun run formatting` in `packages/studio`
- `bun run lint` in `packages/studio` (existing warnings only)
- `bun run make` in `packages/studio`

samohovets:
Thanks for working on this. I’m going to close this for now. 

The initial version wrote internal JSON to the system clipboard. The follow-up avoids that, but now Cmd+X does not use the system clipboard at all — the cut only lives in Studio memory. This means paste does not survive reloads, new tabs, or other Studio windows, which does not match expected clipboard behavior.

I think this needs a more careful clipboard design before we add it.

PR #8731, opened 2026-06-29
Improve `pr` skill boundaries

samohovets:
## Summary

- Clarify that the PR skill is only for opening the initial pull request
- Stop the skill early when a PR already exists for the current branch
- Make follow-up changes stay uncommitted by default unless the user asks for a Git action

## Testing

- `bun run build`
- `bun run stylecheck`


samohovets:
Didn't really work.. follow-up PR incoming

PR #8740, opened 2026-06-30
`@remotion/media`: Add ProRes support and upgrade Mediabunny to 1.50.7

samohovets:
## Summary

- Upgrade Mediabunny packages to 1.50.6
- Document optional ProRes support for `@remotion/media`
- Keep `@mediabunny/prores` user-installed, matching the existing optional AC-3/E-AC-3 flow
- Add `@mediabunny/prores` to extra package metadata for `remotion add`
- Update supported format docs and Mediabunny version docs

## Usage

```bash
npx remotion add @mediabunny/prores
```

```tsx
import {registerProresDecoder} from '@mediabunny/prores';

registerProresDecoder();
```

## Tests

- `bunx turbo run make --filter='@remotion/media'`
- `bunx turbo run lint --filter='@remotion/media'`
- `cd packages/media && bun run format`
- `cd packages/docs && bun run format`
- `cd packages/brand && bun run bundle`
- `bunx turbo run build-docs --filter='docs'`


Vanilagy:
You should consider mentioning this to users as well:

image

It's optional but perf *is* better.

JonnyBurger:
This page can be referenced, and should maybe be updated too

https://www.remotion.dev/docs/miscellaneous/cross-origin-isolation

@vanilagy we implemented the credntialless recommendation by now

Vanilagy:
Very nice

samohovets:
Do not merge yet, experiencing choppy playback in the `prores-media-video` testbed.

Not a preview-only issue, rather something is weird with the extraction speed.

Video rendered using `renderMedia()`. On the timestamp 00:03:00 the video playback visibly gets slower / delayed. 
https://github.com/user-attachments/assets/99218c46-a452-4fe2-a89e-4c7656b920eb

Same thing is happening in the preview: 

https://github.com/user-attachments/assets/6fab7956-3fe4-434c-95c4-4c99e95c0530

samohovets:
The issue appears to be caused by the composition FPS, not by iterator recreation.

`getMediaMetadata()` uses `videoTrack.computePacketStats(50).averagePacketRate`, which returns `30.00600120624245` for `video-prores.mkv`, even though FFmpeg reports the file as `30 fps`.

Because the composition FPS becomes `30.00600120624245`, frame 90 is rendered at:

```ts
90 / 30.00600120624245 = 2.9994s
```

This is just before `3.000s`, so frame extraction returns the previous media frame. In my test render, frame `90` showed the source counter as `89 / 00:02:29`.

After normalizing near-integer FPS to `30`, frame `90` correctly showed `90 / 00:03:00`.

So this does not seem to be a preview iterator recreation issue, as @JonnyBurger suggested. It reproduces in render because the render path is requesting slightly early timestamps due to the fractional FPS estimate.

samohovets:
Follow-up: The delayed frame issue was caused by the example using a slightly wrong composition FPS approximation.

`computePacketStats(50).averagePacketRate` returned `30.006...` for the ProRes test video, even though the video is nominally 30 FPS. This made Remotion frame 90 map to `2.9994s`, so timestamp-based extraction correctly returned the previous source frame.

The current PR now rounds the example's FPS value. I also created a separate follow-up docs PR to clarify that Mediabunny FPS values are packet-timestamp estimates and should be rounded when used to bootstrap a composition FPS:

https://github.com/remotion-dev/remotion/pull/8847

samohovets:
Correction/update: We split the FPS approximation fixes out of this ProRes PR.

This PR no longer changes the shared example metadata helper. Follow-ups:

- Docs clarification: https://github.com/remotion-dev/remotion/pull/8847
- Example helper change: https://github.com/remotion-dev/remotion/pull/8851

samohovets:
I opened a separate PR that changes video selection to sample at the center of the Remotion output frame while keeping audio timing at the frame start:

https://github.com/remotion-dev/remotion/pull/8854

Explainer for the approach:
https://remotion-center-sampling-vercel-6wj6hygvt.vercel.app

Once that lands in `main`, we can merge `main` back into this ProRes branch.

Issue #8277, opened 2026-06-09
Video handling in preview is completely screwed


JonnyBurger:
In NewVideo.tsx, let's switch in `https://remotion.media/remotion-studio-canvas-recording.webm` as an example video.

**First bug**

.seek is called and it tries to extract the first frame, even though the time did not change.  
We should not proceed with the seek and show the next frame when the time has not changed, we should return early

**Second bug**

We just extract the next frame and assume it actually has to be shown. 
That is not the case - the first frame should actually stay for more than 1 second. Instead, we create a new iterator every time...

PR #8760, opened 2026-07-02
`@remotion/media`: Fix preview frame accuracy

### LINKED ISSUE #8758: `@remotion/media`: Fix preview frame stepping accuracy with pooled canvases
## Summary

`@remotion/media` preview frame stepping can show a visual frame that does not match the requested timestamp.

The fix in https://github.com/remotion-dev/remotion/pull/8739 worked for certain videos, but frame stepping became more visibly offset in other cases: what was previously around a 2-frame offset became around a 3-frame offset. This was overlooked during review.

However, the issue pre-existed before that PR. The PR appears to have changed the size/timing of the visible offset rather than introduced the root cause.

## Observed behavior

The visible offset appears to correlate with Mediabunny's `CanvasSink.poolSize`:

- `poolSize: 2` -> preview can be visually offset by around 2 frames
- `poolSize: 3` -> preview can be visually offset by around 3 frames
- `poolSize: 0` -> regular frame stepping appears accurate, but this is not the desired fix

This suggests the timestamp/iterator state can be correct while the pixels shown on the canvas are from a different decoded frame.

## Expected behavior

When stepping or seeking in preview, the displayed frame should match the requested media timestamp/frame.

## Likely root cause

Remotion appears to retain or buffer `WrappedCanvas.canvas` objects returned by Mediabunny's `CanvasSink`. Since `CanvasSink` uses a canvas pool, those canvas objects may later be reused for subsequent decoded frames.

That means a `WrappedCanvas` can still have the correct timestamp metadata, while its underlying canvas pixels have already been overwritten by future decoding/prefetching.

The problem becomes especially visible around loop boundaries because loop prewarming can decode upcoming frames while the currently displayed/active frame is still retained.

## Desired fix

Fix the root cause while preserving `CanvasSink.poolSize`.

Disabling pooling is not the desired solution, because keeping a canvas pool is good for performance and avoids unnecessary allocations. Instead, Remotion should ensure that frames it buffers, retains, or exposes to preview drawing are stable snapshots and cannot be mutated by later pooled canvas reuse.

Relevant files:

- `packages/media/src/canvas-ahead-of-time.ts`
- `packages/media/src/video/video-preview-iterator.ts`
- `packages/media/src/video-iterator-manager.ts`
- `packages/media/src/prewarm-iterator-for-looping.ts`

## Testing gap

We should add regression coverage proving that:

- pooled canvas reuse does not mutate already-buffered preview frames
- frame stepping remains visually accurate with `poolSize > 0`
- loop prewarming does not mutate the currently displayed frame

samohovets:
## Summary

Fixes #8758.

This PR fixes two `@remotion/media` preview frame-accuracy issues:

1. wrong visible frames while stepping/seeking in Player preview
2. wrong frame mapping at `trimAfter + loop` boundaries

It keeps Mediabunny preview decoding pooled with `CanvasSink({poolSize: 3})`, and adds Player-based visual regression tests for the user-visible output.

## Stable preview canvases

`@remotion/media` uses Mediabunny `CanvasSink` with `poolSize: 3` for preview decoding. Mediabunny intentionally reuses pooled canvases in round-robin order. Remotion, however, keeps `WrappedCanvas.canvas` references around as the current, peeked, and buffered frames.

At the time Mediabunny yields a frame, the canvas pixels are expected to match the frame metadata. The issue is the lifetime of that canvas reference: Remotion retained pooled canvas objects beyond their safe reuse window. Once the sink reused a pooled canvas, an older retained `WrappedCanvas` could still have the old `timestamp` while its underlying `canvas` object had been repainted with newer pixels. This made preview frame stepping visually offset by the pool size.

This PR keeps Mediabunny pooling enabled, but copies each resolved `CanvasSink` frame into a Remotion-owned stable canvas before exposing it to the iterator/Player preview path. Those stable canvases are pooled on the Remotion side and released when the preview iterator replaces or destroys retained frames, so we avoid both:

- disabling Mediabunny pooling with `poolSize: 0`
- allocating an unbounded new canvas for every decoded preview frame

## `trimAfter + loop` frame mapping

The PR also fixes a separate frame-stepping issue around `trimAfter + loop` boundaries.

Previously, looped media time was wrapped using seconds-domain modulo. Near frame boundaries, this could drift to a timestamp just before the intended media frame. For example, timeline frame `1711` in the regression scenario could map just before media frame `91`.

Loop wrapping is now performed in frame units and then converted back to seconds:

```ts
((unloopedTimeInSeconds * playbackRate * fps) % loopDurationInFrames) / fps
```

This makes frame stepping deterministic at loop boundaries.

## Regression coverage

Adds/updates tests for:

- retained canvases staying visually stable even if the source canvas is reused
- normal Player frame stepping
- stepping back over a `<Loop>` boundary
- stepping across a `trimAfter + loop` boundary
- direct `trimAfter + loop` time calculation
- high-concurrency loop timestamp assertions in frame units instead of exact JS float spellings

The Player tests screenshot the actual rendered canvas, so they verify the user-visible output rather than only metadata.

## Relation to earlier media preview fixes

This change is intended to preserve the behavior from the earlier preview iterator fixes:

- #8277 / #8278: same-time seeks should return early, and long / zero-duration frames should remain visible for their timestamp range instead of blindly advancing to the next decoded frame.
- #8394: consuming a `peekedFrame` must update `lastReturnedFrame`, avoiding variable-FPS seek regressions.

The stable-canvas change only adds pixel-identity stability and lifecycle cleanup around retained frames. It does not remove the existing `currentSeek`, `peekedFrame`, or timestamp-range logic.

The likely timeline for the pooled-canvas issue is:

- #7172 introduced ahead-of-time retained canvases.
- #7289 increased preview buffering to 3 frames, making Mediabunny canvas reuse much more observable.
- #8278 / #8394 added more retained-frame bookkeeping for preview correctness, which made the metadata path more correct while the retained canvas pixels could still be overwritten by the pool.

## Test plan

```bash
cd packages/media && bunx vitest src/test/player-loop-frame-accuracy.test.tsx --browser --run
cd packages/media && bunx vitest src/test/player-preview-frame-accuracy.test.tsx --browser --run
cd packages/media && bunx vitest src/test/trim-change-seek.test.ts src/test/variable-fps-issue-8303.test.ts src/test/video-iterator-manager.test.ts src/test/canvas-ahead-of-time.test.ts src/test/loop-trim-time-calculation.test.ts --browser --run
cd packages/media && bunx vitest src/test/looping.test.ts --browser --run
cd packages/media && bun run make
cd packages/media && bun run lint
cd packages/media && bun run formatting
```

Passed locally. `bun run lint` reports existing `no-warning-comments` warnings only.


samohovets:
@pullfrog

PR #8761, opened 2026-07-02
`@remotion/zod-types`: Declare zod as a runtime dependency

yasdhf:
## Problem

A clean install of `@remotion/zod-types` can't be imported:

```bash
mkdir t && cd t && npm init -y
npm i @remotion/zod-types
node -e "require('@remotion/zod-types')"
# Error: Cannot find module 'zod'
```

The shipped `dist` imports `zod`, but `zod` is declared only in `devDependencies`, so consumers never receive it.

## Fix

Move `zod` (`catalog:`) from `devDependencies` to `dependencies`. This matches `@remotion/studio`, which already declares `zod` as a runtime dependency. No code change.

## Verification

`npm i @remotion/zod-types zod` + import succeeds (4 exports resolve). The only thing missing on a bare install was `zod` itself.

### DIFF
--- a/packages/zod-types/package.json
+++ b/packages/zod-types/package.json
@@ -21,10 +21,10 @@
 		"url": "https://github.com/remotion-dev/remotion/issues"
 	},
 	"dependencies": {
-		"remotion": "workspace:*"
+		"remotion": "workspace:*",
+		"zod": "catalog:"
 	},
 	"devDependencies": {
-		"zod": "catalog:",
 		"@remotion/eslint-config-internal": "workspace:*",
 		"eslint": "catalog:",
 		"@typescript/native-preview": "catalog:"


JonnyBurger:
Thanks for the reproduction. We recently removed the Zod peer dependency intentionally because it made upgrades harder. We also do not want to make `zod` a regular dependency of `@remotion/zod-types`, because that would tie users too tightly to the Zod version chosen by this package. The intended model is that users install `zod` directly, and the schema docs already show installing both: `npx remotion add @remotion/zod-types zod` (https://www.remotion.dev/docs/schemas). We can clarify the dedicated `@remotion/zod-types` install page separately, but this package.json change is not the direction we want, so I am closing this PR.

PR #8762, opened 2026-07-02
`@remotion/media`: Fix trimAfter loop frame stepping

samohovets:
## Summary

Follow-up to the preview frame accuracy branch:

- map looped trimmed media time in frame units to avoid floating-point drift at `trimAfter` boundaries
- add a focused unit test for the `trimBefore={90}` / `trimAfter={1800}` / `loop` boundary mapping
- make the trimAfter Player visual regression test wait for an actual `onVideoFrame` draw after seeks
- co-locate the loop-boundary and trimAfter-loop Player tests so their screenshots live in one folder
- update the existing high-concurrency looping test to assert mapped frame numbers instead of exact JavaScript float spellings
- slightly relax the normal preview visual screenshot tolerance to avoid a 1-pixel browser screenshot mismatch

## Note about the high-concurrency looping test

The frame-domain timing fix intentionally changes insignificant floating-point representations, for example `6.466666666666666` may become `6.466666666666667`. Both represent the same intended media frame (`194 / 30`). The test now verifies the frame-accurate result by converting seconds back to rounded frame numbers instead of comparing exact float literals.

## Validation

Autoresearch final metric:

```text
target_failures: 1 → 0
preview_failures: 0
loop_boundary_failures: 0
typecheck_failures: 0
```

Additional local checks:

```bash
cd packages/media && bunx vitest src/test/looping.test.ts --browser --run
cd packages/media && bun run lint && bun run formatting
```

The finalized branch was verified against the autoresearch final tree with no non-session differences.


samohovets:
@pullfrog

PR #8800, opened 2026-07-06
`@remotion/core`: Support fontWeight 650 in Interactivity

### LINKED ISSUE #8793: `fontWeight: 650` is not supported by Interactivity
This renders a dropdown but it is not a supported value
Note that the value is not quoted

https://www.remotion.dev/elements/overlays/lower-third/

dpersek:
## Summary
- Add `650` as a supported `style.fontWeight` Interactivity option.
- Extend the existing text schema exposure test to pin the new `650` enum value.

## Why
Fixes #8793.

The lower-third element can use a numeric `fontWeight: 650`, but Interactivity only exposed the existing `100` through `900` hundred-step weights plus `normal` and `bold`. Adding the exact missing weight lets Studio select the current lower-third value instead of leaving it unsupported.

## Tests
- `bun test src/test/with-interactivity-schema-helpers.test.ts`
- `bun run formatting`
- `git diff --check`
- `bunx turbo run make --filter=remotion`
- `bun test src/test`
- `bun run stylecheck`
- `bun run build`

## Scope
- Only adds the missing `650` option to the Interactivity schema and its existing schema test.
- No Studio UI rewrite.
- No arbitrary/free-form variable font-weight editor.
- No docs or lower-third element rewrite.

## Caveats
- I did not add a browser/Studio smoke test. The behavior is schema-level, and the existing enum field path stringifies numeric current values before matching enum ids.


JonnyBurger:
This was just an example, we need more generic handling, but thank you

dpersek:
Ha! I had considered that, but the way it was worded, thought maybe he just wants 650 supported, whoops! =) Sorry for spamming with an incorrect PR!

PR #8854, opened 2026-07-07
`@remotion/media`: Sample video frames at center

samohovets:
## Summary
- Select video frames using a midpoint sample instead of the frame start for normal-speed playback
- Keep speed-up playback anchored to the first represented input frame so existing `playbackRate > 1` semantics remain frame-accurate
- Keep audio timing on the frame start while applying video-only midpoint selection to rendering and preview seeking
- Make keyframe-bank near-boundary selection deterministic by avoiding broad future-frame tolerance
- Add focused regression tests for frame selection, speed-up playback, and keyframe boundary behavior

## Why
A Remotion frame represents a time interval, not just an instant. When composition FPS and source timestamps do not align exactly, sampling video at the frame start can choose the previous source frame even though the output frame mostly overlaps the next source frame.

For normal playback, midpoint sampling approximates choosing the source frame with the greatest overlap and fixes delayed/choppy frame selection without requiring input and output FPS to match.

For `playbackRate > 1`, one output frame represents multiple input frames. To preserve existing Remotion semantics and frame-accuracy tests, the selection stays anchored to the first represented input frame rather than jumping to the middle of the skipped-frame span.

Explainer: https://remotion-center-sampling-vercel-doiypknbb.vercel.app

## Testing
- `cd packages/media && bun run format`
- `cd packages/media && bunx vitest src/test/looping.test.ts src/test/video-frame-selection-time.test.ts --browser --run`
- `bunx turbo run make --filter='@remotion/media'`


samohovets:
Fixed the CI regression for `playbackRate={4}`. The midpoint offset is now capped for speed-up playback so output frame N remains anchored to the first represented input frame (matching the existing frame-accuracy tests), while normal-speed playback still uses midpoint selection for the ProRes/FPS-boundary case. `@remotion/media` tests pass locally.

PR #8879, opened 2026-07-07
`@remotion/media`: Fix preview frame selection and iterator reuse

samohovets:
## Summary

- Fix the `@remotion/media` preview video iterator so normal forward frame stepping can wait for a pending decoded sample instead of treating the iterator as unsatisfied and recreating it.
- Improve preview frame selection for slightly over-precise or wrong media FPS metadata: a decoded next frame may be used only when it is just barely in the future, closer than the previous frame, and still within the media cadence.
- Consume a selected peeked next frame so the preview does not draw the same frame again on the next seek.

This PR is intentionally scoped to the preview path. Render-path dropped-frame/frame-selection behavior will be handled separately. These changes were discovered while validating the ProRes PR #8740, but are independent and should land/review separately.

## Testing

- `cd packages/media && bun run lint` ✅ 0 errors (existing unrelated TODO warnings remain)
- `cd packages/media && bun run test` ✅ 28 files / 61 tests passed
- `cd packages/media && bun run make` ✅
- `cd packages/example && bun run bundle` ✅
- `cd packages/it-tests && bun test src/rendering/new-video-tests.test.ts --run --timeout 60000` ✅ 4 tests passed

### DIFF
--- a/packages/media/src/test/keyframe-bank.test.ts
+++ b/packages/media/src/test/keyframe-bank.test.ts
@@ -34,6 +34,65 @@ const makeVideoSampleSink = (samples: VideoSample[]): VideoSampleSink => {
 	};
 };
 
+test('does not select a near future frame', async () => {
+	const samples = [
+		makeSample({timestamp: 0, duration: 0.01}),
+		makeSample({timestamp: 0.033, duration: 0.033}),
+	];
+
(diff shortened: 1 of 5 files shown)


samohovets:
Closing this in favor of #8917, which is fixing the root cause with durations instead of adding more safety checks and tolerance

PR #8888, opened 2026-07-07
`@remotion/studio`: Add Remotion Convert asset handoff

### LINKED ISSUE #8884: How could a video be edited with remotion.dev/convert in the Remotion Studio?
Let's say I have an asset in the public/ folder in the Remotion Studio.
I select it and the asset inspector is opening.

Now, I want to trim, crop, convert, whatever this asset.
I would like to open remotion.dev/convert (packages/convert) and make the asset available to it.
I do the operation, and in the end, I can move the asset back to the Remotion Studio.

i wonder how and if that is possible technically, given we are on a different domain.

JonnyBurger:
## Summary

- Add a Studio asset action that opens audio/video public assets in a local Remotion Convert tab.
- Add a token-based postMessage bridge so Convert can save finished files back into Studio.
- Allow Studio static asset reads from local Convert origins while keeping writes same-origin through Studio.

Closes #8884

## Testing

- bun run build
- bun run stylecheck

### DIFF
--- a/packages/convert/app/components/ConversionDone.tsx
+++ b/packages/convert/app/components/ConversionDone.tsx
@@ -1,6 +1,8 @@
 import {Button, Button as RemotionButton} from '@remotion/design';
-import React, {useCallback} from 'react';
+import React, {useCallback, useState} from 'react';
 import type {ConvertState, Source} from '~/lib/convert-state';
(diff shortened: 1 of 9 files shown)


JonnyBurger:
Extremely cool, but not top of my priority list right now

PR #8891, opened 2026-07-08
`remotion`: Replace deprecated `MutableRefObject` with `RefObject`

iliya-vidrush:
## Summary

React 19's `@types/react` deprecates `MutableRefObject<T>` and makes `RefObject<T>` mutable (`{ current: T }`), so the two types are now equivalent.

This replaces all remaining `MutableRefObject` usages with `RefObject` across the affected packages:

- `remotion` (core)
- `@remotion/player`
- `@remotion/studio`
- docs, promo-pages, example

The change is **type-only** and has no runtime effect. It type-checks against the installed `@types/react@19`, where writes to `.current` remain valid.


JonnyBurger:
Thanks!

PR #8892, opened 2026-07-08
`remotion`: Fix player buffering subscription leak and report state synchronously

iliya-vidrush:
## Summary

Reworks the player buffer manager to track blocks and listeners in refs and report buffering state synchronously, and fixes a subscription leak in `useIsPlayerBuffering`.

## Changes

- **Track buffering state in refs instead of `useState`.** `blocks`, `onBufferingCallbacks`, and `onResumeCallbacks` are now refs. Adding/removing a block calls `checkBuffering()` directly, so buffer/resume listeners fire synchronously on the `0 <-> >0` blocks transition without waiting for a render + effect.
- **Fire callbacks only on the actual transition.** `checkBuffering` compares against the last reported state (`buffering.current`), so intermediate changes (e.g. `2 -> 3` blocks) no longer re-notify listeners. Callbacks are iterated over a copy so a callback can remove itself while being called. This preserves the behavior the old `useEffect`/`useLayoutEffect` pair was working around, without the extra renders.
- **Fix subscription leak in `useIsPlayerBuffering`.** The cleanup previously called `listenForBuffering(() => undefined)` / `listenForResume(() => undefined)`, which *registered a new no-op listener* instead of removing the original ones — leaking a subscription on every re-subscribe. It now keeps the returned handles and calls `.remove()`.
- **Sync state after subscribing.** `setIsBuffering(bufferManager.buffering.current)` runs right after subscribing, so a buffer state that flipped between the initial render and effect setup (e.g. a media element that blocked before we listened) is not missed.

### DIFF
--- a/packages/core/src/buffering.tsx
+++ b/packages/core/src/buffering.tsx
@@ -2,7 +2,6 @@ import React, {
 	useCallback,
 	useContext,
 	useEffect,
-	useLayoutEffect,
 	useMemo,
 	useRef,
 	useState,
@@ -42,18 +41,49 @@ const useBufferManager = (
 	logLevel: LogLevel,
 	mountTime: number | null,
 ): BufferManager => {
-	const [blocks, setBlocks] = useState<Block[]>([]);
-	const [onBufferingCallbacks, setOnBufferingCallbacks] = useState<
-		OnBufferingCallback[]
-	>([]);
-	const [onResumeCallbacks, setOnResumeCallbacks] = useState<
-		OnBufferingCallback[]
-	>([]);
+	const blocks = useRef<Block[]>([]);
+	const onBufferingCallbacks = useRef<OnBufferingCallback[]>([]);
+	const onResumeCallbacks = useRef<OnResumeCallback[]>([]);
 
 	const env = useRemotionEnvironment();
 	const rendering = env.isRendering;
 
 	const buffering = useRef(false);
+
+	// Fire callbacks only on the `0 <-> >0` blocks transition.
+	// `buffering` holds the last reported state, so intermediate changes
+	// (e.g. 2 -> 3 blocks) are no-ops and never re-notify listeners.
+	const checkBuffering = useCallback(() => {
+		if (rendering) {
+			return;
+		}
+
+		const isBuffering = blocks.current.length > 0;
+		if (isBuffering === buffering.current) {
+			return;
+		}
+
+		buffering.current = isBuffering;
+
+		if (isBuffering) {
+			// Iterate a copy: a callback may remove itself while being called.
+			[...onBufferingCallbacks.current].forEach((c) => c());
+			playbackLogging({
+				logLevel,
+				message: 'Player is entering buffer state',
+				mountTime,
+				tag: 'player',
+			});
+		} else {
+			[...onResumeCallbacks.current].forEach((c) => c());
+			playbackLogging({
+				logLevel,
+				message: 'Player is exiting buffer state',
+				mountTime,
+				tag: 'player',
+			});
+		}
+	}, [logLevel, mountTime, rendering]);
 
 	const addBlock: AddBlock = useCallback(
 		(block: Block) => {
@@ -65,7 +95,8 @@ const useBufferManager = (
 
 			let unblocked = false;
 
-			setBlocks((b) => [...b, block]);
+			blocks.current.push(block);
+			checkBuffering();
 			return {
 				unblock: () => {
 					if (unblocked) {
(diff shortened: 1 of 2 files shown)


iliya-vidrush:
> **Reviewed changes** — the PR reworks the player buffering manager to use refs instead of state, reports buffering state synchronously on block/unblock transitions, and fixes a subscription leak in `useIsPlayerBuffering`.
> 
> * **`useBufferManager` refactors buffering bookkeeping into refs** — `blocks`, `onBufferingCallbacks`, and `onResumeCallbacks` are ref-backed, with a new `checkBuffering` helper that fires callbacks only on the `0 ↔ >0` transition.
> * **`BufferManager` now exposes `buffering` as a `MutableRefObject<boolean>`** so consumers can read the live sync state without subscribing.
> * **`useIsPlayerBuffering` fixes the cleanup leak** — it now retains listener handles and calls `.remove()` instead of registering no-op listeners.
> * **State is synced after subscription** so a buffering state change between initial render and effect setup is not missed.
> 
> Caution
> 
> The cleanup leak is a real regression fix, but the synchronous callback logic is subtle and has correctness gaps that should be addressed before merging.
> 
> ### 🚨 `checkBuffering` may miss the first transition during render / event-handler ordering
> `buffering` is initialized to `false` unconditionally, even if blocks already exist at manager creation. Before this PR, the `useEffect`/layout-effect pair inspected `blocks.length` on each render and would fire enter/exit callbacks as soon as React committed. Now, if a block is added between manager creation and the first listener subscription, the live state is `blocks.current.length > 0` but `buffering.current` is still `false`. When `listenForBuffering` is called it does **not** run `checkBuffering`, so the first listener is never notified that the player is already buffering.
> 
> `useIsPlayerBuffering` mitigates this for its own hook by syncing `setIsBuffering(bufferManager.buffering.current)` after subscribe, but that only reads the stale `buffering.current` (`false`) in the scenario above. A consumer that uses `listenForBuffering` directly (e.g. the media playback loop, tests, or future callers) will silently miss an already-buffering state.
> 
> The fix is to initialize `buffering` from `blocks.current.length` inside `checkBuffering` and/or to re-evaluate state whenever a new listener is registered. Without that, the synchronous-transition contract is broken for any listener that subscribes after the first block.
> 
> ### ⚠️ A listener registered _while_ a transition is being fired can observe a stale state
> `checkBuffering` snapshots `[...onBufferingCallbacks.current]` (or resume) before iterating. That's good for self-removal. But if a callback adds a new listener during iteration, the new listener is not in the snapshot and won't be called for the current transition. More importantly, `buffering.current` has already been updated by the time the new listener exists, so the newly-added listener has missed the transition entirely and has no future event to catch up on.
> 
> This is the same class of problem the old `useEffect` arrangement avoided for initial subscriptions (listeners were always evaluated on next render). With the new synchronously-fired callbacks, registration after `buffering.current` is updated but before the matching `onResume`/onBuffering has any chance to fire leaves the listener out of sync. The current `useIsPlayerBuffering` post-subscribe sync addresses the narrow hook case, but the manager itself is still inconsistent.
> 
> Consider recomputing/emitting state from inside `listenForBuffering`/`listenForResume` when the current state differs from what a newly-registered listener would expect, or documenting that listeners must be present before the transition occurs.
> 
> ### ℹ️ Nitpicks
> * The PR description mentions a version bump for `@remotion/core`, but the diff doesn't touch `version.ts` or package metadata. If that was intended, it should be included; otherwise the description should be updated.
> * `buffering.current` is read inside `useMediaPlayback.ts` at line 172. Confirm that calling sites are happy with a ref read that may lag behind the actual blocks list during the first-render window described above.
> 
> [Pullfrog](https://pullfrog.com)  ｜ [Fix it ➔](https://pullfrog.com/trigger/remotion-dev/remotion/8892?action=fix&review_id=4651174972) ｜ [View workflow run](https://github.com/remotion-dev/remotion/actions/runs/28919669444/job/85793921861) ｜ Using `Kimi K2` (free via [Pullfrog for OSS](https://pullfrog.com/for-oss)) ｜ [𝕏](https://x.com/pullfrogai)

`buffering.current` is now maintained internally by checkBuffering on every block mutation, rather than only by external listeners

iliya-vidrush:
@pullfrog review again

JonnyBurger:
That's too many changes to a core part of Remotion and too many issues flagged for me to merge.
What issue do you face? Do you wanna open an issue first?

iliya-vidrush:
@JonnyBurger Fair point — opened #8894 describing the issue we face.

TL;DR: our app drives a buffering overlay from the Player buffering API. On every navigation to a video, mounting media elements causes spurious `waiting`/`resume` sequences (plus a listener leak in `useIsPlayerBuffering` — its effect cleanup registers no-op listeners instead of removing the real ones), so the overlay flickers.

Regarding the two automated review flags — we believe they are false positives:

1. **"`checkBuffering` may miss the first transition"** rests on the premise that `buffering.current` can be stale while blocks exist. It can't: the only way `blocks` changes is through `addBlock`/`unblock`, and both call `checkBuffering` synchronously, which updates `buffering.current` before returning. Both in-repo subscribers are covered — `useIsPlayerBuffering` seeds and re-syncs from `buffering.current`, and `useBufferStateEmitter` subscribes in a layout effect before any block can be added.
2. **"Listener registered during dispatch sees stale state"** requires a buffering callback to synchronously register a new listener while being invoked — nothing in the codebase does this. The snapshot copy handles the real case (a callback removing itself).

That said, happy to shrink this to a minimal PR that only fixes the listener leak (retain the handles and call `.remove()` in cleanup) if you'd prefer to keep the ref refactor out of core — just say the word.

JonnyBurger:
Superseded by my own PR! Released in latest version

PR #8912, opened 2026-07-08
`@remotion/media`: Share Input across MediaPlayers of the same source

pravchuk:
Previously every MediaPlayer created its own mediabunny Input and UrlSource. When multiple ranges of the same media are mounted (e.g. jump cuts or play ranges, where a new Sequence mounts a fresh <Video> per range), each mount re-fetched and re-parsed the container header and cold-seeked from scratch, producing a visible loader at every cut.

Inputs are now cached and reference counted per (src, credentials, requestInit) in get-shared-input.ts, mirroring the existing sink cache in get-sink.ts. All players of the same source share one Input, so the container is parsed once and subsequent seeks are served from the warm UrlSource byte cache. The shared Input is only disposed once the last player releases it.

Because the Input is now shared, `input.disposed` no longer indicates whether a given player was disposed. MediaPlayer tracks its own lifecycle with a per-player `disposed` flag, and dispose() is guarded against double-release so the shared Input's reference count cannot be over-decremented.

Before:

https://github.com/user-attachments/assets/75e5e471-6ab7-4206-a739-26212e7928a5

After:

https://github.com/user-attachments/assets/60f73295-c1f8-43ba-a458-f4c19b8e0a85

### DIFF
--- a/packages/media/src/get-shared-input.ts
+++ b/packages/media/src/get-shared-input.ts
@@ -1,0 +1,96 @@
+import {ALL_FORMATS, Input, UrlSource} from 'mediabunny';
+import {
+	getMediaRequestInitFingerprint,
+	normalizeMediaRequestInit,
+	resolveRequestInit,
+	type MediaRequestInit,
+} from './request-init';
+
+// A single mediabunny `Input` (backed by one `UrlSource`) is expensive to spin
+// up for network media: on creation it must fetch and parse the container
+// header, and its `UrlSource` keeps a byte cache + demuxer state that is warm
+// only for that instance.
+//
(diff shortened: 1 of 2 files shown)


JonnyBurger:
Thanks! This is very valuable!

Let me actually talk to Vanilagy from Mediabunny about it this weekend because I will see him.
I'm not entirely sure on which level the caching is implemented on Mediabunny side - Input, track, Sink or iterator.

This will help me figure out the right place to do the caching. PR looks quite good though!

JonnyBurger:
@Vanilagy says caching the Input is good!

PR #8917, opened 2026-07-08
`@remotion/media`: Ensure we always get the right duration from mediabunny frame by peeking

### LINKED ISSUE #8916: Mediabunny CanvasSink returns inconsistent durations for ProRes MKV frames at Matroska cluster boundaries
## Summary

While investigating `CanvasSink` timing for MKV inputs through Mediabunny, I found that frame durations can violate the expected invariant:

```ts
duration === nextTimestamp - timestamp
```

This is easy to see with the ProRes MKV below, and it appears to be a general Matroska cluster-boundary issue rather than something specific to ProRes.

## Reproduction inputs

- H.265 MKV: https://remotion.media/video-h265.mkv
- ProRes MKV: https://remotion.media/video-prores.mkv

Both files use Matroska millisecond timestamps and have a nominal 30fps frame duration.

## Observed behavior

Using Mediabunny's `CanvasSink`:

```txt
video-h265.mkv:
0.000 @ 0.033
0.033 @ 0.034
0.067 @ 0.033
0.100 @ 0.033
0.133 @ 0.034
...

video-prores.mkv:
0.000 @ 0.033
0.033 @ 0.033
0.067 @ 0.033
0.100 @ 0.033
0.133 @ 0.033
...
```

For the ProRes file, the frame at `0.033` has duration `0.033`, but the next timestamp is `0.067`, so the true interval is `0.034`.

## Expected behavior

For frame/sample/canvas iteration, I would expect the duration of each frame to match the delta to the next presentation timestamp whenever that next timestamp is known.

## Root cause found in Mediabunny

Mediabunny's Matroska demuxer computes block durations from adjacent presentation timestamps only within the current cluster. If the current block is the last block in its cluster, it falls back to the track's `DefaultDuration`.

That matters here because:

- Both files have `TimestampScale = 1_000_000`, so timestamps are exposed on a millisecond grid: `0`, `0.033`, `0.067`, `0.100`, ...
- Both files have `DefaultDuration = 33_333_333ns`, i.e. `0.033333333s`.
- The H.265 file has many video blocks per cluster, so most durations are inferred from the next in-cluster presentation timestamp and become `0.033`/`0.034` as expected.
- The ProRes file has one video block per cluster, so every frame is the last frame in its cluster and the demuxer keeps using `DefaultDuration` instead of looking into the next cluster.

So yes, this can also happen with Matroska generally whenever a cluster is finished and the next frame for the same track is in the following cluster. ProRes just exposes it on every frame because each ProRes frame is in its own cluster in this sample.

JonnyBurger:
The issue was that we could not trust `.duration` of Mediabunny, we have to make sure by peeking next

Closes https://github.com/remotion-dev/remotion/issues/8916

### DIFF
--- a/packages/media/src/test/keyframe-bank.test.ts
+++ b/packages/media/src/test/keyframe-bank.test.ts
@@ -53,3 +53,23 @@ test('a long frame duration can satisfy requests beyond the jump threshold', asy
 	bank.prepareForDeletion('error', 'test');
 	samples[1].close();
 });
+
+test('uses next sample timestamp instead of reported duration while rendering', async () => {
+	const samples = [
+		makeSample({timestamp: 0, duration: 10}),
+		makeSample({timestamp: 1, duration: 0.1}),
+	];
+
+	const bank = await makeKeyframeBank({
+		logLevel: 'error',
+		src: 'wrong-duration-frame.mp4',
+		videoSampleSink: makeVideoSampleSink(samples),
+		initialTimestampRequest: 0,
+	});
+
+	const frame = await bank.getFrameFromTimestamp(1.5, 30);
+
+	expect(frame?.timestamp).toBe(1);
+
+	bank.prepareForDeletion('error', 'test');
+});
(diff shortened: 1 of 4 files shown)


samohovets:
Thank you! Tested, LGTM

samohovets (approved):

PR #8919, opened 2026-07-08
`@remotion/studio`: Allow Elements to inherit composition dimensions

### LINKED ISSUE #8799: Studio: Let Elements inherit composition dimensions
## Summary

Elements dragged from `/elements` currently need explicit dimensions in their gallery metadata. This means an Element cannot opt into inheriting the target composition dimensions when it is dropped into Remotion Studio.

## Current behavior

- `ElementPage` only creates drag data when `elementWidth` and `elementHeight` are provided.
- `ElementDragData` requires positive finite `dimensions`, and parsing rejects omitted dimensions.
- Studio sends those dimensions to `/api/insert-element`.
- The insertion route validates dimensions and wraps the dropped Element in a `<Sequence width={...} height={...}>`.
- Because `useVideoConfig()` is overridden by `SequenceContext`, an Element that reads `width` from `useVideoConfig()` inherits the generated wrapper dimensions rather than the target composition dimensions.

## Expected behavior

Elements should be able to declare that they inherit dimensions from the target composition. In that mode, the docs page should still be able to preview the Element, the drag button should remain available, and Studio should insert the Element without forcing fixed wrapper dimensions.

## Tasks

- [ ] Allow Element drag data to represent inherited or omitted dimensions.
- [ ] Keep the docs Element drag button enabled for inherited-dimension Elements.
- [ ] Update Studio insertion so inherited-dimension Elements do not get a fixed `Sequence` width/height wrapper.
- [ ] Add tests for the shared drag-data parser and insertion behavior.
- [ ] Update Element docs/examples once the inherited-dimension mode exists.

## Context

Discovered while adding the Paper Texture Element in #8791. That Element wants a `<Solid>` width based on `useVideoConfig().width`, but the current drag system still supplies explicit Element dimensions.

sksingh2005:
Fixes #8799

## Summary

Allow Elements to inherit composition dimensions when dragged into Remotion Studio.

- Elements dropped from `/elements` without explicit dimensions are now inserted directly without a `<Sequence>` wrapper.
- Omitted dimensions are preserved through the drag data payload so the frontend and server know not to constrain the element.
- Retains exact drop position `{x, y}` coordinates by applying the position directly to the Element component instead of a wrapper.
- Allows responsive Elements (like a Paper Texture) to dynamically scale using `useVideoConfig().width`.

## Testing

```bash
bunx turbo run test --filter="@remotion/studio-shared"
bun run build
bun run stylecheck

### DIFF
--- a/packages/docs/elements/backgrounds/paper-texture/index.mdx
+++ b/packages/docs/elements/backgrounds/paper-texture/index.mdx
@@ -10,4 +10,4 @@ import {PaperTexture} from './paper-texture';
 
 A white paper texture background with a slowly changing posterized seed.
 
-<ElementPage component={PaperTexture} displayName="Paper Texture" elementHeight={1080} elementWidth={1920} slug="backgrounds/paper-texture" sourceFile="./paper-texture.tsx" />
+<ElementPage component={PaperTexture} displayName="Paper Texture" slug="backgrounds/paper-texture" sourceFile="./paper-texture.tsx" />
--- a/packages/docs/elements/backgrounds/paper-texture/paper-texture.tsx
+++ b/packages/docs/elements/backgrounds/paper-texture/paper-texture.tsx
@@ -8,6 +8,8 @@ export const PaperTexture: React.FC = () => {
 	return (
 		<Solid
 			color="white"
+			width={1920}
+			height={1080}
 			effects={[
 				paper({
 					colorFront: 'white',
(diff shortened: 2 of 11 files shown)


sksingh2005:
@JonnyBurger Hi, I am a new contributor and have made this pr for fixing #8799 . Though I am not able to see the pullfrog comments, can you retrigger that please if it is a bug. Feel free to review and leave your comments too.
Thanks !!

JonnyBurger:
Thank you!

PR #8951, opened 2026-07-09
`@remotion/studio`: Add docs Element install button

samohovets:
## Summary
- add an Install Element button to docs Element pages that targets the most recently focused Studio
- expose Studio discovery/request endpoints and deliver install requests via targeted SSE
- show a Studio confirmation dialog with a collapsed source-code preview before writing files
- track install targets per Studio client to avoid background tab targeting issues

Related to #7698 

## Validation
- `cd packages/studio-server && bun test src/test/element-install-state.test.ts`
- `bun run build`
- `bun run stylecheck`

### DIFF
--- a/packages/docs/src/components/Elements/ElementPage.tsx
+++ b/packages/docs/src/components/Elements/ElementPage.tsx
@@ -2,7 +2,13 @@ import {
 	makeElementDragData,
 	type ComponentDimensions,
 } from '@remotion/studio-shared';
-import React, {useMemo, type ComponentType, type ReactNode} from 'react';
+import React, {
+	useCallback,
+	useMemo,
+	useState,
+	type ComponentType,
+	type ReactNode,
+} from 'react';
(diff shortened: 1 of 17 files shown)


samohovets:
`/api/update-element-install-target` is not exposed through the public CORS allowlist. It is registered in `allApiRoutes` and handled through `handleRequest()`, which calls `validateSameOrigin(request)`. The relaxed element-install CORS policy in `routes.ts` only applies to the docs-facing discovery/request endpoints.

Therefore a random localhost page cannot publish `canInstall`, `compositionFile`, or `compositionId` from the browser. The remaining trust boundary is same-origin Studio JS, which already has access to stronger Studio write APIs. File mutation is still gated by the Studio confirmation dialog.

`sendEventToClientId()` only guarantees that the event was written to the currently connected SSE socket, not application-level processing by React. That is intentional for this flow: the API returns `status: "sent"`, and the docs UI tells the user to confirm the install in Studio rather than implying completion.

PR #8972, opened 2026-07-10
Forward AnimatedImage data attributes to canvas

alec-watts:
## Summary
- Forward `data-*` and `aria-*` props from `<AnimatedImage>` to its rendered canvas despite the internal `layout=\"none\"` Sequence wrapper.
- Add a regression test covering `aria-label` and `data-testid` on the rendered canvas.

## Test plan
- `cd packages/core && bun test src/test/sequence-register-once.test.tsx`
- `cd packages/core && bun run lint`


JonnyBurger:
Thanks!

PR #8999, opened 2026-07-10
Skills: Split up into multiple skills, make "Remotion Best Practices" the router

JonnyBurger:
## Summary

- split the Remotion skills into more focused skill packages
- fix internal skill links after the refactor
- make the blank template props type pass lint after the new structure

## Checks

- bun run build
- bun run stylecheck
- checked 67 relative markdown links across 46 skill markdown files


packages/template-blank/src/Composition.tsx
```diff
@@ -1,3 +1,25 @@
+import { CalculateMetadataFunction, Composition } from "remotion";
+
+type Props = Record<string, unknown>;
```
JonnyBurger: This should be {} and ESLint rule disabled

JonnyBurger (commented):
Must "bundle" the skill in case you only have remotion-best-practices

Issue #418, opened 2021-06-02
Disable rule "Missing return type on function (@typescript-eslint/explicit-module-boundary-types)"?


JonnyBurger:
Maybe overkill for people to get started, the inference is most of the time pretty good

PR #9000, opened 2026-07-10
`docs`: Fix CrawlChat embed causing hydration errors and duplicate script execution

### LINKED ISSUE #8790: CrawlChat embed causes client-side React errors (Minified React error #418) and duplicate identifier — hurts SEO
# Bug Report: CrawlChat embed causes client-side React errors (Minified React error #418) and duplicate identifier — hurts SEO

On the docs site CrawlChat embeds are producing client-side errors which likely harm SEO by causing recoverable/unrecoverable React errors during client render.

Error output observed (copied from console):

```
main.d8fa2500.js:27 Docusaurus React Root onRecoverableError: Error: Minified React error #418; visit <https://react.dev/errors/418?args[]=text&args[]=> for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
    at rI (main.d8fa2500.js:1:305887)
    at rD (main.d8fa2500.js:1:306920)
    at main.d8fa2500.js:1:390416
    at ck (main.d8fa2500.js:1:395288)
    at cS (main.d8fa2500.js:1:387077)
    at main.d8fa2500.js:1:382137
    at cl (main.d8fa2500.js:1:382238)
    at cZ (main.d8fa2500.js:1:405377)
    at MessagePort.A (main.d8fa2500.js:1:471414) Object
n @ main.d8fa2500.js:27
embed.js:1 Uncaught SyntaxError: Identifier 'CrawlChatEmbed' has already been declared
entry.client-BrT_nwts.js:26 Uncaught Error: Minified React error #418; visit <https://react.dev/errors/418?args[]=HTML&args[]=> for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
    at Hi (entry.client-BrT_nwts.js:26:31198)
    at Ic (entry.client-BrT_nwts.js:26:81843)
    at Ru (entry.client-BrT_nwts.js:26:115784)
    at Lu (entry.client-BrT_nwts.js:26:115757)
    at Iu (entry.client-BrT_nwts.js:26:115617)
    at Su (entry.client-BrT_nwts.js:26:111211)
    at md (entry.client-BrT_nwts.js:26:123099)
    at re (entry.client-BrT_nwts.js:19:40163)
    at chat-box-D13KVkAO.js:2:89413
    at f (chat-box-D13KVkAO.js:2:89524)
    at p (chat-box-D13KVkAO.js:2:89625)
```

Bundles referenced in console: main.d8fa2500.js, embed.js, entry.client-BrT_nwts.js, chat-box-*.js

Steps to reproduce
1. Visit docs pages containing the CrawlChat embed.
2. Open the browser console.
3. Observe the React minified errors and the "Identifier 'CrawlChatEmbed' has already been declared" syntax error.

Expected behavior
- No client-side JS errors from CrawlChat embed.
- Embed should not cause React hydration mismatches or duplicate global declarations.
- Server-rendered content should remain crawlable for SEO.

Notes / possible causes
- The embed script may be injected/executed multiple times causing duplicate global declarations (CrawlChatEmbed).
- React hydration mismatch or invalid render target can trigger Minified React error #418 (text/HTML render mismatch).
- The embed may not be SSR/hydration-safe and might run during server render or double-initialize.

Suggested fixes / next steps
- Guard the embed initialization to run only once (idempotent init) and avoid declaring globals in page scope.
- Ensure the embed defers client-only DOM mutations until after hydration, or renders safely with SSR placeholders.
- Reproduce in a non-minified dev build to get full React error traces and stack frames.
- Add an automated check to ensure docs pages do not produce console errors.

Priority: High — affects SEO and public site stability.

Environment: observed in production bundles on the docs site (minified builds).

[View original Slack conversation](https://remotionglobal.slack.com/archives/C0BE9EMGS5V/p1783334608492969?thread_ts=1783334608.492969&cid=C0BE9EMGS5V)

pelazas:
Fixes #8790

## What's happening

The CrawlChat embed is rendered as a raw `<script>` tag inside JSX in `packages/docs/src/theme/Layout/index.tsx`. That tag ends up in the server-rendered HTML of every page, which causes two of the three errors from the issue:

1. The browser executes `embed.js` before React hydrates. The script mounts the widget and restyles the TOC column right away, so the DOM no longer matches the server HTML when hydration runs → minified React error #418 from the Docusaurus root.
2. When hydration fails, React falls back to client-rendering the tree, which inserts a *fresh* `<script>` element — and freshly inserted scripts execute. `embed.js` declares a top-level `class CrawlChatEmbed`, so the second run throws `Identifier 'CrawlChatEmbed' has already been declared`. Remounting the layout wrapper (e.g. navigating to/from `/experimental_new`) can trigger the same thing.

You can see the "before" state on production right now — `curl https://www.remotion.dev/docs/the-fundamentals` has the embed script tag in the SSR HTML.

## The fix

Inject the script from a `useEffect`, guarded by a `getElementById` check so it only ever runs once. It never appears in the SSR HTML (nothing executes before hydration) and never lives in the React tree (nothing can re-create it). No cleanup on unmount on purpose: removing the tag wouldn't undo the global class declaration, and re-inserting it later would throw again.

The element keeps the same id and data attributes, so `getScrapeId()` in `embed.js` and the `#crawlchat-script` CSS on `/experimental_new` keep working.

FWIW, `crawlchat-client`'s own `<CrawlChatScript>` component renders the same JSX `<script>` tag, so switching to it wouldn't have helped — probably worth reporting upstream.

## What this doesn't fix

The third error in the issue (`entry.client-BrT_nwts.js` / `chat-box-*.js`) is a hydration error *inside* CrawlChat's iframe, thrown by their own app bundle on crawlchat.app. I confirmed it still fires with a stack pointing entirely at their assets — nothing in this repo can fix that one, it needs a report to CrawlChat.

## Tradeoffs

The embed now loads strictly after hydration instead of in parallel with it, so the widget button appears marginally later. Since `embed.js` is `async` and waits for `window.onload` anyway, the difference is negligible.

## How I tested

- `bun run build-docs`, then grepped the output: zero HTML files contain the embed script tag (all 1016 pages had it before)
- Served the build and drove it with headless Chrome: no React #418 from the docs bundle, no duplicate-identifier error, `window.crawlchatEmbed` mounts, widget container in the DOM, and exactly one script tag after initial load, after client-side navigation, and after an `/experimental_new` roundtrip
- `bun test src`, `eslint`, `oxfmt --check` all pass

If this regressed, the symptom would be the widget not appearing on docs pages — no impact outside the docs site.

### DIFF
--- a/packages/docs/src/theme/Layout/index.tsx
+++ b/packages/docs/src/theme/Layout/index.tsx
@@ -4,29 +4,44 @@ import '@remotion/promo-pages/dist/tailwind.css';
 import Layout from '@theme-original/Layout';
 import type LayoutType from '@theme/Layout';
 import {useCrawlChatSidePanel} from 'crawlchat-client';
-import React, {type ReactNode} from 'react';
+import React, {useEffect, type ReactNode} from 'react';
 
 type Props = WrapperProps<typeof LayoutType>;
 
 const isStandaloneRoute = (pathname: string) =>
 	pathname === '/experimental_new' || pathname.startsWith('/experimental_new/');
 
+const CRAWLCHAT_SCRIPT_ID = 'crawlchat-script';
+
+// The script must be injected imperatively rather than rendered as JSX:
+// embed.js declares a top-level `class CrawlChatEmbed` and mutates the DOM
+// on load, so having it in the server-rendered HTML breaks hydration
+// (React error #418) and any re-render of the <script> element executes it
+// a second time ("Identifier 'CrawlChatEmbed' has already been declared").
+const useCrawlChatScript = () => {
+	useEffect(() => {
+		if (document.getElementById(CRAWLCHAT_SCRIPT_ID)) {
+			return;
+		}
+
+		const script = document.createElement('script');
+		script.src = 'https://crawlchat.app/embed.js';
+		script.async = true;
+		script.id = CRAWLCHAT_SCRIPT_ID;
+		script.dataset.id = '67c0a28c5b075f0bb35e5366';
+		script.dataset.sidepanel = 'true';
+		script.dataset.small = 'true';
+		document.body.appendChild(script);
+		// No cleanup: removing the tag would not undo the global declaration,
+		// and re-inserting it later would throw.
+	}, []);
+};
+
 const LayoutWithCrawlChat = (props: Props): ReactNode => {
 	useCrawlChatSidePanel({history: useHistory()});
+	useCrawlChatScript();
 
-	return (
-		<>
-			<Layout {...props} />
-			<script
-				src="https://crawlchat.app/embed.js"
-				async
-				id="crawlchat-script"
-				data-id="67c0a28c5b075f0bb35e5366"
-				data-sidepanel="true"
-				data-small="true"
-			/>
-		</>
-	);
+	return <Layout {...props} />;
 };
 
 const LayoutWrapper = (props: Props): ReactNode => {


JonnyBurger:
Thanks!

PR #9008, opened 2026-07-11
fix(renderer): apply backpressure to parallel encoding

### LINKED ISSUE #6041: Correctly implement backpressure for parallel encoding
> Igor Samokhovets — Yesterday at 21:53
> i'm just afraid for users in the new banger.show update

> because if their phyiscal RAM is full, if they render ~1h of full hd video with dynamic visuals, their disk might be fucked lol
> i just experienced this myself, 70GB of frames on the disk
> Jonny Burger — 09:03
> hmm, this is a problem indeed
> upon reflection, I now realize that this is just because I did not understand how video encoding works when I implemented this
for clarification, this mechanic was always there
> or at least since a couple of years
> so, why do we even measure the memory that is available on one's computer?
> and fall back to writing frames to disk when there is not a huge amount of memory available?
> my suspicion is that when I implemented parallel encoding, I observed the memory filling up because of lack of backpressure
> the frames were rendering faster that they were encoding, and so a lot of memory was used over time, which led me to implement a check > whether there is enough memory
> while the correct solution would have been to slow down the rendering of the frames instead

dexhunter:
Fixes #6041.

## Summary

- Await the prestitcher stream's `drain` event when `stdin.write()` applies
  backpressure, so frame rendering cannot outrun FFmpeg.
- Use the bounded streaming path for supported codecs even when the memory
  estimate is low, avoiding the temporary PNG/JPEG frame spool.
- Propagate stream, cancellation, and page errors while cleaning temporary
  listeners on every path.
- Add focused coverage for drain, close, async error, synchronous `EPIPE`,
  callback awaiting/rejection, and page-error races.

## Measurements

The strict production-path evaluator reduced pending frame plus Writable queue
bytes from `9,437,184` to `196,608` (`-97.9%`) while preserving exact frame order
and content hashes.

An eight-schedulable-CPU real-render follow-up produced:

| Profile | Temporary frames | Runtime | Full process-tree RSS |
| --- | ---: | ---: | ---: |
| 1080p, 600 frames | 72.1 MiB -> 0 B | 25.8 s -> 19.0 s | 1.30 GiB -> 2.07 GiB |
| 4K, 240 frames | 57.5 MiB -> 0 B | 34.1 s -> 23.8 s | 2.80 GiB -> 4.60 GiB |

Both profiles kept exact frame counts and durations, matching output sizes, and
matching elementary H.264 hashes. No child processes survived.

This fixes the unbounded disk spool and bounds the producer queue; it is not a
claim of lower process RSS. Running Chromium and FFmpeg concurrently increased
tree RSS by 59.7% at 1080p and 64.0% at 4K in this hardware-sensitive test.

Autoresearch with Weco explored the drain, error, and cancellation variants that
shaped this patch. The final diff was manually reduced and hardened against the
strict evaluator: https://dashboard.weco.ai/share/Ksvw4gVC2CZXsxBN8WbPPN5NUy9nxH06

## Validation

- Strict order/hash, bounded-queue, low-memory, cancellation, page-error,
  stream-error, close, `EPIPE`, and listener-cleanup gates pass.
- Seven focused helper and frame-error tests pass.
- Renderer ESLint, TypeScript build, `oxfmt --check`, and `git diff --check`
  pass.
- On latest main `f1a07cd164`, the full renderer suite reports 294 pass / 43
  fail for the candidate versus 287 pass / 43 fail for baseline. The normalized
  failure set is identical; the seven additional passes are the new tests.

### DIFF
--- a/packages/renderer/src/race-with-frame-error.ts
+++ b/packages/renderer/src/race-with-frame-error.ts
@@ -1,0 +1,16 @@
+export const raceWithFrameError = async <T>({
+	cleanup,
+	frameError,
+	operation,
+}: {
+	cleanup: () => void;
+	frameError: Promise<never>;
+	operation: () => Promise<T> | T;
+}): Promise<T> => {
+	try {
+		return await Promise.race([Promise.resolve().then(operation), frameError]);
+	} catch (error) {
+		cleanup();
+		throw error;
+	}
+};
--- a/packages/renderer/src/render-frame-and-retry-target-close.ts
+++ b/packages/renderer/src/render-frame-and-retry-target-close.ts
@@ -11,11 +11,11 @@ import {getRetriesLeftFromError} from './is-delay-render-error-with-retry';
 import type {LogLevel} from './log-level';
 import {Log} from './logger';
 import type {CancelSignal} from './make-cancel-signal';
-import {cancelErrorMessages, isUserCancelledRender} from './make-cancel-signal';
+import {isUserCancelledRender} from './make-cancel-signal';
 import type {NextFrameToRender} from './next-frame-to-render';
 import type {Pool} from './pool';
 import {renderFrame} from './render-frame';
-import type {FrameAndAssets, OnArtifact} from './render-frames';
+import type {FrameAndAssets, OnArtifact, OnFrameBuffer} from './render-frames';
 import type {BrowserReplacer} from './replace-browser';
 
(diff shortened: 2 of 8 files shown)


JonnyBurger:
Sorry, this part of Remotion is too delicate that we want it to be addressed by AI + first time external contributor.

Feel free to tackle another issue! Thanks for sending this in!

Issue #9015, opened 2026-07-11
Studio: Hot-reload public dir config changes


JonnyBurger:
## Problem

#9011 can reload the Remotion config file and refresh the Studio page, but `Config.setPublicDir()` is still effectively captured when the Studio server starts.

## Why this blocks #9011

A page reload can pick up config values that are serialized into the Studio HTML or otherwise evaluated at request time. The public directory is different because it is currently threaded through server-side state that lives for the duration of the Studio process:

- static asset serving under the Studio static hash route
- `getStaticFileFallbackHint()`
- the public folder listing sent to Studio
- static-file write/delete/rename routes
- the filesystem watcher created by `initPublicFolderWatch()`

Changing `Config.setPublicDir()` in `remotion.config.ts` and reloading the page will not switch all of these server-side consumers to the new directory unless the watcher and route plumbing become runtime-aware too.

## Possible approach

Replace the captured `publicDir` string in the Studio server with a `getPublicDir()` callback based on the current config, and teach the public-folder watcher to tear down and recreate itself when the resolved directory changes. After switching directories, emit a `new-public-folder` event so the Studio UI refreshes its static file tree.

JonnyBurger:
A test PR introduced too much technical debt for this.
Let's just continue to require a restart when changing the public dir.

PR #9017, opened 2026-07-11
`@remotion/studio`: Reload Studio when config changes

### LINKED ISSUE #9011: Config file changes should reload
When we change the remotion.config.ts file, it should be reloaded.
I guess a reload of the Studio is in order for now, since there can not be unsaved changes anymore.

Some things might not be hot reloadable, like Webpack config.
For that, we can just ignore it.

From the user perspective, it should work similarly as `.env`

JonnyBurger:
## Summary

- Reload the loaded Remotion config file when it changes, reset previously applied config options, and trigger a Studio page reload.
- Inject Studio runtime config through generated HTML and read it via explicit Studio helpers instead of relying on `process.env` in Studio UI code.
- Add reset coverage for config options that need manual cleanup before reloading.

## Test plan

- `bun run build`
- `bun run stylecheck`
- `bun test packages/cli/src/test/config-reset.test.ts`

Closes #9011

Public dir hot-reload remains tracked separately in #9015.

### DIFF
--- a/packages/browser-studio/src/BrowserStudio.tsx
+++ b/packages/browser-studio/src/BrowserStudio.tsx
@@ -144,6 +144,14 @@ export const BrowserStudio: React.FC<BrowserStudioProps> = ({
 				renderQueue: [],
 				sampleRate: null,
 				staticHash: '',
+				studioRuntimeConfig: {
+					askAIEnabled: false,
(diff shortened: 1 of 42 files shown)


JonnyBurger:
Superseded by stacked PRs #9022 and #9024.

PR #9034, opened 2026-07-12
`@remotion/renderer`: Make calculateAssetPositions() faster 

dexhunter:
## Why

`calculateAssetPositions()` converts per-frame audio/video entries into positioned segments for stitching. PR #4575 removed one repeated flattening cost, but the remaining path still scanned previous/next frames, every open segment, and the flattened list for compressed references.

On long renders with many short or reopened assets, that work grows quadratically.

## What

- deduplicate each frame with a `Set`;
- track open segments by asset id in a `Map`;
- pre-index only compressed references that actually occur, preserving the first eligible source match;
- close segments when an id disappears and close remaining segments after the final frame.

The function remains behavior-compatible with corrupt compressed references by falling back to the existing diagnostic/error path. No public API changes.

## Performance

Measured against `main` at `69d5a16b`; the relevant baseline source is unchanged on the rebased current main. Seven alternating runs on the same pinned CPU used an unchanged 40,000-frame workload with 172,156 entries and 6,005 output segments. Every output matched the same full-result SHA-256.

| Implementation | Median | MAD |
|---|---:|---:|
| Current `main` | 937.182 ms | 1.157 ms |
| This change | 70.084 ms | 5.015 ms |

Median speedup: **13.372x**.

The late-source/reopened compressed fixture also scaled from `80.449 -> 3.338 ms` at 8,000 frames, `249.909 -> 6.470 ms` at 16,000, and `918.214 -> 13.641 ms` at 32,000.

## Correctness

- 1,204 differential cases matched current `main` exactly;
- 2,408 checks found no input mutation;
- tests cover gaps/reopening, durations, `trimLeft`, compressed cross-id sources, first-match behavior, output ordering, and within-frame deduplication;
- focused asset tests: 18 passed;
- `bun run build`: 71/71 tasks passed;
- `bun run stylecheck`: 233/233 tasks passed;
- full renderer suite on the rebased head: 318 passed; 18 environment/setup failures were unrelated to this patch;
- `git diff --check` passed.

Weco identified the open-segment map direction. A stronger scaling review then found repeated compressed-reference lookup in that first result; the final patch adds the reference index and revalidates the complete algorithm. [Public autoresearch trajectory](https://dashboard.weco.ai/share/3drdEMvQWxlFgajm2b47ijbmw2EiWgyW)

## Tradeoff

The function now allocates a `Set` of distinct compressed references and a `Map` containing their first eligible source entries. Auxiliary memory is proportional to distinct referenced assets.

### DIFF
--- a/packages/renderer/src/assets/calculate-asset-positions.ts
+++ b/packages/renderer/src/assets/calculate-asset-positions.ts
@@ -5,28 +5,16 @@ import {convertAssetToFlattenedVolume} from './flatten-volume-array';
 import type {Assets, MediaAsset, UnsafeAsset} from './types';
 import {uncompressMediaAsset} from './types';
 
-const areEqual = (a: TRenderAsset | UnsafeAsset, b: TRenderAsset) => {
-	return a.id === b.id;
-};
-
-const findFrom = (target: TRenderAsset[], renderAsset: TRenderAsset) => {
-	const index = target.findIndex((a) => areEqual(a, renderAsset));
-	if (index === -1) {
-		return false;
-	}
-
-	target.splice(index, 1);
-	return true;
-};
-
-const copyAndDeduplicateAssets = (
+const deduplicateAssets = (
 	renderAssets: TRenderAsset[],
 ): AudioOrVideoAsset[] => {
 	const onlyAudioAndVideo = onlyAudioAndVideoAssets(renderAssets);
+	const seenIds = new Set<string>();
 	const deduplicated: AudioOrVideoAsset[] = [];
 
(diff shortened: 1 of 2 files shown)


dexhunter:
Rebased onto current main at 2ac65701 with an identical patch-id. Focused tests passed 6/6; bun run build passed 71/71 tasks; bun run stylecheck passed 233/233 tasks; renderer typecheck and git diff --check passed. The full renderer suite passed 318 tests; 18 environment/setup cases were unrelated to this patch.

JonnyBurger:
Very cool!

PR #9042, opened 2026-07-13
Preserve pitch with WSOLA in `<Audio>` preview and render

### LINKED ISSUE #8021: `preservePitch` / tone shifting for `@remotion/media`
For this we need to register an audio worklet, similar to https://www.npmjs.com/package/@soundtouchjs/audio-worklet

Then we need to implement phase-vocoder / WSOLA / RubberBand/SoundTouch-style processing

Vanilagy recommends this Gist: https://gist.github.com/Vanilagy/05f7901f4c4398356657e3a86c7aee05
Says 200x realtime 🔥

iliya-vidrush:
Closes #8021.

## Summary

- Make `<Audio>`'s `preservePitch` and `toneFrequency` sound identical in preview and render by sharing one duration-preserving pitch shifter. `preservePitch` now affects native `@remotion/media` playback and rendering, not just the `<Html5Audio>` fallback.
- New `WsolaPitchShifter` (clean-room WSOLA) and `computePitchRatio` live in `@remotion/core` and are used by both pipelines: the preview routes audio through a `remotion-pitch-shifter` AudioWorklet, and the renderer replaces the ffmpeg tone-frequency pass with the same class. When the ratio is `1` the shifter is bypassed (bit-exact passthrough).

## Validation

- Unit tests for `computePitchRatio`, `WsolaPitchShifter`, and `applyPitchShiftToWav`.
- CI green.


samohovets:
Thanks for working on this and exploring WSOLA. After reviewing the implementation, we’ve decided to handle pitch preservation internally rather than continue iterating on this PR.

The main issues are:

- The renderer reads the complete WAV and creates multiple full-size buffers. This can consume several gigabytes for long audio and blocks the event loop.
- The AudioWorklet uses an exhaustive correlation search and allocates during processing. Its real-time performance has not been demonstrated and is likely insufficient at some ratios/channel counts.
- The linear resampling stage has no anti-alias filtering, which can noticeably degrade audio when pitching up.
- The tests mostly cover mono sine waves. The renderer test uses `pitchRatio: 1`, bypassing WSOLA, and there is no real AudioWorklet or preview-versus-render comparison.
- Putting the DSP implementation in `remotion` core increases the foundational bundle and causes unrelated core code to be included in the generated worklet.
- The PR also contains unrelated player timing changes and refactors, making the feature harder to review safely.

WSOLA may still be part of the eventual solution, but addressing these points requires a substantial architectural redesign rather than incremental changes here. We’ll take over the implementation from the team side.

Thank you again for the investigation and contribution. Closing this PR.

PR #9058, opened 2026-07-15
`@remotion/studio`: Open layer in editor on double-click

### LINKED ISSUE #9057: Double clicking a layer should open the editor
I think remapping this to rename was not a good move. @samohovets 
Sure, this is a convention in Figma, but Figma has no "code" part.

Yes, I can right-click to open in the editor. I can also single-click and then click in the inspector.
But it now happened like 10 times that I double clicked a layer to see how the code representation looks like.
I cannot re-train my reflex.

When working with Remotion, I feel the urge to quickly switch between interactivity, agent and code.
I think it would be cool if we can embrace this quick switching and make it the primary action.

samohovets:
## Summary

- Restore opening a timeline layer in the configured editor on double-click
- Keep modifier-key double-click handling unchanged
- Keep sequence renaming available through its existing dedicated actions

## Validation

- `bun run build`
- `bun run stylecheck`
- `bunx turbo run lint test --filter='@remotion/studio'`

Closes #9057

### DIFF
--- a/packages/studio/src/components/Timeline/TimelineSequenceItem.tsx
+++ b/packages/studio/src/components/Timeline/TimelineSequenceItem.tsx
@@ -724,22 +724,21 @@ export const TimelineSequenceItem: React.FC<{
 		codeHiddenStatus !== null &&
 		codeHiddenStatus.status === 'static';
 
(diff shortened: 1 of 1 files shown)


JonnyBurger:
🙌

PR #9077, opened 2026-07-15
`@remotion/media`: Throw a helpful error when ProRes decoding is not …

sksingh2005:
## Summary
- Detects ProRes sources (`videoTrack.codec === 'prores'`) that the browser can't decode, at both decode sites — preview (`media-player.ts`) and render (`get-frames-since-keyframe.ts`).
- Replaces the generic "could not be decoded" fallback with a targeted error naming `registerProresDecoder()` from `@mediabunny/prores` and linking to `/docs/videos/prores`.
- Message explicitly distinguishes decoding a ProRes *source* from *exporting* ProRes, so playback support isn't confused with export support.
- ProRes now **hard-fails by default** (throws in preview, `cancelRender` in render), ignoring `disallowFallbackToOffthreadVideo`.
- Routed through the standard error pipeline, so `onError` still fires and an explicit `onError` returning `'fallback'` can opt back into `<OffthreadVideo>`.
- New `cannot-decode-prores` result threaded as a typed message across the broadcast channel (survives the worker→main-tab hop, unlike an error class).
- Breaking: renders that silently relied on the OffthreadVideo/FFmpeg fallback for ProRes now throw until the decoder is registered.

## Verification
- `tsc --noEmit` on `@remotion/media` passes with no new errors (3 remaining are pre-existing on `main`).
- `oxfmt src` reports formatting clean.
- `eslint src` reports 0 errors (1 non-blocking max-params warning, consistent with existing code in this package).
- Both preview and render paths verified to detect and hard-throw independently.

### DIFF
--- a/packages/media/src/audio/audio-for-rendering.tsx
+++ b/packages/media/src/audio/audio-for-rendering.tsx
@@ -188,6 +188,12 @@ export const AudioForRendering: React.FC<AudioProps> = ({
 					);
 				}
 
+				if (result.type === 'cannot-decode-prores') {
+					throw new Error(
+						`Encountered ProRes media for ${src}. But this should never happen, since you used the <Audio> tag. Please report this as a bug.`,
+					);
+				}
+
 				if (result.type === 'network-error') {
 					handleError(
 						new Error(`Network error fetching ${src}.`),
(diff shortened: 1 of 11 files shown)


JonnyBurger:
This is indeed a problem - it should not fallback to OffthreadVideo. It can also not play it.
Please send a working PR if you want to send one

PR #9082, opened 2026-07-15
`@remotion/media`: Throw a helpful error when ProRes decoding is not enabled

### LINKED ISSUE #8955: `@remotion/media`: Throw a helpful error when ProRes decoding is disabled
Throw a targeted error when `@remotion/media` encounters ProRes media but the ProRes decoder is not enabled.

## Problem

If a user renders or previews a ProRes source through the `@remotion/media` tags without enabling the ProRes decoder, the current failure mode should be replaced or augmented with a clearer actionable message.

## Expected behavior

- Detect that the source requires ProRes decoding.
- Throw an error that names the missing ProRes decoder/setup.
- Link to the dedicated ProRes documentation page once it exists.
- Keep the message specific enough that users do not confuse ProRes rendering support with browser-native playback support.

## Notes

This should cover both preview and rendering paths if they can fail independently.

sksingh2005:
## Summary
- Detects ProRes sources (`videoTrack.codec === 'prores'`) the browser can't decode, at both decode sites preview (`media-player.ts`) and render (`get-frames-since-keyframe.ts`).
- Replaces the generic "could not be decoded" fallback with a targeted error naming `registerProresDecoder()` from `@mediabunny/prores` and linking to `/docs/videos/prores`.
- Message explicitly distinguishes decoding a ProRes *source* from *exporting* ProRes, so playback support isn't confused with export support.
- ProRes is treated as an **unrecoverable** error: it **never** falls back to `<OffthreadVideo>` (render) or a native `<video>` (preview), since neither can decode ProRes  registering the decoder is the only fix.
- Preview throws a dedicated `ProResDecoderNotEnabledError` that the `.catch` recognizes and rethrows (surfacing via the Studio error overlay) instead of resolving to a fallback; render calls `cancelRender`. `onError` is still notified exactly once.
- New `cannot-decode-prores` result threaded as a typed message across the broadcast channel (survives the worker→main-tab hop, unlike an error class).
- renders that previously relied on the silent OffthreadVideo fallback for ProRes now fail until the decoder is registered.

## Verification
- `tsc --noEmit` on `@remotion/media` passes with no new errors.
- `oxfmt src` clean; `eslint src` reports 0 errors.
- Confirmed no error boundary wraps the video components, so the thrown error can't be silently converted back into a fallback.

### Fixes - #8955

### DIFF
--- a/packages/media/src/audio/audio-for-preview.tsx
+++ b/packages/media/src/audio/audio-for-preview.tsx
@@ -277,6 +277,14 @@ const AudioForPreviewAssertedShowing: React.FC<NewAudioForPreviewProps> = ({
 						return;
 					}
 
+					if (result.type === 'cannot-decode-prores') {
+						// ProRes is video-only, so the MediaPlayer never returns this for
+						// an audio track. Guard it to keep the union exhaustive.
+						throw new Error(
+							`Encountered ProRes media for ${preloadedSrc}. But this should never happen, since you used the <Audio> tag. Please report this as a bug.`,
+						);
+					}
+
 					if (result.type === 'success') {
 						setMediaPlayerReady(true);
 						setMediaDurationInSeconds(result.durationInSeconds);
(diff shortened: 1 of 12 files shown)


JonnyBurger:
Thanks!

PR #9130, opened 2026-07-16
Elements: Add Location Lower Third

### LINKED ISSUE #8620: Remotion Element: Live / Location Strap
Part of: #7698

Category: Identity & Attribution

## Element

Live / Location Strap

## Purpose

A broadcast-style strap that communicates live status, location, venue, date, or event context.

## Real-world use case

Used in live streams, news segments, sports coverage, event recaps, conference videos, and YouTube explainers to establish where/when footage is happening.

Examples:

- `LIVE · Berlin`
- `On location · NAB Show`
- `June 22 · Main Stage`

## Requirements

- Zero dependency: Remotion + React + CSS only
- Designed to be placed inside a `<Sequence>`
- Small, focused, and composable
- Should support live/on-location/event-status variants
- Safe to overlay on footage without taking over the full frame
- No unexpected global styles or layout side effects
- Remixable source code users can edit
- Single self-contained `.tsx` file
- No local imports except Remotion / React / allowed npm packages
- All sample content and default values must be inlined in the TSX file
- Do not create separate data/config files, JSON fixtures, or imported local constants
- Use `Interactive.withSchema()` for editable fields
- The Element should work when copied into a Remotion project as-is

## Editable fields

Expose these through `Interactive.withSchema()` with inline defaults:

- `statusLabel`
- `location`
- `venue`
- `date`

Optional/empty values should be handled gracefully so the Element can cover `LIVE · Berlin`, `On location · NAB Show`, and event/date-only variants without requiring separate data.

## Acceptance criteria

- Element implementation exists
- Demo/preview exists
- Editable fields are wired through `Interactive.withSchema()`
- Source code is linked or embedded as expected for Remotion Elements
- Example usage is documented

samohovets:
## Summary

- add a self-contained Location Lower Third with an animated location pin and backgroundless layout
- animate the location and venue with staggered typewriter reveals and smooth fades
- expose the location, venue, and pin as named `Interactive` layers for Visual Mode editing
- rename the existing Lower Third to Name Lower Third for clearer gallery naming
- register the Location Lower Third in the gallery, sidebar, preview definitions, and Element page

## Verification

- `bun run build`
- `bun run stylecheck`
- `cd packages/docs && bun test src/test/elements.test.ts`
- rendered and inspected `element-overlays-location-lower-third`

Closes #8620


JonnyBurger (changes requested):
Seems a bit lackluster to me.
Would anyone every wanna use this?

samohovets:
@JonnyBurger my motivation here is to make the Elements catalog feel timeless and grounded in established media patterns, rather than filling it with promo-video effects that look good in a demo but rarely get reused.

This kind of location/context strap is used in conference and event coverage, documentaries, interviews on location, sports recaps, keynote clips, news-style content, and YouTube explainers.

The reusable pattern is basically **status + location + date + venue/context**.

I do think `LIVE` is probably not the best default example, since Remotion is usually used for rendered videos rather than actual live broadcasts. Something like this would probably represent the use case better 💡 

`DAY 2 · Berlin`
`Local First Conf · July 14`

So I don’t see it as a literal live-broadcast graphic. More as a small, reusable context overlay based on a media pattern that has been used for decades and still fits a lot of real Remotion videos.

JonnyBurger (approved):

PR #9159, opened 2026-07-17
`remotion`: Improve shared audio tag pool performance

Octember:
## Summary

Improves the shared audio tag pool in three independent ways:

- memoizes the pool's `<audio>` elements instead of recreating the element array whenever the provider rerenders
- persists refreshed `audioMounted` state for every pool record, avoiding repeated reconciliations after refs mount
- avoids assigning the empty audio data URL when an unused tag already has it

## Motivation

We found these while profiling a Remotion `<Player>` with `numberOfSharedAudioTags={64}` and premounted audio:

- 11,648 audio React element creations in 182 playback commits (64 tags per commit)
- 1,923 `emptied` events in a 15-second device trace
- repeated calls through the native media `src` setter from shared-audio reconciliation

The stale `audioMounted` flag and unconditional `src` assignment compound: once a mounted record's cached flag is stale, later updates keep reconciling the whole pool, and each reconciliation reloads every unused tag.

## Testing

- `cd packages/core && bun run test` (576 passing)
- `cd packages/core && bun run lint` (passes with one pre-existing `use-lazy-component.ts` warning)

## Tradeoffs

The memoized element array retains the same lifetime as `refs`, which is already memoized by audio context and pool size. No public API or rendered output changes.

### DIFF
--- a/packages/core/src/audio/shared-audio-tags.tsx
+++ b/packages/core/src/audio/shared-audio-tags.tsx
@@ -525,7 +525,10 @@ export const SharedAudioTagsContextProvider: React.FC<{
 			}
 
 			if (data === undefined) {
-				current.src = EMPTY_AUDIO;
+				if (current.src !== EMPTY_AUDIO) {
+					current.src = EMPTY_AUDIO;
+				}
+
 				return;
 			}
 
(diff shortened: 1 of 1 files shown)


JonnyBurger:
Thanks!

You might also benefit from using the new <Video> tag which all runs on 1 audio context.

PR #9161, opened 2026-07-17
Docs: Mention ProRes in comparison of video tags

### LINKED ISSUE #9154: Mention ProRes in comparison of video tags

stantheman0128:
Closes #9154

On the [Comparison of video tags](https://www.remotion.dev/docs/video-tags) page, the supported codecs cell for `<Video />` / `<Audio />` from `@remotion/media` did not list ProRes, while the `<OffthreadVideo />` column did. Since v4.0.487, `<Video />` can decode ProRes once the decoder from `@mediabunny/prores` is registered, so the table understated what `@remotion/media` supports.

This adds `ProRes ([must be enabled](/docs/videos/prores))` to the codec list, linking to the [Embedding ProRes](https://www.remotion.dev/docs/videos/prores) guide. The qualifier matches the wording on the [Supported media](https://www.remotion.dev/docs/media/support) page, which states that ProRes decoding can be enabled using `@mediabunny/prores`.

The diff touches every row of the table because Prettier realigns the column padding after the edited cell got longer. The only content change is in the supported codecs cell.

## Verification

The change is Markdown only, so I verified formatting and syntax rather than running TypeScript checks.

Prettier check with the version pinned in the root package.json (3.8.1) and the repo `.prettierrc.js`:

```
$ prettier --config .prettierrc.js --check packages/docs/docs/video-tags.mdx
Checking formatting...
All matched files use Prettier code style!
```

MDX syntax check, compiling the changed file with `@mdx-js/mdx` and `remark-gfm`:

```
MDX COMPILE OK
```

I also confirmed that applying only the one-cell edit to the file on `main` and running Prettier reproduces this branch's file byte for byte, so the table realignment is exactly Prettier's output. The `/docs/videos/prores` link target exists and is the same path already used by `/docs/media/support` and `/docs/media`.

This fix was developed with AI assistance (Claude Code) and verified locally as shown above.


JonnyBurger (changes requested):
Let's mention this in a separate row below HLS, since it is so important.

Differentiate between media and offthread: The new video tag can also play ProRes during preview, offthread only during rendering

JonnyBurger:

PR #9183, opened 2026-07-17
Docs: Share Twoslash cache between worktrees

samohovets:
## Summary

- add a content-addressed Twoslash cache shared through Git's common directory so every worktree can populate and reuse it
- include the language, compiler/rendering configuration, package versions, lockfile/package metadata, and generated declarations in cache keys
- atomically publish checksum-validated entries while retaining worktree-local pruning and bounded shared garbage collection
- add coverage for environment fingerprints, linked worktrees, concurrent writers, corruption recovery, and garbage collection

## Performance

Measured on this checkout:

- cold prewarm: 35–52 seconds
- hydration from the shared cache: 0.67 seconds
- fully local warm cache: approximately 0.4 seconds

## Verification

- `bun run build`
- `bun run stylecheck`
- `bunx turbo run lint test --filter='@remotion/docusaurus-plugin' --no-update-notifier`
- real cold, shared-hydration, and local-warm prewarm runs

### DIFF
--- a/packages/docs/prewarm-twoslash.ts
+++ b/packages/docs/prewarm-twoslash.ts
@@ -1,16 +1,23 @@
 import type {ChildProcessWithoutNullStreams} from 'child_process';
 import {spawn} from 'child_process';
-import {createHash} from 'crypto';
 import {existsSync, mkdirSync, readdirSync, readFileSync, unlinkSync} from 'fs';
 import {createRequire} from 'module';
 import {availableParallelism, cpus, totalmem} from 'os';
-import {dirname, join, resolve} from 'path';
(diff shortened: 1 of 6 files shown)


samohovets:
Thanks for flagging this. `twoslash-worker.mjs` is only used as the entrypoint for `Bun.build()` in `prewarm-twoslash.ts`. The prewarm script waits for that build to produce `twoslash-worker-bundle.mjs`, then spawns Node with the generated bundle rather than the source file.

I also inspected the generated bundle and confirmed that it contains no `.ts` imports. Therefore Node 16 does not need native TypeScript support for this worker.

PR #9193, opened 2026-07-17
fix(studio-server): resolve composition location through local re-exports

### LINKED ISSUE #9172: Location of Composition is not always read correctly
After new automators homepage is merged:

packages/brand/src/Applications.tsx

`ApplicationSimpleApp`, `ApplicationVideoEditor`, `ApplicationRenderButton` compositions all point to Applications.tsx:1

I think it might trip up on `export function Applications() {`

Or because of the re-export of `export {ApplicationRenderButton, ApplicationSimpleApp, ApplicationVideoEditor};`

stantheman0128:
Closes #9172

## Summary

Studio was reporting the wrong source location for compositions that live behind a local re-export barrel, for example:

```ts
import {ApplicationSimpleApp} from './Applications/SimpleApp';
export function Applications() { /* ... */ }
export {ApplicationSimpleApp};
```

`ApplicationSimpleApp` resolved to `Applications.tsx:1` instead of `Applications/SimpleApp.tsx`, because the resolver understood `export {X} from './mod'` but not `import {X} from './mod'; export {X}`.

This change:

- Follows local `export {X}` specifiers (no `from`) back through their matching import
- Prefers the identifier `loc` when Recast leaves `loc` null on `export function` / `export class` nodes (so `Applications` lands on the real function line instead of line 1)

## Verification

```
$ bun test packages/studio-server/src/test/resolve-composition-component.test.ts
 37 pass
 0 fail
 155 expect() calls
Ran 37 tests across 1 file.
```

Includes a regression test that mirrors the Applications.tsx / Applications/SimpleApp.tsx layout from the issue.

### DIFF
--- a/packages/studio-server/src/helpers/resolve-composition-component.ts
+++ b/packages/studio-server/src/helpers/resolve-composition-component.ts
@@ -403,6 +403,56 @@ const findReExportTargets = ({
 	return targets;
 };
 
+const findLocalExportImportTarget = ({
+	ast,
+	exportName,
+}: {
+	ast: File;
+	exportName: string | 'default';
+}): ImportTarget | null => {
+	if (exportName === 'default') {
+		return null;
+	}
+
+	let localName: string | null = null;
+
+	recast.types.visit(ast, {
+		visitExportNamedDeclaration(astPath) {
+			if (localName) {
+				return false;
+			}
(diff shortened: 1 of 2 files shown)


JonnyBurger:

PR #9205, opened 2026-07-17
@remotion/timeline-utils: Fix Studio waveform for playbackRate > 1

### LINKED ISSUE #7353: Audio waveform is affected by playbackRate < 1, but not by playbackRate > 1
https://github.com/remotion-dev/remotion/blob/main/packages/example/src/OffthreadRemoteVideo/ChangingTrimBefore.tsx

that seems wrong to me.

https://github.com/user-attachments/assets/bf1ada98-7011-48ef-8d34-bcb926f12dc8

stantheman0128:
## Summary

- Studio timeline waveforms already stretched for `playbackRate < 1`, but when `playbackRate > 1` asked for media past EOF, `sliceWaveformPeaks` clamped to the available peaks and `drawBars` stretched that shorter slice across the full sequence width. Visually that looked stuck at the slow-rate stretch.
- Pad the sliced peak buffer with silence up to the requested window length so the real peaks compress into the left portion of the clip (media ends early at faster rates).
- Add regression tests for rate `< 1`, rate `> 1` with enough media, and rate `> 1` past EOF.

Fixes #7353

## AI assistance

This PR was written with AI assistance (Cursor/Grok). I confirmed the EOF clamp against current main, added tests, ran the commands below, and reviewed the diff before opening.

## Verification

```text
cd packages/timeline-utils
bun test src/test/slice-waveform-peaks.test.ts
# 6 pass, 0 fail, 18 expect() calls
```

## Test plan

- [x] `bun test packages/timeline-utils/src/test/slice-waveform-peaks.test.ts`
- [ ] Studio: open a media clip whose sequence duration matches full media length; set `playbackRate` below 1 and confirm the waveform stretches; raise it above 1 and confirm the waveform compresses (content denser, silence/empty after media ends) instead of staying stretched

### DIFF
--- a/packages/timeline-utils/src/audio-waveform/slice-waveform-peaks.ts
+++ b/packages/timeline-utils/src/audio-waveform/slice-waveform-peaks.ts
@@ -25,8 +25,23 @@ export const sliceWaveformPeaks = ({
 		(startTimeInSeconds + durationInSeconds) * TARGET_SAMPLE_RATE,
 	);
 
-	return peaks.subarray(
-		Math.max(0, startPeakIndex),
-		Math.min(peaks.length, endPeakIndex),
+	const startIndex = Math.max(0, startPeakIndex);
+	const availableEnd = Math.min(
+		peaks.length,
+		Math.max(startIndex, endPeakIndex),
 	);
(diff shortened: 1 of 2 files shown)


JonnyBurger:

PR #9207, opened 2026-07-17
Studio: Configure License modal that writes public license key

### LINKED ISSUE #9012: Modal for declaring license in Studio
We want to add more awareness for the Remotion License in the Studio.

With this, one can pass a license key in `remotion.config.ts` or 

```ts
Config.setPublicLicenseKey('your-license-key');
```
or if eligible for free license
```ts
Config.setPublicLicenseKey('free-license');
```

https://www.remotion.dev/docs/config#setpubliclicensekey

Idea: Have a subtle nudge in the top right corner "Configure License"

This opens a modal, where one is educated about the license.
If eligible for free license, we set `Config.setPublicLicenseKey('free-license')` in remotion.config.ts, which should auto-reload
Same for if one has a license key

stantheman0128:
## Summary

Closes #9012

Adds a subtle **Configure License** control in the Studio toolbar (top right). It opens a modal that reuses the existing web-render license education UI. Choosing the free license or pasting a valid company public key writes `Config.setPublicLicenseKey(...)` into `remotion.config.ts` (or `.js`) and restarts Studio so the key is picked up.

Previously, the License tab in the web-render modal only showed a copy-paste snippet. It never persisted the config for you.

### Behavior
- Nudge is hidden in read-only Studio and when `publicLicenseKey` is already set
- Free path writes `free-license`
- Company path requires a 55-character `rm_pub_` key (alphanumeric after the prefix), matching the existing Studio validation
- Server-side validation matches that shape; the writer uses `JSON.stringify` so the value cannot break out of the string literal
- Existing `Config.setPublicLicenseKey(...)` calls are replaced (including multiline / quote / backtick forms). Complex template literals with `${...} ` are rejected with a clear error instead of appending a second call
- Custom `--config` paths are not rewritten; only the default `remotion.config.ts` / `.js` under the project root

## Verification

`
$ bun test src/test/set-public-license-key-in-config.test.ts
bun test v1.3.11 (af24e281)

 14 pass
 0 fail
 20 expect() calls
Ran 14 tests across 1 file.
`

Covered: create missing config, append, replace single/double/backtick/multiline, prefer `.ts` over `.js`, accept realistic company key, reject breakout payload / short keys / non-alnum after prefix.

## Notes

Developed with AI assistance (Cursor / Grok). Independently re-checked after a first pass found that the initial writer accepted arbitrary `rm_pub_*` strings; that is tightened in the follow-up commit.

### DIFF
--- a/packages/studio-server/src/preview-server/api-routes.ts
+++ b/packages/studio-server/src/preview-server/api-routes.ts
@@ -34,6 +34,7 @@ import {reorderSequenceHandler} from './routes/reorder-sequence';
 import {handleRestartStudio} from './routes/restart-studio';
 import {saveEffectPropsHandler} from './routes/save-effect-props';
 import {saveSequencePropsHandler} from './routes/save-sequence-props';
+import {handleSetPublicLicenseKey} from './routes/set-public-license-key';
 import {splitJsxSequenceHandler} from './routes/split-jsx-sequence';
 import {subscribeToDefaultProps} from './routes/subscribe-to-default-props';
 import {subscribeToFileExistence} from './routes/subscribe-to-file-existence';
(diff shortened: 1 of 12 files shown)


JonnyBurger:

PR #9211, opened 2026-07-17
Docs: Warn against matching composition FPS to video FPS

### LINKED ISSUE #8815: Matching video FPS is a bad idea
matching FPS is a bad idea generally because video could be variable FPS
improvment opportunity: don't enocurage it in https://www.remotion.dev/docs/mediabunny/metadata

mediabunny could maybe help here as well detect videos which have variable FPS. I filed an issue previously https://github.com/Vanilagy/mediabunny/issues/320

but if the containers have rounding errors, it's not possible to derive such a field without also specifying a tolerance

stantheman0128:
Closes #8815

## Summary

Adds a warning on the Mediabunny metadata docs page: do not set composition `fps` from the estimated video FPS to match the source.

Reasoning matches the issue: many videos are variable frame rate, and packet-based FPS is only an estimate (also affected by container timestamp rounding). Remotion samples media by timeline time, so a standard composition FPS (`24` / `25` / `30` / `60`) is the right default.

Upstream VFR detection in Mediabunny (Vanilagy/mediabunny#320) is left for a follow-up once that API exists.

## Verification

Prettier 3.x with repo MDX options (`useTabs: false`, `printWidth: 300`):

```
Checking formatting...
All matched files use Prettier code style!
```

Touched file:

- `packages/docs/docs/mediabunny/metadata.mdx`

Docs-only change; no package behavior changed.

### DIFF
--- a/packages/docs/docs/mediabunny/metadata.mdx
+++ b/packages/docs/docs/mediabunny/metadata.mdx
@@ -18,16 +18,7 @@ then [Mediabunny](https://mediabunny.dev) is the ideal library for doing so.
 ```tsx twoslash title="get-media-metadata.ts"
 import {Input, ALL_FORMATS, UrlSource} from 'mediabunny';
 
-const commonFpsValues = [
-  24000 / 1001,
-  24,
-  25,
-  30000 / 1001,
-  30,
-  50,
-  60000 / 1001,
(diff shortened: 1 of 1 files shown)


JonnyBurger:

PR #9213, opened 2026-07-17
@remotion/studio: Filter assets in the Assets panel

### LINKED ISSUE #8427: Studio: Should be able to filter assets
Image

Having a search bar here would be nice

stantheman0128:
## Summary

- Adds a Filter assets search field above the Studio Assets tree
- Filters `public` files by path (case-insensitive substring)
- Expands ancestor folders for matches so nested hits stay visible
- Shows an empty state when nothing matches

Fixes #8427

## AI assistance

This PR was written with AI assistance (Cursor/Grok). I confirmed the Assets panel had no filter UI at HEAD, implemented the helper + Studio wiring, ran the commands below, and reviewed the diff before opening.

## Verification

```text
cd packages/studio
bun test src/test/filter-static-files.test.ts
# 3 pass, 0 fail, 5 expect() calls
```

## Test plan

- [x] `bun test packages/studio/src/test/filter-static-files.test.ts`
- [ ] Studio Assets panel: type a filename fragment and confirm only matching assets remain
- [ ] Nested path match expands parent folders
- [ ] Clearing the filter restores the full tree and prior expand state


JonnyBurger:

PR #9216, opened 2026-07-17
@remotion/studio: Support `durationInFrames - n` in Interactivity

### LINKED ISSUE #9170: Can `interpolate(frame, [0, durationInFrames - 1], [0, 1])` be interactively edited?
This is interesting because it allows setting a keyframe based on the composition duration.

We recently supported multiplication values: `durationInFrames * 0.8` works lol, even though it is pretty useless.  
I think the subtraction usecase would be nice to have

stantheman0128:
Closes #9170

## Summary

Studio Interactivity already understands video-config multiplication such as `fps * 2` / `2 * fps` (#9139). This adds the subtraction form requested in #9170, so expressions like `durationInFrames - 1` stay editable in the timeline and inspector instead of falling back to a computed / non-editable value.

Parse and rewrite cover both operand orders (`durationInFrames - n` and `n - durationInFrames`). Moving a keyframe or updating a numeric prop keeps the subtraction shape when possible (for example dragging `durationInFrames - 1` to frame 100 becomes `durationInFrames - 20`).

## Verification

`bun test` in `packages/studio-server` for the related suites:

```
111 pass
0 fail
```

Files:

- `src/test/video-config-numeric-expression.test.ts`
- `src/test/update-keyframes.test.ts`
- `src/test/update-sequence-props.test.ts`
- `src/test/compute-sequence-props-status.test.ts`

### DIFF
--- a/packages/core/src/use-schema.ts
+++ b/packages/core/src/use-schema.ts
@@ -43,6 +43,14 @@ export type VideoConfigNumericExpression =
 			multiplicand: number;
 			factorPosition: 'left' | 'right';
 			value: number;
+	  }
+	| {
+			type: 'video-config-subtraction';
+			identifier: string;
+			amount: number;
+			configValue: number;
+			amountPosition: 'left' | 'right';
+			value: number;
 	  };
 
 export type CanUpdateSequencePropStatusLinearEasing = {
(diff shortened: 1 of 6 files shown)


JonnyBurger:

PR #9220, opened 2026-07-17
@remotion/studio: Preserve subframe media duration caps

### LINKED ISSUE #7672: Timeline audio layer truncates subframe durations
Follow-up from #7668 / #7670.

When using the `subframe-audio` composition, the parent `Series.Sequence` with `durationInFrames={23.5}` now renders seamlessly in the timeline, but the nested `<Audio>` layer is still visualized as only 23 frames long.

Expected: The `<Audio>` layer should visually extend for the full 23.5 frames, matching the sequence/media duration.

Actual: The `<Audio>` layer appears truncated to 23 frames.

This should be handled separately from the timeline sequence boundary fix so we do not forget the remaining audio layer discrepancy.

stantheman0128:
## Summary

- Timeline `maxMediaDuration` no longer uses `Math.floor` on `duration * fps`, so media layers can keep fractional widths (for example 23.5 frames)
- Extracted `getMediaDurationInFrames` (same `toFixed(10)` rounding as `getTimelineDuration`)
- Added regression tests for nested `<Audio>` under a subframe `Series.Sequence` and for fractional max-media clipping

Fixes #7672

## AI assistance

This PR was written with AI assistance (Cursor/Grok). I reproduced the remaining gap after #7673: core duration caps already keep 23.5, but Studio still floored asset metadata when computing `maxMediaDuration`. I re-applied that Studio fix, added tests, and reviewed the diff before opening.

## Verification

`	ext
cd packages/core
bun test src/test/get-media-duration.test.ts
# 10 pass, 0 fail

cd packages/studio
bun test src/test/get-media-duration-in-frames.test.ts
# 2 pass, 0 fail
`

`timeline-sequence-layout.test.ts` also gained a floored-vs-fractional width assertion; that file needs a built `@remotion/studio-shared` locally, so CI should cover it.

## Test plan

- [x] `bun test packages/core/src/test/get-media-duration.test.ts`
- [x] `bun test packages/studio/src/test/get-media-duration-in-frames.test.ts`
- [ ] Open `subframe-audio` in Studio and confirm the nested Audio layer extends to 23.5 frames (matches the parent Series.Sequence)
- [ ] Spot-check a normal integer-duration audio clip still lays out correctly

### DIFF
--- a/packages/core/src/test/get-media-duration.test.ts
+++ b/packages/core/src/test/get-media-duration.test.ts
@@ -66,6 +66,21 @@ test('parentSequence with subframe duration caps without truncation', () => {
 	expect(duration).toBe(23.5);
 });
 
+// Nested <Audio> inside Series.Sequence: useVideoConfig() returns the parent
+// sequence duration (23.5), so both inputs are fractional. Must not floor to 23.
+test('nested media under subframe parent keeps fractional duration', () => {
+	const duration = getTimelineDuration({
+		compositionDurationInFrames: 23.5,
+		playbackRate: 1,
+		trimBefore: undefined,
+		trimAfter: undefined,
+		parentSequenceDurationInFrames: 23.5,
+		loop: false,
+	});
+
+	expect(duration).toBe(23.5);
+});
+
 test('subframe sequence duration is not truncated', () => {
 	const duration = getTimelineDuration({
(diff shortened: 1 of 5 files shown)


JonnyBurger:

PR #9222, opened 2026-07-18
Support numeric fontWeight (1 to 1000) in Interactivity (#8793)

### LINKED ISSUE #8793: `fontWeight: 650` is not supported by Interactivity

stantheman0128:
`@remotion/studio`: Support numeric CSS font-weight values in Interactivity

Fixes #8793.

## Problem

In Interactivity, `style.fontWeight` was modeled as an enum whose only allowed values were the nine 100-step weights (100 to 900) plus `normal` and `bold`. A valid CSS weight such as `fontWeight: 650` was therefore not recognized: the Studio rendered an enum dropdown, 650 was not one of the options, and the value could not be edited without being changed to a 100-step value.

CSS Fonts Level 4 allows any numeric font-weight from 1 to 1000, so 650 is valid and should not be coerced to 600 or 700.

A previous attempt (#8800) hardcoded 650 into the enum and was closed. This PR instead follows the approach the maintainer described on the issue: make font-weight a real numeric field with the string keywords kept as presets.

## What changed

- Added a dedicated `font-weight` field type to the Interactivity schema (`packages/core/src/interactivity-schema.ts`). It carries a numeric range (`min: 1`, `max: 1000`, `step: 1`) and a `presets` list (`normal`, `bold`, and 100 to 900). `style.fontWeight` now uses this type instead of `enum`.
- Registered the new type in `@remotion/studio-shared` (`schema-field-info.ts`) and kept it non-keyframable alongside the enum and font-family fields (`keyframe-interpolation-function.ts`).
- Added a Studio control (`TimelineFontWeightField.tsx`) that renders a numeric `InputDragger` (min 1, max 1000, step 1) plus a compact preset dropdown for `normal`, `bold`, and 100 to 900. This replaces the enum-only dropdown for font-weight.
- Moved the value parsing and the save decision into a small DOM-free helper (`font-weight-utils.ts`) so they can be unit tested directly.

Numeric values are read from and written back to the source as numbers, so an existing `fontWeight: 650` round-trips as the number 650. The keywords `normal` and `bold` are read and written as strings. The numeric field shows `normal` as 400 and `bold` as 700 for editing convenience, but the stored string is left untouched unless the user actually changes the value.

## Acceptance criteria

- `fontWeight: 650` recognized by Interactivity and shown in Studio: the field is no longer an enum, so 650 is read from props and displayed in the numeric control. Covered by `fontWeight: 650 is read from props and preserved as a number` in the core test.
- User can type 650, drag/edit it, and save it back as a numeric value: the control uses `InputDragger` and calls `onSave` with a JS number, which the codemod serializes as a numeric literal. Covered by `editing fontWeight to a new number saves a number, not a string` (core) and `editing to an arbitrary weight saves it as an exact number` (studio util test).
- User can still choose normal, bold, and common 100-step values: the preset dropdown offers `normal`, `bold`, and 100 to 900. Presets are asserted in `fontWeight is a numeric font-weight field, not a 100-step enum` (core).
- Values outside 1..1000 rejected or clamped per the existing number-field behavior: the control clamps to the field min/max. Covered by `values outside 1..1000 are clamped to the field range` (studio util test).
- Studio does not normalize arbitrary valid values to the nearest 100: the field uses `step: 1`, and the save path writes the exact value. Covered by the `step` assertion and by `editing to an arbitrary weight saves it as an exact number (no snapping)`.
- normal/bold are not rewritten unless the value changes: covered by `normal/bold are not rewritten when the numeric field is left unchanged` and `fontWeight string presets normal and bold survive as strings`.

## Verification

Targeted tests, run with the repo's Bun test runner:

```
$ bun test src/test/with-interactivity-schema-helpers.test.ts   (packages/core)
 24 pass
 0 fail
 68 expect() calls

$ bun test src/test/font-weight-utils.test.ts   (packages/studio)
 8 pass
 0 fail
 16 expect() calls

$ bun test src/test/schema-field-info.test.ts src/test/keyframe-interpolation-function.test.ts   (packages/studio-shared)
 14 pass
 0 fail
```

Type checking passed by building the affected packages (`turbo run make` runs `tsgo -d`): `remotion`, `@remotion/studio-shared`, and `@remotion/studio` all built successfully.

Caveat: the `InputDragger` visual appearance and drag interaction cannot be exercised in a headless environment, so I did not add a rendering or drag test for the Studio control. The control logic that decides what value to save (parsing, clamping, string-preset preservation) is covered by the unit tests above; only the pointer-drag UI itself is unverified here.

## AI assistance

This change was implemented with AI assistance (Claude) and reviewed before submission.

### DIFF
--- a/packages/core/src/interactivity-schema.ts
+++ b/packages/core/src/interactivity-schema.ts
@@ -112,6 +112,19 @@ export type FontFamilyFieldSchema = {
 	keyframable?: false;
 };
 
+export type FontWeightFieldSchema = {
+	type: 'font-weight';
+	min?: number;
+	max?: number;
+	step?: number;
+	default: number | string | undefined;
+	description?: string;
+	// Quick-select presets. Strings like 'normal'/'bold' are saved as strings,
+	// numbers like 400 are saved as numbers.
+	presets: readonly (number | string)[];
+	keyframable?: false;
+};
+
 export type EnumFieldSchema = {
 	type: 'enum';
 	default: string;
@@ -197,6 +210,7 @@ export type VisibleFieldSchema =
 	| ColorFieldSchema
 	| TextContentFieldSchema
 	| FontFamilyFieldSchema
+	| FontWeightFieldSchema
 	| ArrayFieldSchema
 	| EnumFieldSchema;
 
@@ -276,22 +290,13 @@ export const textSchema = {
 		hiddenFromList: false,
 	},
 	'style.fontWeight': {
-		type: 'enum',
-		default: '400',
+		type: 'font-weight',
+		default: undefined,
 		description: 'Font weight',
-		variants: {
-			'100': {},
-			'200': {},
-			'300': {},
-			'400': {},
-			'500': {},
-			'600': {},
-			'700': {},
-			'800': {},
-			'900': {},
-			normal: {},
-			bold: {},
-		},
+		min: 1,
+		max: 1000,
+		step: 1,
+		presets: ['normal', 'bold', 100, 200, 300, 400, 500, 600, 700, 800, 900],
 	},
 	'style.fontStyle': {
 		type: 'enum',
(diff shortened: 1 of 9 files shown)


JonnyBurger:

PR #9276, opened 2026-07-19
`remotion-monorepo`: Check gcloud authentication before releases

JonnyBurger:
## Summary

- verify that gcloud can issue an access token before starting a release
- require `gcloud auth login` and a successful recheck when authentication has expired

## Checks

- `bun run build`
- `bun run stylecheck`


JonnyBurger:
Superseded by #9278, which is based directly on main and contains only the intended release-skill change.

PR #9398, opened 2026-07-21
`@remotion/cli`: Fix package manager spawning on Windows

misak1:
## Summary
- Add shared package manager spawn options that use a shell on Windows so npm/pnpm/yarn/bun command shims can be resolved by child_process.spawn().
- Use the shared options for both the CLI add command and the Studio package installation route.
- Reject spawn errors instead of throwing inside async event handlers.

## Testing
- Ran git diff --check.
- Verified the failure mode in a Windows Remotion Studio project where Tools > Install package failed with spawn pnpm ENOENT.
- Verified the workaround in that project by installing @remotion/animation-utils from Studio in Chrome; the install completed successfully.

Note: I did not run the monorepo test suite locally because this clean upstream checkout does not have the Bun-based dependency environment installed.

### DIFF
--- a/packages/cli/src/add.ts
+++ b/packages/cli/src/add.ts
@@ -186,17 +186,23 @@ export const addCommand = async ({
 		stdio: RenderInternals.isEqualOrBelowLogLevel(logLevel, 'info')
 			? 'inherit'
 			: 'ignore',
+		...StudioServerInternals.getPackageManagerSpawnOptions(),
 	});
 
-	await new Promise<void>((resolve) => {
+	await new Promise<void>((resolve, reject) => {
+		task.on('error', (err) => {
+			reject(err);
+		});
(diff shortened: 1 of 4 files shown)


JonnyBurger:
What bug does this fix?

misak1:
This PR fixes an issue where attempting to install a package via "Tools > Install package" (or running `npx remotion add`) on Windows causes the application to crash / throw an uncaught error.

On Windows, executing package manager commands (such as `pnpm`, `npm`, `yarn`, or `bun`) requires running through a shell (`shell: true`) to properly resolve command shims (`.cmd` / `.bat`). Without this, `child_process.spawn()` fails with `ENOENT` and unhandled rejections cause the process to exit.

JonnyBurger:
Got it! Thanks for the explanation and for fixing it

PR #9428, opened 2026-07-22
Fix two ways the Player's playback loop can stall permanently while isPlaying() stays true

dukeeagle:
## Symptom

Scrubbing a media-heavy composition while it plays can leave the Player in a state where video and audio elements keep running but the frame clock never advances again: getCurrentFrame() returns the same value forever, isPlaying() returns true, the buffering flag is false, and no waiting/resume events fire. Only a seek or a pause/play cycle revives it. In our editor this reproduced within a minute of scrubbing on most sessions.

We instrumented the playback loop in production builds and traced the stall to two independent causes in how usePlayback parks and resumes itself. Both fixes are small; each was verified against its own failure trace.

## Cause 1: resume listeners registered through React state can miss the resume

useBufferManager keeps its listener arrays in useState, so listenForResume() only takes effect after the next commit. usePlayback registers its resume listener from a rAF callback while buffering is active. If the last buffering block unblocks before that registration commits, the resume dispatch iterates the previous callback array, the listener never fires, and the loop stays parked with nothing left to wake it. The window is a full React commit, which rapid scrub-driven buffering churn hits often.

Fix: keep the listener registries in refs so registration is visible to the dispatch of the same commit. Dispatch iterates a snapshot, preserving the existing guarantee that a callback removing itself does not skip the next one. Added a regression test that registers a listener from a child layout effect inside the commit that empties the block list; it fails on the previous implementation and passes now.

## Cause 2: waiting on AudioContext.resume() with no deadline

queueNextFrame() awaits sharedAudioContext.getIsResumingAudioContext() before scheduling the next frame. In Chromium, a resume() promise can stay pending indefinitely when a suspend() lands while the resume is still in flight, and the loop's own teardown calls suspend() on every effect re-run. During scrubbing the playback effect restarts constantly, so an orphaned resume promise is easy to produce, and the loop then waits on it forever.

Fix: race the wait against a 500ms deadline. Past the deadline the loop resumes ticking without audio (the bypass is sticky per loop instance, so it does not re-park on the same pending promise once per frame) and audio rejoins when the promise settles. Also added a hasBeenStopped guard to queueNextFrame so a torn-down loop cannot keep re-parking from a late promise resolution.

## Notes for review

- The trace that isolated cause 2 looked like: loop-start -> park on getIsResumingAudioContext -> nothing, repeated across sessions; after the deadline was added without the sticky bypass, the loop re-parked on the same pending promise once per timeout, so the bypass flag is load-bearing.
- I could not find a practical way to unit-test cause 2 without mounting a full Player against a mocked shared audio context; suggestions welcome. Cause 1 has a test in packages/core/src/test/buffering.test.tsx.
- bun test src/test/buffering.test.tsx passes (4 tests), and bun run make succeeds in packages/core and packages/player.

Opening as a draft to hear whether you would rather handle the audio wait with cancellation on suspend() instead of a deadline.

### DIFF
--- a/packages/core/src/buffering.tsx
+++ b/packages/core/src/buffering.tsx
@@ -43,12 +43,14 @@ const useBufferManager = (
 	mountTime: number | null,
 ): BufferManager => {
 	const [blocks, setBlocks] = useState<Block[]>([]);
-	const [onBufferingCallbacks, setOnBufferingCallbacks] = useState<
-		OnBufferingCallback[]
-	>([]);
-	const [onResumeCallbacks, setOnResumeCallbacks] = useState<
-		OnBufferingCallback[]
-	>([]);
+	// Listener registries are refs, not state: `usePlayback` parks its loop
+	// during buffering and registers its resume listener from a rAF callback.
+	// With state, that registration only lands after the next React commit -
+	// if the last block unblocks before then, the resume dispatch reads the
+	// previous array, the listener is never called, and the playback loop
+	// stays parked forever (frame clock frozen while isPlaying() is true).
+	const onBufferingCallbacks = useRef<OnBufferingCallback[]>([]);
+	const onResumeCallbacks = useRef<OnBufferingCallback[]>([]);
 
 	const env = useRemotionEnvironment();
 	const rendering = env.isRendering;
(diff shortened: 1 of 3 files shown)


dukeeagle:
Superseded by #9429 - same two fixes, clean two-commit history.

PR #9429, opened 2026-07-22
Fix two ways the Player playback loop can stall forever while isPlaying() stays true

dukeeagle:
## Symptom

Scrubbing a media-heavy composition while it plays can leave the Player with its frame clock stopped: getCurrentFrame() returns the same value forever, isPlaying() returns true, no waiting/resume events fire, and any still-mounted audio or video elements keep running on their own. Only a seek or a pause/play cycle revives it. In our editor this reproduced within a minute of scrubbing on most sessions.

We instrumented the playback loop in production builds and traced the stall to two independent causes in how usePlayback parks and resumes itself.

## Cause 1: resume listeners registered through React state can miss the resume

useBufferManager keeps its listener arrays in useState, so listenForResume() only takes effect after the next commit. usePlayback registers its resume listener from a rAF callback while buffering is active. If the last buffering block unblocks before that registration commits, the resume dispatch iterates the previous callback array, the listener never fires, and the loop stays parked with nothing left to wake it. The window is a full React commit, which rapid scrub-driven buffering churn hits often.

Fix (first commit): keep the listener registries in refs so registration is visible to the dispatch of the same commit. Dispatch iterates a snapshot, preserving the existing guarantee that a callback removing itself does not skip the next one. A regression test registers a listener from a child layout effect inside the commit that empties the block list; it fails on the previous implementation and passes now.

## Cause 2: the resume wait polls for output progress that can never come

queueNextFrame() awaits getIsResumingAudioContext() before scheduling the next frame. That promise is backed by waitUntilActuallyResumed(), a rAF poll that resolves once the AudioContext's output timestamp advances. Two situations leave that poll spinning forever, and isResuming stays set so every later playback loop parks on it too:

- A suspend() lands while the poll is running (routine during scrubbing, because the playback effect restarts constantly and every teardown calls suspend()). The context clock stops, so the timestamp can never advance.
- The resume() itself never takes effect while the intent to play remains, for example when an autoplay policy blocks a resume issued outside a user gesture.

Fix (second commit): the poll settles in both situations. suspend() already records the withdrawn intent in audioContextIsPlayingEventually, so waitUntilActuallyResumed() takes a shouldStillWait callback and resolves as soon as it returns false. For a resume that hangs while the intent persists, the poll gives up after one second with a warning and playback continues without audio sync; audio joins once the context recovers. Healthy resumes complete in well under 100ms, so the deadline only fires in the pathological case. The commit also adds a hasBeenStopped guard to queueNextFrame so a torn-down loop cannot reschedule itself from a promise that settles late.

## Notes for review

- Supersedes #9428 (same fixes; this branch has a clean two-commit history).
- The production trace that isolated cause 2: loop-start, park on getIsResumingAudioContext, then no further events, repeated across sessions, while an audio element kept playing.
- Cause 1 has a regression test in packages/core/src/test/buffering.test.tsx. I did not find a practical way to unit-test cause 2 without mounting a full Player against a mocked shared audio context; suggestions welcome.
- bun test src/test/buffering.test.tsx passes (4 tests); bun run lint and bun run make succeed in packages/core and packages/player.


dukeeagle:
CI note: the one red check (Build Node 16 on windows-latest) failed on the transition-series-stack test, which waits a fixed 10ms before asserting sequence registration. It passes 8/8 locally on this branch, does not involve the buffering or audio code this PR touches, and the same test failed once on macos-latest for the predecessor branch while passing everywhere else in the same runs - it looks load-sensitive on busy runners. Repushed to rerun.

JonnyBurger:
Thank you for the detailed investigation and for separating the two root causes so clearly. I split them into separate issues under the Remotion Core masterplan:

- #10170 — missed buffering resume events can leave the Player loop parked
- #10171 — an unresolved AudioContext resume wait can stall the Player frame clock

We will track the two fixes independently there, so I am closing this PR.

PR #9471, opened 2026-07-22
@remotion/renderer: Normalize paths before isPathInside check

santhiprakash:
## Problem

`isPathInside` did not resolve `.` or `..` segments before the prefix check. A path like `/foo/../bar` was accepted as inside `/foo` even though it resolves to `/bar`, which could bypass the path-traversal guard in the serve handler / preview server.

## Triage / Root cause

`path.relative` and the subsequent prefix checks operate on the literal string, not the resolved path. Relative segments therefore affected the outcome.

## Fix

- Call `path.normalize` on both the candidate path and the potential parent before comparing.
- Add unit tests in `packages/renderer/src/test/is-path-inside.test.ts` for child paths, traversal attempts, and trailing separators.

## Verification

- `bun test packages/renderer/src/test/is-path-inside.test.ts` passed.
- `bun run stylecheck` passed (only pre-existing warnings outside `renderer`).

## Notes / Risks

Security fix; no public API change. The function is used internally by the serve handler.

### DIFF
--- a/packages/renderer/src/serve-handler/is-path-inside.ts
+++ b/packages/renderer/src/serve-handler/is-path-inside.ts
@@ -4,6 +4,11 @@ export const isPathInside = function (
 	thePath: string,
 	potentialParent: string,
 ) {
+	// Resolve . and .. segments before checking containment so that a path
+	// containing traversal components cannot be mistaken for a child path.
+	thePath = path.normalize(thePath);
+	potentialParent = path.normalize(potentialParent);
+
 	// For inside-directory checking, we want to allow trailing slashes, so normalize.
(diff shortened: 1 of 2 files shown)


JonnyBurger:
Thanks for the contribution! I don’t think this currently changes behavior in the affected request paths: the renderer serve handler creates `absolutePath` with `path.join(current, relativePath)`, and the Studio static routes likewise call `path.join()` before `isPathInside()`. `path.join()` already normalizes `.` and `..` segments, so this helper does not receive the unnormalized traversal examples from its production callers. Since this is redundant hardening rather than fixing an active bypass, I’d prefer not to add it. Closing.

PR #9283, opened 2026-07-19
`@remotion/studio`: Add caption JSON inspector

JonnyBurger:
## Summary

- make `AnimatedCaptions` reusable through a required `src` prop
- expose caption sources as interactive asset fields in Remotion Studio
- add a Studio inspector and editor for caption JSON files
- keep the captions tester composition pointed at the existing captions JSON

## Checklist

- [ ] Update the rendered caption preview live as caption text is typed in the Studio editor
- [ ] Make the terminal diff less noisy when captions are changed

## Testing

- `bun run build`
- `bun run stylecheck`

PR #9513, opened 2026-07-23
`@remotion/studio`: Add interactive caption editing

samohovets:
## Summary

- add a dedicated Studio timeline mode for editing caption timings
- resize caption boundaries with frame snapping and adjacent-caption constraints
- round calculated caption timestamps to remove JavaScript floating-point artifacts while preserving microsecond precision
- select caption blocks to scroll the matching inspector phrase into view
- persist inline literal caption text and timing changes as targeted source patches, preserving untouched caption formatting and printing compact terminal diffs
- save caption text as it is typed and retain optimistic caption values when the selection is reopened before the runtime refreshes
- keep computed and imported caption arrays read-only
- make `Interactive.withCaptions()` inherit standard Sequence timing, visibility, and timeline behavior
- keep the example transcript in a dedicated component with an inline editable caption array
- remove the JSON-asset caption editor

## Testing

- `bun run build`
- `bun run stylecheck`
- `cd packages/core && bun test src/test/sequence-register-once.test.tsx`
- `cd packages/studio && bun test src/state/caption-timing-edit.test.ts src/components/Timeline/caption-timing.test.ts`
- `cd packages/studio-server && bun test src/test/update-inline-caption-patches.test.ts src/test/compute-sequence-props-status.test.ts src/test/save-sequence-props.test.ts`

## Stack

- Stacked on #9283


samohovets:
Moving this stack into a new stack in #9717

PR #9554, opened 2026-07-23
@remotion/renderer: Clarify --disable-headless does not work on Windows

### LINKED ISSUE #3332: --disable-headless does not seem to work
# Bug Report 🐛
Hi everyone

I have tried to render the helloworld sample video in non-headless mode. This is to investigate a problem with the <IFrame> component. (more on this later, I will open an issue about this)
I noticed that `--disable-headless` does not seem to work. After sending the render command `remotion render --disable-headless HelloWorld out/video.mp4` the browser window opens but closes immediately and the render process gets stuck.

Can someone verify this or is there something wrong with my installation?

Best regards

santhiprakash:
Closes #3332

The `--disable-headless` / `headless` option is already deprecated and documented as non-functional with Chrome Headless Shell, but the docs did not mention that it is also broken on Windows and can hang the render process. This change adds that note to the option description so it appears in the CLI help and the generated docs pages.

## Verification

- `bun run format` passed
- `bunx turbo run lint --filter=@remotion/renderer` passed
- `bunx turbo run formatting --filter=@remotion/renderer` passed
- `bun test src/test/boolean-option-precedence.test.ts` passed

### DIFF
--- a/packages/renderer/src/options/headless.tsx
+++ b/packages/renderer/src/options/headless.tsx
@@ -16,8 +16,12 @@ export const headlessOption = {
 			</a>
 			, this option is not functional anymore.
 			<br />
-			<br /> If disabled, the render will open an actual Chrome window where you
-			can see the render happen. The default is headless mode.
+			<br /> If disabled, the render was intended to open an actual Chrome
+			window where you can see the render happen. The default is headless mode.
+			<br />
+			<br />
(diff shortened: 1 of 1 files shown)


JonnyBurger:
Already covered by https://github.com/remotion-dev/remotion/pull/9509, thanks!

PR #9602, opened 2026-07-24
Docs: Remove duplicate --log section in benchmark CLI docs

santhiprakash:
## Problem

`packages/docs/docs/cli/benchmark.mdx` documented the `--log` flag twice: once after `--disable-web-security` and again after `--every-nth-frame`. The duplicate section is redundant and confusing because the flag ordering no longer mirrors `npx remotion render`.

## Triage / Root cause

A copy-paste or merge artifact left the second `### `--log`` block at lines 126-128. It is identical to the first occurrence and interrupts the logical CLI option grouping.

## Fix

Remove the duplicate `### `--log`` section that appears between `--every-nth-frame` and `--muted`.

## Verification

- `bunx oxfmt packages/docs/docs/cli/benchmark.mdx --check` passed.
- `packages/docs` formatting check passed.
- The change is docs-only and touches no runtime code.

## Notes / Risks

Docs-only change; no runtime behavior.

### DIFF
--- a/packages/docs/docs/cli/benchmark.mdx
+++ b/packages/docs/docs/cli/benchmark.mdx
@@ -123,10 +123,6 @@ Inherited from [`npx remotion render`](/docs/cli/render#--port)
 
 Inherited from [`npx remotion render`](/docs/cli/render#--every-nth-frame)
 
-### `--log`
-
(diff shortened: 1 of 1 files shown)


JonnyBurger:
Thanks!

PR #9622, opened 2026-07-24
`@remotion/web-renderer`: don't add silent AAC track for compositions without audio

### LINKED ISSUE #7099: `@remotion/web-renderer`: AAC encoder priming inflates container duration when emitting silent audio
## Problem

After #7074 (commit `3ce1f16a07`), `onlyInlineAudio` always returns an `AudioData` — even when the composition has zero audio assets. This means every render now includes an audio track of silent AAC frames. The AAC codec introduces **priming delay** (extra samples at the start for encoder initialization), which makes the audio track slightly longer than the video track. Since the MP4 container duration is `max(video_duration, audio_duration)`, the container ends up longer than expected.

## Reproduction

Render 10 frames (frame range `[10, 19]`) at 30 fps and check the resulting container duration:

```ts
const result = await renderMediaOnWeb({
  composition: { component: () => null, id: 'test', width: 100, height: 100, fps: 30, durationInFrames: 20 },
  inputProps: {},
  frameRange: [10, 19],
  outputTarget: 'arraybuffer',
  licenseKey: 'free-license',
});
const blob = await result.getBlob();
using input = new Input({ formats: ALL_FORMATS, source: new BlobSource(blob) });
const duration = await input.computeDuration();
// Expected: ~0.3333s (10 frames / 30 fps)
// Actual:   ~0.4053s
```

### Audio created per video frame

```
expectedLength = Math.round((2 channels × 48000 Hz) / 30 fps) = 3200 interleaved samples
numberOfFrames = 3200 / 2 = 1600 audio frames per video frame
```

### Total audio for 10 video frames

- 10 × 1600 = **16,000 audio frames** (per channel) = 16,000 / 48,000 = **0.3333 s** ✓

### AAC encoding overhead

AAC-LC encodes in fixed **1024-sample frames**. The WebCodecs `AudioEncoder` adds **priming delay** (~2 AAC frames / 2048 samples) and pads the last frame:

| Component | AAC frames | Samples |
|---|---|---|
| Priming delay | ~3 | 3,072 |
| Content (ceil(16000/1024)) | 16 | 16,384 |
| **Total** | **19** | **19,456** |

19,456 / 48,000 = **0.40533… s** — matches the actual result exactly.

The audio data itself is the correct size. The problem is that AAC encoding inflates the track duration, and the muxer writes it as-is into the container.

## Possible solutions

1. **Don't add an audio track when there are no audio assets** — Track whether *any* non-silent audio was collected during the render. Only add the audio track to the muxer output if actual audio was present. The change in `3ce1f16a07` was needed for compositions that intermittently have audio assets; it should still emit silence to fill gaps, but shouldn't create an audio track when the composition has zero audio for the entire render.

2. **Trim audio track via MP4 edit list** — Use the `elst` (edit list) box in the MP4 container to indicate the intended media duration, which offsets the priming delay. This is the standard approach in MP4 authoring tools. Would need mediabunny support.

3. **Pre-trim the audio** — Stop feeding audio samples slightly before the last video frame so the encoded audio track aligns with the video duration. This is fragile and codec-dependent.

Option 1 seems most pragmatic.

dexhunter:
AAC encoder priming makes an all-silence audio track longer than the video track, inflating the MP4 container duration beyond durationInFrames / fps. Defer sending audio samples to the encoder until the first frame that actually contains inline audio, backfilling the preceding silence so audio stays in sync. Compositions that never emit audio no longer get an audio track.

Fixes #7099

## Problem

`renderMediaOnWeb()` registers an audio track whenever rendering is not muted and an audio codec is available, and `onlyInlineAudio()` produces an all-zero sample for every frame even when the composition contains no audio. AAC encoder priming then makes the silent audio track longer than the video track, so the MP4 container duration (max of track durations) exceeds `durationInFrames / fps`.

Repro on `main` (10 frames @ 30fps, no audio): container duration `0.3627s` with a silent AAC track, vs `0.3333s` expected.

## Fix

Defer sending audio samples to the encoder until the first frame that actually contains an `inline-audio` asset:

- Frames before the first audio asset only increment a `silentFramesBeforeFirstAudio` counter.
- When the first real audio appears, the preceding silence is backfilled (one sample per skipped frame, with the correct timestamps) before the current sample is added, so audio stays in sync.
- If the composition never emits audio, no samples reach the encoder and the output has no audio track; the container duration then matches the video track.

## Tradeoffs

- The audio track is still registered upfront (mediabunny requires `addAudioTrack()` before `output.start()`); it is finalized empty when no samples are added, which mediabunny drops from the output.
- Compositions where audio starts mid-way pay a one-time backfill of the leading silence at the frame where audio begins (same samples as before, just deferred).

## Tests

- New test: a silent composition produces no audio track and the container duration does not exceed `durationInFrames / fps` (fails on `main`, passes with this change).
- New test: audio starting after a silent lead-in (`<Sequence from={5}>`) still starts at timestamp 0 and covers the full composition duration.
- Verified with ffprobe/ffmpeg out-of-band: silent renders (full and frame-range) have zero excess duration; renders with audio keep an audible, full-length AAC track; mid-start audio has silent lead-in and correctly timed content.

Co-Authored-By: Aiden <aiden@weco.ai>

### DIFF
--- a/packages/web-renderer/src/render-media-on-web.tsx
+++ b/packages/web-renderer/src/render-media-on-web.tsx
@@ -488,6 +488,13 @@ const internalRenderMediaOnWeb = async <
 			};
 		};
 
+		// If a composition never emits audio, no samples are sent to the AAC
+		// encoder - encoder priming would make the audio track longer than the
+		// video track and inflate the container duration (#7099). Silence is
+		// backfilled once the first real audio sample appears.
+		let hasEncounteredAudio = false;
+		let silentFramesBeforeFirstAudio = 0;
+
 		for (let frame = realFrameRange[0]; frame <= realFrameRange[1]; frame++) {
 			if (signal?.aborted) {
 				throw new Error('renderMediaOnWeb() was cancelled');
@@ -588,9 +595,24 @@ const internalRenderMediaOnWeb = async <
 
 			await waitForPageResponsiveness();
 
-			const audio = muted
-				? null
-				: onlyInlineAudio({assets, fps: resolved.fps, timestamp, sampleRate});
+			let audio: AudioData | null = null;
+			if (audioSampleSource) {
+				if (assets.some((a) => a.type === 'inline-audio')) {
+					hasEncounteredAudio = true;
+				}
+
+				if (hasEncounteredAudio) {
+					audio = onlyInlineAudio({
+						assets,
+						fps: resolved.fps,
+						timestamp,
+						sampleRate,
+					});
+				} else {
+					silentFramesBeforeFirstAudio++;
+				}
+			}
+
 			internalState.addAudioMixingTime(performance.now() - audioCombineStart);
 
 			await waitForPageResponsiveness();
@@ -607,6 +629,20 @@ const internalRenderMediaOnWeb = async <
 			}
 
 			if (audio && audioSampleSource) {
+				for (let i = 0; i < silentFramesBeforeFirstAudio; i++) {
+					const silence = onlyInlineAudio({
+						assets: [],
+						fps: resolved.fps,
+						timestamp: Math.round((i / resolved.fps) * 1_000_000),
+						sampleRate,
+					});
+					encodingPromises.push(
+						addAudioSample(silence, audioSampleSource.audioSampleSource),
+					);
+				}
+
+				silentFramesBeforeFirstAudio = 0;
+
 				encodingPromises.push(
 					addAudioSample(audio, audioSampleSource.audioSampleSource),
 				);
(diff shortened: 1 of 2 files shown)


JonnyBurger:
I don't want to do that because it creates weird media file.
If audio only starts after 1 minute in the video, the audio track now also starts later which leads to disjointed tracks.

Media Players might do surprising stuff with it and if you are trying to figure out the sample rate of the audio, libraries will trip up.

My suggestion: Just determine if there are no audio source and pass `muted: true`!
Thanks nonetheless for the PR!

PR #9662, opened 2026-07-25
`@remotion/studio`: Speed up timeline nonce-history sort

dexhunter:
## What

Speeds up `sortItemsByNonceHistory` in the Studio timeline (the sort that orders Sequences, Compositions and Folders by nonce history, added in #6797). The result and behavior are unchanged; only the internal algorithm is faster.

## Why

The current implementation builds a pairwise ordering matrix and then computes a transitive closure with a triple-nested loop over a boxed `(number | null)[][]` matrix — `O(n³)` with per-cell object access. For a large project this runs on every timeline re-sort and becomes visibly slow.

Directly measured, median wall-clock of one sort of mixed multi-epoch inputs (the case that actually exercises the closure):

| items | before | after |
| ---: | ---: | ---: |
| 200 | ~24 ms | ~1.5 ms |
| 500 | ~291 ms | ~10 ms |
| 1000 | ~2181 ms | ~58 ms |

Where it matters: for a typical timeline of a handful to a few dozen items this is sub-millisecond either way and users won't notice. The win shows up for large projects with hundreds of Sequences, Compositions and Folders, where the old `O(n³)` closure caused noticeable Studio lag whenever the list was re-sorted.

## How

Two changes, both behavior-preserving:

- The transitive closure now runs over `Uint32Array` bitsets instead of a boxed matrix, so its inner sweep combines 32 relationships per word (`before[k] & ~known[i]`). A newly derived edge is locked immediately, so the first path decides an otherwise-undetermined pair — this keeps the exact output on the inputs where the newest-shared-epoch comparison is intentionally non-transitive.
- The pairwise comparison no longer rebuilds a `Map` and a `Set` for every pair. Each history is collapsed once into epoch/nonce arrays sorted by descending epoch, and pairs are compared with a linear merge-scan. Once the closure is cheap, this per-pair setup is the larger remaining cost.

The public `compareNonceHistories` export is unchanged.

## Testing

- `bun test packages/studio/src/test/sort-by-nonce-history.test.ts` — 8/8 pass.
- Additionally validated against the previous implementation as an oracle over randomized inputs across the single-epoch, multi-epoch, duplicate-epoch, unsorted-history, tie-dense and duplicate-object cases, plus the handcrafted non-transitive and closure-chain fixtures; the emitted order matches on every case.
- Typecheck and `oxfmt` formatting clean on the touched file.

Supplementary autoresearch — a public record of the exploration behind this change (baseline and the variants measured, including one that was discarded because it changed behavior on cyclic inputs): https://dashboard.weco.ai/share/sSea_OsAHQq29q4g8HtD-PC96KLHxoqu

### DIFF
--- a/packages/studio/src/helpers/sort-by-nonce-history.ts
+++ b/packages/studio/src/helpers/sort-by-nonce-history.ts
@@ -38,6 +38,35 @@ export const compareNonceHistories = (
 	return aLatestEpoch - bLatestEpoch;
 };
 
+// Compare two histories, each pre-collapsed to parallel arrays of epochs and
+// nonces sorted by descending epoch, without rebuilding a Map per comparison.
+// Equivalent to getSharedEpochOrder: both find the newest shared epoch and
+// compare its nonces.
+const sharedEpochOrderFromSorted = (
+	aEpochs: number[],
+	aNonces: number[],
+	bEpochs: number[],
+	bNonces: number[],
+): number | null => {
+	let i = 0;
+	let j = 0;
+	while (i < aEpochs.length && j < bEpochs.length) {
+		const ae = aEpochs[i];
+		const be = bEpochs[j];
+		if (ae === be) {
+			return aNonces[i] - bNonces[j];
+		}
+
+		if (ae > be) {
+			i++;
+		} else {
+			j++;
+		}
+	}
+
+	return null;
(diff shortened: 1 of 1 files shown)


JonnyBurger:
Thanks!

PR #9675, opened 2026-07-26
Skills: Push forgotten maps changes

JonnyBurger:


JonnyBurger:
Consolidated into #9678 with the original commits preserved.

PR #9752, opened 2026-07-27
docs: Document renderMedia licenseKey requirement for v5

### LINKED ISSUE #9539: Docs: Document the v5 `renderMedia()` license key requirement
In Remotion 4.x, the `licenseKey` option has been available since 4.0.409 and is optional.

On the current `5-0-license` branch, the v5 type for `renderMedia()` requires callers to provide `licenseKey: string | null`. This breaking API change is not mentioned in the [v5.0 migration guide](https://www.remotion.dev/docs/5-0-migration), and the [`renderMedia()` reference](https://www.remotion.dev/docs/renderer/render-media#licensekey) still labels the option as `licenseKey?`.

## Tasks

- [ ] Confirm whether requiring the `licenseKey` property in v5 is intentional.
- [ ] If intentional, document the breaking change and required action in the v5.0 migration guide.
- [ ] Update the `renderMedia()` API reference so its optionality matches v5.
- [ ] Explain when callers should pass a license key and when `null` is appropriate.
- [ ] If the required property is not intentional, make it optional again and remove the v5-only call-site changes.

Related to #3750.

santhiprakash:
Closes #9539

## Problem

In Remotion 4.x, `licenseKey` on [`renderMedia()`](/docs/renderer/render-media) was optional. Remotion 5.0 makes `licenseKey: string | null` a required property, but the v5 migration guide had no entry for this breaking change and the API reference still labeled the option as `licenseKey?`.

## Triage / Root cause

JonnyBurger filed #9539 to track the documentation follow-up after the v5 type change landed behind `ENABLE_V5_BREAKING_CHANGES`. The migration guide and `render-media.mdx` were never updated to match.

## Fix

- Add a `renderMedia()` / `licenseKey` section to `packages/docs/docs/5-0-migration.mdx` with required action and guidance on when to pass a license key vs `null`.
- Update `packages/docs/docs/renderer/render-media.mdx` to mark `licenseKey` as optional in v4.x and required from v5.0 (same pattern as `inputProps` on `select-composition.mdx`).

## Verification

- Pre-commit docs formatter passed on commit.
- `python3 ../scripts/preflight_ship.py` against `remotion-dev/remotion` — clear before edit (no duplicate PRs; upstream still missing the migration section).
- `bun run stylecheck` at repo root fails on unrelated upstream `@remotion/starburst#make` / `@remotion/light-leaks#make` (`TS2307` missing `@remotion/effects/*` modules). No files outside `packages/docs/` were changed.

## Notes / Risks

Docs-only change scoped to `renderMedia()` as described in #9539. `renderStill()` still documents optional `licenseKey?` because the v5 required-property change applies only to `renderMedia()` in the current types.

### DIFF
--- a/packages/docs/docs/5-0-migration.mdx
+++ b/packages/docs/docs/5-0-migration.mdx
@@ -180,6 +180,16 @@ This prevents the Player from continuing while media or images are still loading
 
 **Required action**: If you want to keep the old behavior, set `pauseWhenBuffering={false}` on video and audio components and `pauseWhenLoading={false}` on `<Img>` explicitly.
 
+## `renderMedia()` now requires `licenseKey`
+
+In Remotion 4.x, [`licenseKey`](/docs/renderer/render-media#licensekey) was optional on [`renderMedia()`](/docs/renderer/render-media).
+In Remotion 5.0, you must pass `licenseKey: string | null` explicitly.
+
+Pass your license key from [remotion.pro](https://www.remotion.pro) to report usage telemetry.
+Pass `null` to skip sending a usage event.
+
+**Required action**: Add `licenseKey` to every `renderMedia()` call. See [Licensing](/docs/licensing) for when telemetry is required and how to obtain a license key.
+
 ## License changes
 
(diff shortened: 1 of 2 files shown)


JonnyBurger:
Not sure about this, we have not yet set this in stone!
We might change the API, but thank you

PR #9799, opened 2026-07-28
fix(@remotion/web-renderer): treat compositions without audio as muted

### LINKED ISSUE #7099: `@remotion/web-renderer`: AAC encoder priming inflates container duration when emitting silent audio

dexhunter:
Follow-up to #9622 implementing the suggested approach from https://github.com/remotion-dev/remotion/pull/9622#issuecomment-5091262300: determine that a composition has no audio sources and behave as if `muted: true` was passed.

Fixes #7099

## Summary

`renderMediaOnWeb()` registers an audio track whenever rendering is not muted and an audio codec is available, and every frame feeds an all-zero sample to the AAC encoder even when the composition contains no audio. Encoder priming then makes the silent audio track longer than the video track, so the MP4 container duration exceeds `durationInFrames / fps` (repro on `main`, 10 frames @ 30fps: `0.3627s` vs `0.3333s` expected).

Unlike #9622, this does not change how audio samples are timed or emitted — compositions that contain audio anywhere in the range render exactly as before (silence-padded from frame 0, no disjointed tracks).

## Implementation

- Before creating the muxer tracks, scan the frame range (`timeUpdater.update(frame)` → `waitForRenderReady()` → check assets for `inline-audio`, breaking at the first hit) and reset back to the first frame. If no audio is found, `muted: true` is effectively passed to `createAudioSampleSource()`, so no audio track is registered at all.
- The scan is skipped for `muted` renders, containers without an audio codec, and audio-only containers (which keep their track so the output is not empty).
- Added `peekAssets()` alongside `collectAssets()` on the scaffold's asset ref: `collectAssets()` clears the registered assets, and assets only re-register when the frame value changes, so a consuming pre-scan would have swallowed the first frame's assets from the render loop (caught by the existing `toneFrequency` test).

## Tradeoffs

- Compositions with an audio codec pay an extra React render pass per frame (no screenshots/encoding) up to the first frame containing audio. Compositions with audio from frame 0 only pay a single extra frame render.

## Tests

- New test: a silent composition produces no audio track and the container duration does not exceed `durationInFrames / fps` (fails on `main`, passes with this change).
- New test: audio starting after a silent lead-in (`<Sequence from={5}>`) still starts at timestamp 0 and covers the full composition duration.
- Verified with ffprobe/ffmpeg out-of-band: silent renders (full and frame-range) have zero excess container duration; renders with audio keep an audible, full-length AAC track; mid-start audio keeps a silent lead-in and correctly timed content.

Co-Authored-By: Aiden <aiden@weco.ai>

### DIFF
--- a/packages/core/src/RenderAssetManager.tsx
+++ b/packages/core/src/RenderAssetManager.tsx
@@ -25,6 +25,7 @@ export const RenderAssetManager = createContext<RenderAssetManagerContext>({
 
 export type CollectAssetsRef = {
 	collectAssets: () => TRenderAsset[];
+	peekAssets: () => TRenderAsset[];
 };
 
 export const RenderAssetManagerProvider: React.FC<{
@@ -49,6 +50,9 @@ export const RenderAssetManagerProvider: React.FC<{
 					renderAssetsRef.current = [];
 					setRenderAssets([]);
 					return assets;
+				},
+				peekAssets: () => {
+					return renderAssetsRef.current;
 				},
 			};
 		}, []);
--- a/packages/web-renderer/src/create-scaffold.tsx
+++ b/packages/web-renderer/src/create-scaffold.tsx
@@ -166,6 +166,7 @@ export function createScaffold<Props extends Record<string, unknown>>({
 	timeUpdater: React.RefObject<TimeUpdaterRef | null>;
 	collectAssets: React.RefObject<{
 		collectAssets: () => TRenderAsset[];
+		peekAssets: () => TRenderAsset[];
 	} | null>;
 	errorHolder: ErrorHolder;
 	htmlInCanvasContext: HtmlInCanvasContext | null;
@@ -251,6 +252,7 @@ export function createScaffold<Props extends Record<string, unknown>>({
 
 	const collectAssets = createRef<{
 		collectAssets: () => TRenderAsset[];
+		peekAssets: () => TRenderAsset[];
 	}>();
 
 	flushSync(() => {
--- a/packages/web-renderer/src/render-media-on-web.tsx
+++ b/packages/web-renderer/src/render-media-on-web.tsx
@@ -408,6 +408,36 @@ const internalRenderMediaOnWeb = async <
 			throw new Error('renderMediaOnWeb() was cancelled');
 		}
 
+		// A composition without any audio should not get an audio track:
+		// An all-silence AAC track would be longer than the video track due to
+		// encoder priming, inflating the container duration (#7099).
+		// Scan the frame range for audio and treat the render as muted if there
+		// is none. Audio-only containers keep their track.
+		let compositionHasAudio = false;
+		if (!muted && finalAudioCodec !== null && videoEnabled) {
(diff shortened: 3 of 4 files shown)


JonnyBurger:
Sorry, it is a misunderstanding - you should determine from your own data if there is no audio in there.

This PR literally iterates over every frame and sees if there is audio, it makes the whole thing much slower!
I cannot merge it..

PR #9815, opened 2026-07-28
License FAQ and Terms: More explicit about collaborations

MehmetAdemi:


JonnyBurger (approved):
Looks good.
I wish the terms were shorter, but if you see no way to condense, 👍 to merging it

PR #10011, opened 2026-07-31
@remotion/studio: Distinguish render modal environment icon

santhiprakash:
## Problem
The Render modal sidebar used the same `DataIcon` for both Input Props and Environment, which made the two tabs harder to distinguish at a glance.

## Triage / Root cause
`packages/studio/src/components/RenderModal/ServerRenderModal.tsx` rendered `DataIcon` for the Environment tab, so the two adjacent categories shared the same visual marker.

## Fix
- Keep `DataIcon` on Input Props.
- Swap the Environment tab to `SnowflakeIcon` so the sidebar has a distinct icon for environment variables.

## Verification
- `bunx tsc -p packages/studio/tsconfig.json --noEmit --pretty false`

## Notes / Risks
- UI-only change in the server render modal sidebar.
- No behavioral changes to render logic.


JonnyBurger:
Fix is already merged! Thanks

PR #10060, opened 2026-08-02
Elements: Split timed captions element

### LINKED ISSUE #9826: Remotion Elements: Split Timed Captions into three Elements
The timed captions Element in `packages/docs/elements/text/timed-captions/timed-captions.tsx` currently combines three visually distinct caption styles behind a `mode` prop (`highlight`, `scale`, and `background`). These should be separate Remotion Elements instead, so each style is independently discoverable and installable.

## Tasks

- [ ] Split the `highlight`, `scale`, and `background` variants into three separate Remotion Elements.
- [ ] Give each Element its own public component and Elements library entry rather than exposing the styles through one `mode` selector.
- [ ] Keep shared implementation code where useful to avoid unnecessary duplication.
- [ ] Update the relevant previews, metadata, documentation, exports, and tests for the three Elements.
- [ ] Remove the `style.translate` usage from the current timed captions Element/default example.

## Additional context

The current `style.translate` value exists to make the captions fit within their parent container:

- Sequence dimensions: `900x180`
- Captions dimensions: `681x252`
- Translation: `109.5px -36px`

These hardcoded translation coordinates are too arbitrary. Use CSS layout to center the captions within the parent container instead.

yuanran32:
## Summary
  Fixes  #9826
  This PR splits the existing timed captions Element into three standalone Elements:

  - `text/timed-captions-highlight`
  - `text/timed-captions-scale`
  - `text/timed-captions-background`

  Each caption style now has its own public component, Element metadata entry, documentation page, and installable source file instead of being selected through a shared `mode` prop.

  ## Changes

  - Removed the old combined `timed-captions` Element API.
  - Added separate Elements for the highlight, scale, and background caption styles.
  - Updated the Elements sidebar and Element definitions.
  - Updated tests to cover the new timed captions Element slugs.
  - Removed the hardcoded `style.translate` positioning from the default example.
  - Centered the caption layer using CSS layout instead.

  ## Tradeoffs

  The three Elements intentionally keep their implementation self-contained instead of sharing local helper files. This adds some duplication, but keeps each Element independently installable from the Elements library without requiring extra local source dependencies.

  ## Testing

  - `git diff --cached --check`
  - Static validation script for:
    - Element filesystem slugs matching `element-definitions.ts`
    - timed captions slugs being exactly the three new variants
    - no remaining old `mode` API markers
    - no remaining hardcoded `translate: '109.5px -36px'`
    - declared dependencies matching imports


samohovets:
Thanks!

PR #10072, opened 2026-08-03
`@remotion/renderer`: destroy serve-handler read streams on aborted requests (fd leak)

### LINKED ISSUE #10071: serve-handler never destroys fs read streams on aborted requests — fd leak (~0.8 fds/frame on media-heavy renders) ending in EBADF render crash
## Bug

`packages/renderer/src/serve-handler/index.ts` creates an `fs.createReadStream` and pipes it into the HTTP response with no close/abort handler. The leak site is unchanged from **4.0.466 through 4.0.504** (verified against the published `dist/serve-handler/index.js`, currently lines ~193/206):

```js
stream = createReadStream(absolutePath, streamOpts ?? {});
// ...
response.writeHead(response.statusCode || 200, headers);
stream.pipe(response);
```

During a render, Chrome's media demuxer constantly issues and aborts Range requests while seeking `<Video>`/`<Audio>` (`<OffthreadVideo>` is unaffected; plain `<Video>`/`<Audio>` tags are the trigger). Every aborted request leaves the piped `ReadStream` stranded on backpressure: the response socket dies, the stream never gets `destroy()`ed, and its fd stays open for the life of the render process.

## Impact (measured)

On a 16,096-frame 4K composition with 82 video clips (remotion 4.0.466, macOS, M4 Max, concurrency 8), the renderer node process accumulated open fds monotonically at ~0.8/frame — 543 fds at frame 290 → 3,159 at frame 3,619, with hundreds of duplicate REG handles held on individual mp4s long after their sequences ended. At ~frame 15,494 (~12–13k stranded streams) the render died with `EBADF` — `print-useful-error-message.js` attaches the #2452 hint on exactly this condition. It killed three consecutive 52-minute renders, each at ~96% complete.

Controlled repro, 2,000 frames of the same composition, `--concurrency=4 --scale=0.5`, sampling `lsof -p <renderer pid> | wc -l` every 10s:

|  | start | f~570 | f~1,230 | f2,000 |
|---|---|---|---|---|
| unpatched 4.0.466 | 62 | 483 | 819 | **1,257** |
| with the one-line fix below | 58 | 52 | 52 | **55** |

Same wall time in both runs (182 s) — the handler is free.

## Fix (one line, at the single stream-creation site)

```js
response.on('close', () => {
  stream.destroy();
});
stream.pipe(response);
```

`'close'` fires on both normal completion and client abort; `destroy()` on an already-ended stream is a no-op, so only the abort path changes. We have been running this in production as a `patch-package` patch since June and the fd curve went from +0.6–0.8/frame monotonic to flat (±3) over the full render, with byte-identical output.

Happy to open a PR with this if it's useful.

## Environment

remotion 4.0.466 (leak site byte-identical through 4.0.504), node 25.x, macOS 26 (Darwin 25.5.0), Apple Silicon.

Related: #2452 — the historical `EBADF` report this failure mode reproduces.

trevor-sykes:
Fixes #10071.

`serveHandler` pipes an `fs.createReadStream` into the response with no close/abort handler. Chrome's media demuxer aborts Range requests constantly while seeking `<Video>`/`<Audio>` (`<OffthreadVideo>` is unaffected), and every aborted request leaves the piped stream stranded on backpressure — never destroyed, its file descriptor held for the life of the render process.

`'close'` fires on normal completion as well as abort, and `destroy()` on an already-ended stream is a no-op, so only the abort path changes behaviour.

## Measured

2,000 frames of a 4K composition with 82 video clips, `--concurrency=4 --scale=0.5`, sampling `lsof -p <renderer pid> | wc -l` every 10s:

|  | start | f~570 | f~1,230 | f2,000 |
|---|---|---|---|---|
| unpatched | 62 | 483 | 819 | **1,257** |
| patched | 58 | 52 | 52 | **55** |

Same wall time in both runs (182 s). On a 16,096-frame render the unpatched curve climbed at ~0.8 fds/frame and died with `EBADF` at ~frame 15,494, three renders in a row at ~96%.

## Notes

- Equivalent to the `patch-package` patch we have run in production since June on 4.0.466; output byte-identical, no measurable cost.
- `stream` is declared as `let stream = null`, so the local `const readStream` keeps the callback's reference non-null for the type checker rather than widening the diff.
- I have not added a test — reproducing the leak needs a real render loop with an aborting client, which didn't look like it belonged in the existing suite. Happy to add one if you'd like it, in whatever shape suits you.

### DIFF
--- a/packages/renderer/src/serve-handler/index.ts
+++ b/packages/renderer/src/serve-handler/index.ts
@@ -279,6 +279,11 @@ export const serveHandler = async (
 		readStream.destroy();
 	});
 
+	if (response.destroyed) {
+		readStream.destroy();
+		return;
+	}
+
 	response.writeHead(response.statusCode || 200, headers);
 	readStream.pipe(response);
 };
--- a/packages/renderer/src/serve-handler/index.ts
+++ b/packages/renderer/src/serve-handler/index.ts
@@ -268,6 +268,17 @@ export const serveHandler = async (
 
 	headers['Cache-Control'] = 'no-cache, no-store, must-revalidate';
 
+	const readStream = stream;
+
+	// Chrome's media demuxer aborts Range requests constantly while seeking
+	// <Video>/<Audio>. Without this, an aborted request leaves the piped stream
+	// stranded on backpressure and its file descriptor open for the life of the
+	// render process. 'close' fires on completion as well as abort, and
+	// destroy() on an already-ended stream is a no-op.
+	response.on('close', () => {
+		readStream.destroy();
+	});
+
 	response.writeHead(response.statusCode || 200, headers);
-	stream.pipe(response);
+	readStream.pipe(response);
 };
--- a/packages/renderer/src/test/serve-handler.test.ts
+++ b/packages/renderer/src/test/serve-handler.test.ts
@@ -1,43 +1,32 @@
 import {expect, test} from 'bun:test';
 import {once} from 'node:events';
-import {promises} from 'node:fs';
 import type {IncomingMessage, ServerResponse} from 'node:http';
-import os from 'node:os';
 import path from 'node:path';
 import {PassThrough} from 'node:stream';
 import {serveHandler} from '../serve-handler';
 
 test('does not serve a file after the response has closed', async () => {
-	const temporaryDirectory = await promises.mkdtemp(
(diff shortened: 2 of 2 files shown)


JonnyBurger:
Thank you for the detailed report, measurements, and fix!

PR #10154, opened 2026-08-04
Speed up wide HTML-in-canvas renders by settling siblings in parallel

Aryagm:
## Summary

`drawWithHtmlInCanvas()` currently schedules `2 * canvasCount + 1` settling cycles before capturing a nested HTML-in-canvas tree. That is conservative for a chain, but it serializes independent siblings even though Chromium dispatches nested canvas paint events deepest-first within the same rendering update.

This changes the bound to `2 * maxNestingDepth + 1`:

- sibling canvases settle during the same cycles;
- a chain keeps exactly the existing number of cycles;
- mixed trees use their deepest dependency path;
- depth is measured after the initial root paint, preserving the existing observation point for mounts and remounts.

The implementation remains conservative: it does not reduce the existing two cycles per nesting level.

Chromium's nested HTML-in-canvas implementation sorts paint events by descending DOM depth: https://chromium.googlesource.com/chromium/src/+/7cfce65f70cf60df4bb6847ac0c8b57b165a9212

## Performance

Measured with the real `renderStillOnWeb()` path in Chrome for Testing 152.0.7977.13 with `CanvasDrawElement` enabled. Each case is the median of three renders; every baseline/candidate PNG SHA-256 matched exactly.

| Scene | Before | After | Speedup |
|---|---:|---:|---:|
| 1 sibling, depth 1 (control) | 81.4 ms | 81.0 ms | 1.01x |
| 10 siblings, depth 1 | 381.4 ms | 81.8 ms | 4.66x |
| 50 siblings, depth 1 | 1710.8 ms | 81.9 ms | **20.89x** |
| 1 sibling, depth 3 (control) | 148.7 ms | 147.8 ms | 1.01x |
| 1 sibling, depth 5 (control) | 215.2 ms | 215.2 ms | 1.00x |
| 10 siblings, depth 3 | 1048.1 ms | 147.4 ms | **7.11x** |

For 50 siblings, root `requestPaint()` calls drop from 102 to 4. Depth-only controls are unchanged.

An 8-frame `renderMediaOnWeb()` benchmark with 32 sibling layers improved from a median 8.912 s to 0.635 s: **14.03x faster**. A separate checksum-enabled run copied every pre-encode `VideoFrame` to RGBA and measured **15.21x** while matching all 48 baseline/candidate frame hashes exactly. Internal timings localized the gain to frame creation (8730.2 ms → 481.8 ms, **18.12x**).

## Correctness and tests

- Added topology tests for wide siblings, nested chains, and mixed trees.
- Added a native Chrome 152 integration test with 24 sibling `<HtmlInCanvas>` components. Every async `onPaint` waits for a later animation frame before drawing a unique RGB sentinel; the final PNG is decoded and every tile's RGBA is checked.
- `@remotion/web-renderer` build passes.
- Targeted formatting, ESLint, TypeScript, and `git diff --check` pass.
- System Chrome 150 suite: 10/10 pass; unsupported native tests skip normally.
- Chrome 152 native tests for stills, videos, transient missing records, and async siblings pass.
- The existing forced software-composer golden differs under Chrome 152 in both clean baseline and candidate by the same 8,551 pixels, so it is unrelated to this native-path change.

### DIFF
--- a/packages/web-renderer/src/html-in-canvas.ts
+++ b/packages/web-renderer/src/html-in-canvas.ts
@@ -182,14 +182,45 @@ export const supportsNestedHtmlInCanvas = (): Promise<boolean> => {
 	return nestedHtmlInCanvasSupport;
 };
 
-const countLayoutSubtreeCanvases = (element: HTMLElement): number => {
-	return Array.from(element.querySelectorAll('canvas')).filter(
-		(canvas) => (canvas as HTMLCanvasWithLayoutSubtree).layoutSubtree === true,
-	).length;
+const isLayoutSubtreeCanvas = (
+	element: Element,
+): element is HTMLCanvasWithLayoutSubtree => {
+	return (
+		element.localName === 'canvas' &&
+		(element as HTMLCanvasWithLayoutSubtree).layoutSubtree === true
+	);
+};
+
+export const getMaxLayoutSubtreeCanvasDepth = (
+	element: HTMLElement,
+): number => {
+	let maxDepth = 0;
+	const stack = Array.from(element.children, (child) => ({
+		element: child,
+		layoutSubtreeDepth: 0,
+	}));
+
+	while (stack.length > 0) {
+		const current = stack.pop()!;
+		const layoutSubtreeDepth =
+			current.layoutSubtreeDepth +
+			(isLayoutSubtreeCanvas(current.element) ? 1 : 0);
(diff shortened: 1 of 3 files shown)


JonnyBurger:
@Aryagm Please undraft it if you think this is legit, or close it if you don't wanna pursue it

PR #10194, opened 2026-08-05
`@remotion/studio`: Optimize selected outline rendering

JonnyBurger:
## Summary

- calculate selectable outline targets once per frame and keep measurement/rendering in a memoized renderer boundary
- keep frame-wide targets limited to geometry and ordering data while selected or hovered outline leaves resolve their own controls
- let only active UV leaves subscribe to the timeline and calculate UV handles
- resolve multi-selection drag targets imperatively when pointer or keyboard interactions begin
- isolate fine-grained canvas hover updates, keep one ResizeObserver connected, and calculate snap targets only during drags

## Testing

- `bunx turbo run make --filter='@remotion/studio'`
- `bun run --cwd packages/studio lint`
- `bun run --cwd packages/studio test`


JonnyBurger:
Superseded by #10236, which contains both layers and has been merged.

PR #10199, opened 2026-08-06
@remotion/studio-server: Open source locations in the existing Zed window

### LINKED ISSUE #9744: Studio: Open source files in the existing Zed window
Opening a source location from Remotion Studio invokes Zed with only the `file:line:column` argument. If the user has `cli_default_open_behavior` set to `new_window`, every source-opening action creates another Zed window.

## Expected behavior

Studio source-opening actions should explicitly reuse an existing Zed window instead of depending on the user's default CLI behavior.

## Implementation notes

- Pass Zed's `-e` / `--existing` option when Studio opens a source location.
- Keep the `create-video` behavior separate because opening a newly created project in a new window is intentional there.
- Add coverage for the Zed launch arguments.

## Upstream context

Zed exposes `-e` / `--existing` in its CLI help and current source, but the published CLI reference currently omits it. The documentation mismatch is tracked in zed-industries/zed#61730:

https://github.com/zed-industries/zed/issues/61730

garudaccs:
## Summary

- When Remotion Studio opens a source file in Zed, it now passes the `--existing` flag so the file opens in the current Zed window instead of spawning a new window every time.
- The `create-video` path is intentionally left unchanged, as it is expected to open a new window.

Closes #9744

#### Test plan
- [x] `bunx eslint src/helpers/open-in-editor.ts` passes
- [x] `bun test src/test/open-in-editor.test.ts` passes

### DIFF
--- a/packages/studio-server/src/helpers/open-in-editor.ts
+++ b/packages/studio-server/src/helpers/open-in-editor.ts
@@ -277,7 +277,7 @@ function getArgumentsForLineNumber(
 	const isFolder =
 		fs.existsSync(fileName) && fs.lstatSync(fileName).isDirectory();
 	if (isZedEditor) {
-		return [fileName + ':' + lineNumber + ':' + colNumber];
+		return ['--existing', fileName + ':' + lineNumber + ':' + colNumber];
 	}
 
 	switch (editorBasename) {


garudaccs:
Closed: opened from the wrong account due to an org block on the authorized account. Will retry once the block is resolved.

PR #10208, opened 2026-08-06
`@remotion/media`: add `clearMediaCaches()` to release sinks between renders

alec-watts:
## Problem

`sinkPromises` in `get-sink.ts` is a module-level cache that is only ever written to:

```ts
export const sinkPromises: Record<string, Promise<GetSink>> = {};
```

Each entry owns a mediabunny `Input`, which holds container/demuxer state plus the source's byte cache. Nothing disposes them, so the memory is process-lifetime. In an app that runs several renders in one page, everything every render touched stays resident.

Two things in the package already point at this gap:

- `get-shared-input.ts` reference counts its `Input`s and disposes on last release. Its comment says *"This mirrors the sink cache in `get-sink.ts`"* — but the sink cache has no release path at all.
- `keyframeManager.clearAll()` is implemented and exported, and called from nowhere in the repo.

## Measurement

Rendering a composition with 6 video layers and 6 audio layers off local files (4K HEVC, `renderMediaOnWeb`, 1800 frames), heap **after `renderMediaOnWeb` resolved**:

| | Retained heap |
|---|---|
| after render completes | 812 MB |

That is memory the page cannot reuse for the next render.

## Change

- `getSinks()` returns `dispose()`, so the `Input` it closes over is reachable.
- `clearSinks()` disposes every cached sink and empties `sinkPromises`. It uses `allSettled`, so a sink that never opened (404, unsupported container) neither blocks the others nor surfaces as an unhandled rejection.
- `audioManager` gains `clearAll()`, matching the one `keyframeManager` already has.
- `clearMediaCaches()` ties the three together and is exported from the package entry point, since `@remotion/media` only exposes `index.ts` and nothing else in the repo currently depends on it.

This adds an API rather than changing existing behaviour — no current call path changes. Happy to wire it into the renderer or a lifecycle hook instead if you'd prefer it not be public; I couldn't find an existing "render finished" signal inside the package to hang it on.

## Tests

`src/test/clear-sinks.test.ts` covers disposal of all cached sinks, the rejected-sink case, and the empty-cache no-op.

Typecheck is unchanged against `main` (32 pre-existing errors before and after, none in the touched files) and `eslint` is clean on the changed files.


JonnyBurger:
Thank you for investigating this and putting together the fix, Alec — this was very helpful.

We went with #10240, which scopes the media caches and extraction queues to each client-side render and disposes them with the render lifecycle. I’m closing this PR as superseded, but really appreciate the contribution nonetheless.

PR #10261, opened 2026-08-07
Docs: Mention @remotion/media tags on the Lambda cost page

### LINKED ISSUE #10250: https://www.remotion.dev/docs/lambda/optimizing-cost should mention new video tags
https://www.remotion.dev/docs/lambda/optimizing-cost

Many customers can benefit from new media tags over OffthreadVideo, especially on Lambda.

suzunn:
Fixes #10250.

The [Optimizing for cost](https://www.remotion.dev/docs/lambda/optimizing-cost) page did not mention the `@remotion/media` tags, even though they are the biggest cost lever for many Lambda users.

Adds a section covering the two ways they lower the bill:

- **Less compute time** — they render fastest, and Lambda is billed per GB-second.
- **Less data transfer** — byte range requests mean a function only downloads the part of the file it renders, instead of pulling the whole file into every function.

Both points link to pages that already document them ([Data transfer cost](https://www.remotion.dev/docs/lambda/data-transfer-cost#option-2-use-remotionmedia-tags), [Comparison of video tags](https://www.remotion.dev/docs/video-tags)) rather than restating the detail here. I also noted that a fallback to `<OffthreadVideo>` forgoes the savings, linking to [Fallback behavior](https://www.remotion.dev/docs/media/fallback), and added the two pages to "See also".

The section is placed first on the page to match "many customers can benefit" from the issue.

Docs-only change (one `.mdx` file). I verified every link target resolves to an existing page and that the `#option-2-use-remotionmedia-tags` anchor matches the heading on the data transfer page, but I did not run a full `docusaurus build` locally.

### DIFF
--- a/packages/docs/docs/lambda/cost.mdx
+++ b/packages/docs/docs/lambda/cost.mdx
@@ -8,6 +8,15 @@ crumb: 'Lambda'
 
 On this page, a few strategies for optimizing the cost of using Remotion Lambda are presented.
 
+## Use the `@remotion/media` tags
+
+Use [`<Video />`](/docs/media/video) and [`<Audio />`](/docs/media/audio) from [`@remotion/media`](/docs/media) instead of [`<OffthreadVideo />`](/docs/offthreadvideo) or [`<Html5Video />`](/docs/html5-video). On Lambda, this reduces the bill in two ways:
+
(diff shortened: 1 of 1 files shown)


JonnyBurger:
Already covered by another PR, thank you

Issue #10161, opened 2026-08-05
OffthreadVideo: compositor returns a wrong (far-ahead) frame after frame-cache eviction on deep B-frame-pyramid sources, then repeats it for the life of the process


Haroldas-Matulevicius:
**Remotion version:** 4.0.455 (`remotion`, `@remotion/renderer`, `@remotion/lambda` all pinned). Reproduced with the darwin-arm64 compositor binary; first observed on Lambda (linux-arm64, same version).

### Summary

When `maximum_frame_cache_size_in_bytes` is small enough that the frame cache must evict, `ExtractFrame` returns a frame roughly **one second ahead** of the requested timestamp on a source encoded as a deep B-frame pyramid — and then **stays wedged**, returning that same frame for every subsequent request in that compositor process. With the default (large) cache the same process, source, and request sequence are 100% correct, so this looks like cache-eviction state corruption in frame selection, not a decode bug.

On Lambda this surfaces as rendered MP4s that freeze for ~1 s and then visibly jump **backward**: `framesPerLambda` gives each chunk a fresh compositor process, so each chunk renders its first frames correctly, wedges, holds one far-ahead frame to the chunk end, and the next chunk's correct restart lands *behind* the held frame.

### Source characteristics (the content trigger)

- HEVC (`hvc1`), 1080×1080, 60 fps CFR (single 16.67 ms packet interval across all 1420 packets), 23.7 s.
- Picture types (histogram approximate): **4 I-frames, 0 P-frames, all remaining ~1416 frames B-frames** — GOPs of ~420/407/373 frames (~7 s random-access granularity). Typical of social-platform re-encodes. (A separate frame-type tally recorded 1417 B, one more than the 1420-packet total supports; the one-frame discrepancy is immaterial to the trigger — the source is B-frames wall-to-wall either way.)

### Reproduction (no browser involved)

Drive the long-running compositor directly (`startLongRunningCompositor` → `executeCommand('ExtractFrame', …)`) at the exact timestamps `<OffthreadVideo>` requests during a 30 fps render (`t = k/30`, k = 0..689), sequentially, request pool of 1 (zero concurrency):

| `maximum_frame_cache_size_in_bytes` | wrong frames / 690 |
|---|---|
| 100–1500 MB | 122 |
| 3000 MB | 88 |
| 6000 MB | 2 |
| 12000 MB / 24000 MB / default | 0 |

[...]

PR #10265, opened 2026-08-08
fix: Prevent ExtractFrame panic with zero extra threads

mvanhorn:
## Summary

Starting the long running compositor with `extraThreads: 0` and issuing an `ExtractFrame` command panics the Rust compositor: the thread map is created with zero capacity, so the least loaded thread selection unwraps an empty iterator. This change clamps the worker thread count to at least one in `LongRunningProcess::new`, which guarantees a worker slot for frame extraction while leaving non-frame commands unchanged.

## Why

`LongRunningProcess::new` uses the `threads` value both as the `ThreadMap` capacity and as the divisor for per-thread frame cache limits. With zero extra threads there is no existing or allowable worker when the first `ExtractFrame` request arrives, so thread selection panics on an empty map, and the same value invites a divide by zero in the cache limit math. Clamping at this boundary keeps `extraThreads: 0` usable for non-frame commands such as metadata extraction and makes frame extraction safe under the same setting.

This covers the independent zero thread panic described in the issue. The primary frame cache report there is awaiting reporter verification of the maintainer-linked fix released in 4.0.506, so this PR intentionally does not touch frame cache selection behavior.

Refs #10161

## Testing

Added a test in `packages/renderer/src/test/extract-frame-rust.test.ts` that starts the compositor with `extraThreads: 0`, extracts a frame from the Big Buck Bunny example video, asserts the returned BMP payload has the expected 1280x720 dimensions, and waits for the compositor to shut down cleanly, which fails on the current panic. The full monorepo suite was not run in this environment.


JonnyBurger:
Thanks!

PR #10271, opened 2026-08-09
Docs: How to get the FPS of a video

### LINKED ISSUE #8815: Matching video FPS is a bad idea

JonnyBurger:
## Summary

- add a Mediabunny guide for detecting constant and variable video frame rates from packet timestamps
- show how to align a Remotion composition's FPS with the source video
- update the metadata example to reuse the frame-rate helper and add navigation and preview metadata

Fixes #8815

## Validation

- `bun run build`
- `bun run stylecheck`

## Preview

- [Getting the frame rate of a video](https://remotion-git-codex-docs-video-frame-rate-remotion.vercel.app/docs/mediabunny/frame-rate)
- [Getting metadata from a video](https://remotion-git-codex-docs-video-frame-rate-remotion.vercel.app/docs/mediabunny/metadata)


packages/docs/docs/mediabunny/frame-rate.mdx
```diff
@@ -0,0 +1,196 @@
...
+  }
+
+  timestamps.sort((a, b) => a - b);
+
+  const intervals = timestamps.slice(1).map((timestamp, index) => {
+    return timestamp - timestamps[index];
+  });
+  const averageInterval =
+    intervals.reduce((sum, interval) => sum + interval, 0) / intervals.length;
+  const toleranceInSeconds = 1 / (await videoTrack.getTimeResolution());
```
JonnyBurger: **[P1] Timestamp resolution is not always timestamp precision.** The tolerance can be much narrower than rounding inherited by the timestamps. I reproduced this with `packages/example/public/av1-bbb.mp4`: timestamps alternate between 33 ms and 34 ms, but `getTimeResolution()` returns 16000, producing a tolerance of only 0.0625 ms. The helper therefore reports variable FPS with a maximum of 30.303 instead of reaching the 30 FPS snap. Derive tolerance from observed timestamp quantization or allow the expected millisecond rounding before classifying the stream.

packages/docs/docs/mediabunny/frame-rate.mdx
```diff
@@ -0,0 +1,196 @@
...
+  const isConstant = intervals.every(
+    (interval) => Math.abs(interval - averageInterval) <= toleranceInSeconds,
+  );
+
+  if (isConstant) {
```
JonnyBurger, line 99: **[P1] A prefix cannot establish constant frame rate.** Only the first 121 packets are inspected, so a video that starts at a constant rate and becomes variable later is returned as `type: 'constant'`. Either sample at several positions throughout the track or clearly document that this is a prefix-based heuristic that can produce false `constant` results.

packages/docs/docs/mediabunny/frame-rate.mdx
```diff
@@ -0,0 +1,196 @@
...
+): Promise<VideoFrameRate | null> => {
+  const videoTrack = await input.getPrimaryVideoTrack();
+  if (!videoTrack) {
+    return null;
+  }
+
+  const timestamps: number[] = [];
+  const sink = new EncodedPacketSink(videoTrack);
+
+  for await (const packet of sink.packets(undefined, undefined, {
```
JonnyBurger, line 66: **[P2] Avoid truncating a decode-order iterator.** Mediabunny documents that `EncodedPacketSink.packets()` yields packets in decode order. With B-frames, stopping at exactly packet 121 can include a future reference packet while omitting earlier presentation-order packets that appear later in decode order. Sorting cannot restore timestamps that were never collected. Continue until the current reorder group has been flushed, then sort and select the desired presentation-order prefix.

packages/docs/docs/mediabunny/frame-rate.mdx
```diff
@@ -0,0 +1,196 @@
...
+  });
+
+  const frameRate = await getVideoFrameRate(input);
+  if (frameRate === null) {
+    return {
+      fps: 30, // Fallback if none or not enough video frames are available
+    };
+  }
+
+  return {
```
JonnyBurger, line 200: **[P2] Preserve duration when overriding FPS.** Returning a new `fps` while leaving `durationInFrames` unchanged changes the composition’s duration in seconds. For example, 300 frames changes from 10 seconds at 30 FPS to 5 seconds at 60 FPS. Either calculate and return `durationInFrames` using the source duration and selected FPS, or explicitly warn that it must be adjusted separately.

JonnyBurger (commented):
Peer review: I found four correctness issues in the frame-rate article. The first two affect the helper’s central accuracy; the remaining two cover packet ordering and composition-duration behavior.

PR #10266, opened 2026-08-08
Docs: Clarify that Automators applies only to Company License users

MehmetAdemi:
## Summary

Clarify that Remotion for Creators and Remotion for Automators are Company License options that apply only after determining that a user does not qualify for the Free License.

Changes to `packages/docs/docs/license/faq.mdx`:

- state that Free License users may create automated pipelines and render programmatically
- move the unqualified automation statement into “When do I need to purchase Renders?” and add the Free License condition

Closes remotion-dev/business#1426.

## Split PRs

- #10276 aligns the Free License eligibility wording with the Terms
- #10277 reorders the open-source FAQ item

## Test plan

- confirm the Free License clarification renders in the eligibility answer
- confirm the automation statement appears only under “When do I need to purchase Renders?”
- review the wording against remotion-dev/business#1426

PR #10277, opened 2026-08-10
Docs: Reorder the open-source license FAQ

JonnyBurger:
## Summary

Move the “Is Remotion open-source?” FAQ item after the free-versus-paid functionality answer so the licensing overview reads in a more logical order.

This editorial change was split out of #10266 so it can be reviewed independently.

## Test plan

- confirm the FAQ order renders correctly
- confirm the existing heading anchors remain unchanged


JonnyBurger:
Seems like a no-op

PR #10310, opened 2026-08-10
License FAQ - Free License eligibility

MehmetAdemi:


packages/docs/docs/license/faq.mdx
```diff
@@ -14,7 +14,7 @@ Remotion is a React-based technology for programmatic video creation. It lets yo
 
 You are eligible to use Remotion for free if you are:
 
-- an individual
+- an individual, whether for personal or commercial use
```
JonnyBurger, line 17: Why do we need to change this?
What does this even clarify?

Let's keep changes minimal, because people can see the diff and they might think some policy of ours has changed.

JonnyBurger (dismissed):

packages/docs/docs/license/faq.mdx
JonnyBurger: Okay, nevermind I see this not a license change. I am actually okay with it 👍

MehmetAdemi: FYI; someone reached out and was confused whay individual didn't had an explicit mentioning of using Remotion commercially, that's why.

PR #10311, opened 2026-08-10
`remotion`: Re-check audio resume success before reporting failure

dukeeagle:
`waitUntilActuallyResumed()` polls for output-clock progress with `requestAnimationFrame`, but declares failure with a wall-clock `setTimeout`. When the main thread is busy for around a second (heavy rendering, or a bundler rebuild during development), the timeout can fire before the rAF callback ever gets to run. A resume that actually completed is then reported as `'failed'`, and `use-playback.ts` responds by muting the Player - even though playback was started by a user gesture and audio is audibly running. Nothing unmutes it afterwards, so one busy second silences the Player for its whole lifetime.

This change re-checks the output timestamp once inside the timeout callback before finishing with `'failed'`. If the context did audibly start, the wait resolves `'resumed'` as it would have if the rAF loop had been scheduled in time. Genuine autoplay blocks still time out and report failure exactly as before.

Repro: start playback from a click, then keep the main thread busy past the 1000ms window (e.g. a large synchronous task right after `play()`). Without this change the Player logs the autoplay warning and mutes despite the gesture; with it, playback continues unmuted.

### DIFF
--- a/packages/core/src/audio/wait-until-actually-resumed.ts
+++ b/packages/core/src/audio/wait-until-actually-resumed.ts
@@ -39,12 +39,14 @@ export const waitUntilActuallyResumed = (
 
 		onAbort = () => finish('cancelled');
 
-		const hasAudiblyStarted = () => {
+		const hasAudiblyStarted = (
+			startPerformanceTime: number | undefined,
+		): startPerformanceTime is number => {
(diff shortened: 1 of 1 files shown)


JonnyBurger:
Thank you

PR #10320, opened 2026-08-11
`remotion-monorepo`: Fix pre-commit hook on Linux

anthod0:
The pre-commit hook currently uses `bun run --filter` to format changed packages. On Linux, this fails immediately with:

```text
  error: ENOENT
```

This is caused by a known Bun bug when resolving recursive workspace patterns containing mixed-case directory names:

- https://github.com/oven-sh/bun/issues/21670
- https://github.com/oven-sh/bun/pull/29941

This PR avoids workspace resolution in the hook. It tracks the directories of changed packages and runs each formatter directly using bun run --cwd.

Validation

- Reproduced the failure on Linux with a change in packages/docs.
- Ran bun pre-commit.ts after the fix; oxfmt src completed successfully.
- Verified --cwd using the repository's pinned Bun 1.3.3.
- Ran bunx oxfmt pre-commit.ts --check.

Tradeoff

Formatters now run sequentially when multiple packages are changed. In practice, pre-commit changes usually affect only a small number of packages, and this avoids relying on Bun's currently broken workspace filtering on Linux.

### DIFF
--- a/pre-commit.ts
+++ b/pre-commit.ts
@@ -13,7 +13,7 @@ if (changedFiles.length === 0) {
 	process.exit(0);
 }
 
-const formatPackageNames = new Set<string>();
+const formatPackageDirs = new Set<string>();
 const lintPackageNames = new Set<string>();
 
 for (const file of changedFiles) {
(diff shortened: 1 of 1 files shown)


JonnyBurger:
Thanks!

PR #10325, opened 2026-08-11
Add snapcn to resources and a terminal typing Element

SriSomanaath:
Reopens #10323, which closed itself when I force-pushed the branch. Same two changes, same content, clean history.

**Resources**: adds snapcn to the Components list. Remotion components for product demo videos: AI answer streams, terminal sessions, device frames, captions and full scenes. MIT licensed, every component has a preview on the site.

**Element**: adds a Terminal Typing Element in a new Screens category. A terminal window types a command and prints its output line by line, then leaves a blinking prompt. Software demos and tutorials need this shot constantly and the gallery has nothing like it.

Built with the scaffold-element and submit-element skills. Status markers are CSS dots rather than glyphs, so the Element cannot break on a font subset that lacks check marks or arrows, and only opacity and translation animate — nothing scales, which keeps the monospace text from crawling the pixel grid while the window settles.

`bun test src/test/elements.test.ts` 142 passing, `bun run stylecheck` clean, previews rendered with `render-element-previews` and committed under static/elements as the external contribution path describes.


SriSomanaath:
Closing while I settle on the entry wording. Will reopen a fresh PR with just the resources line.

PR #10336, opened 2026-08-12
Fix editor fallback action

### LINKED ISSUE #10076: If no editor is running and no default editor is set, it should still show an icon to open in the editor
The first one in the list, but that is alphabetical.
My suggested order is:

- Zed, because if you have that installed, you probably want that
- VS Code, because that is a super solid choice
- Cursor, hope that we don't land in Agents window

then alphabetical

YUHAO-corn:
## Summary

Fixes #10076.

When Studio cannot detect a running editor and no default editor is configured, the primary "Open in editor" action now uses the best installed editor instead of becoming disabled. The fallback order follows the issue: Zed first, then VS Code, then Cursor, then the remaining installed editors alphabetically.

The existing running-editor behavior stays unchanged. If Studio has a running editor name, the primary action still opens that editor and does not silently switch to a fallback.

## Verification

- bun test src/test/default-editor-info.test.ts src/test/sequence-context-menu-items.test.ts
- OPEN_SOURCE_PW_BROWSERS_PATH=/Users/godcorn/cursor/open-source/.cache/ms-playwright PLAYWRIGHT_BROWSERS_PATH="$OPEN_SOURCE_PW_BROWSERS_PATH" bunx playwright test e2e/studio.test.mts -g "should fall back to the preferred installed editor"
- bunx turbo run make --filter="@remotion/cli" --filter="@remotion/studio"
- bunx turbo run make --filter="@remotion/example"
- bun run formatting
- bun run lint (0 errors; existing warnings only)

### DIFF
--- a/packages/example/e2e/studio.test.mts
+++ b/packages/example/e2e/studio.test.mts
@@ -562,6 +562,92 @@ test.describe('visual mode', () => {
 			),
 		).toEqual([]);
 	});
+
+	test('should fall back to the preferred installed editor if no editor is running', async ({
+		page,
+	}) => {
+		await page.addInitScript(() => {
+			Object.defineProperty(window, 'remotion_editorName', {
+				configurable: true,
+				get: () => null,
+				set: () => undefined,
+			});
+		});
+		await page.route('**/api/default-editor-info', async (route) => {
+			await route.fulfill({
+				json: {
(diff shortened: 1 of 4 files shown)


JonnyBurger:
Thanks!

PR #10416, opened 2026-08-13
`@remotion/bundler`: do not crash when TypeScript exposes no `sys`

Octember:
## Problem

`ESBuildLoader` reads the project's `tsconfig.json` via `typescript.readConfigFile(tsConfigPath, typescript.sys.readFile)`. On workspaces where the installed `typescript` is the native preview (TypeScript 7 / `@typescript/native-preview`), the CommonJS export does not expose `sys`, so bundling crashes with `Cannot read properties of undefined (reading 'readFile')`.

## Fix

Guard on `sys` being present. When it is not, fall back to the exact behavior Remotion already has when TypeScript is not installed at all (esbuild transforms without `tsconfigRaw`).

## Tradeoffs

On TS installs without `sys`, `tsconfigRaw` is not populated — same as the no-TypeScript path today. That seems strictly better than crashing the bundle step.

We have been running this patch in production at Bevyl (via `patchedDependencies`) since 4.0.484.


JonnyBurger:
Thanks!

PR #10424, opened 2026-08-14
`@remotion/media`: Restart audio after iterator destruction

JonnyBurger:
## Problem

`audioIteratorManager.seek()` deduplicates identical seek parameters. After `destroyIterator()` removes the active iterator, the cached parameters remain, so the next seek to the same time is skipped and audio does not restart.

## Solution

Represent the cached seek as nullable state and clear it whenever the iterator is destroyed. This models the cache validity directly instead of assigning a magic `-1` time sentinel.

Starting without a cached seek also removes the initial seek parameters that were previously required only to construct the sentinel state.

## Testing

- Added a browser regression test covering seek → destroy → identical seek.
- Verified that the test fails without invalidation and passes with the fix.
- Ran `bun run build`.
- Ran `bun run stylecheck`.

Supersedes #10417.

### DIFF
--- a/packages/media/src/audio-iterator-manager.ts
+++ b/packages/media/src/audio-iterator-manager.ts
@@ -78,13 +78,6 @@ export const audioIteratorManager = ({
 	getStartTime,
 	initialMuted,
 	drawDebugOverlay,
-	initialPlaybackRate,
-	initialTrimBefore,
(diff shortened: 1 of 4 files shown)


Octember (commented):
ty!

PR #10471, opened 2026-08-14
`@remotion/media`: Add effect output size to `<Video>`

dukeeagle:
`<Video>` currently evaluates effects at the encoded dimensions. Pixel-based effects therefore change scale when the same video is swapped between a proxy and an original, which apps relying on Remotion might do often ;) 

This adds an optional `effectsOutputSize` prop to choose the effect canvas size in preview and rendering. If it is omitted, behavior is unchanged.

Before: `pixelate({blockSize: 128})` on 512 × 512 and 2048 × 2048 VP8 encodes of the same video:

Different pixelate scale at the encoded dimensions

After: both use `effectsOutputSize={{width: 1024, height: 1024}}`:

Matching pixelate scale at an explicit effect output size

The screenshots were rendered at frame 60 with `--gl=angle`. The full `@remotion/media` browser suite passes (123 tests), along with `bun run build` and `bun run stylecheck`.


packages/media/src/video/video-for-rendering.tsx
```diff
@@ -450,6 +469,7 @@ export const VideoForRendering: React.FC<InnerVideoProps> = ({
 		credentials,
 		effectChainState,
 		effects,
+		effectsOutputSize,
```
JonnyBurger: This could lead to a lot of re-renders if user inlines it

```
effectsOutputSize={{width: 1920, height: 1080}}
```

should put .width and .height independently in effect depdendencies

packages/media/src/media-player.ts
```diff
@@ -141,6 +145,7 @@ export class MediaPlayer {
 			width: number,
 			height: number,
 		) => EffectChainState | null;
+		getEffectsOutputSize?: () => EffectsOutputSize | undefined;
```
JonnyBurger: Follow convention and don't make properties optional
Use `/nullable-new-params` skill to do it

JonnyBurger (changes requested):

alec-watts:
I addressed the requested dependency and nullable-parameter changes in [dukeeagle/remotion#1](https://github.com/dukeeagle/remotion/pull/1). It targets this PR's head branch directly, so merging it will update this PR. The media package build passes and the relevant browser tests pass; two unrelated frame-accuracy screenshot comparisons differed on my macOS run.

JonnyBurger:
@alec-watts Please undraft and fix CI if you want me to review

PR #10472, opened 2026-08-14
`@remotion/web-renderer`: Allow still renders during media renders

alec-watts:
## Summary
- use independent serial queues for `renderMediaOnWeb()` and `renderStillOnWeb()`
- allow one still render to run while one media render is in progress
- keep media-media and still-still renders serialized as before

The render queue was introduced alongside the OPFS media output target. Still renders do not use that output target, but currently wait behind long-running media exports because both operations share the same queue.

## Test plan
- [x] Add a browser regression test that pauses a media render and verifies a still render starts before it finishes
- [x] Run `bun run makewebrenderer`
- [x] Run web-renderer lint and formatting checks
- [x] Run the full web-renderer browser suite (687 passed, 27 skipped)


JonnyBurger:
Sounds good 👍

Issue #10235, opened 2026-08-07
Feature Request: Export timeline / edit decision list (EDL / FCPXML / Premiere XML / OTIO) for DaVinci Resolve & Premiere Pro


krishnachitanyapidugu-sys:
# Feature Request 🛍️

Add the ability to export a Remotion composition or timeline as an industry-standard interchange format (EDL, FCPXML, Premiere XML, or OTIO) so it can be imported into DaVinci Resolve or Adobe Premiere Pro as an editable sequence with source media links.

## Use Case

Many users (including me) use Remotion + AI tools (Codex plugin, Claude, etc.) to quickly generate rough cuts and hard cuts.  

After the AI creates the structure, we want to hand the edit over to a human editor in **DaVinci Resolve** or **Adobe Premiere Pro** so they can do the final polish, color grading, sound design, and fine-tuning using the original camera files.

Currently Remotion only outputs a finished rendered video (MP4 / ProRes / etc.). There is no way to export the actual edit decision list (cuts, clip placements, in/out points, and source references). This forces us to either:
- Drop a rendered ProRes as a single flat clip (loses all editability), or
- Manually rebuild the entire cut in the NLE (very time-consuming).

A proper timeline export would enable a clean hybrid workflow:  
**AI does the hard cuts → Human finishes the masterpiece in a professional NLE.**

## Possible Solution

Provide an export function (CLI flag + Node.js API) that generates one or more of the following formats from a Remotion composition:

- EDL (CMX 3600) – simplest and widely supported
- FCPXML
- Premiere XML (xmeml)
- OTIO (OpenTimelineIO)

The export should include:
- Clip placements and durations
- Source media paths / file names (so Resolve/Premiere can relink)
- Basic track structure
- In/out points

This could start as a simple sequential export of the main video track and later expand to support multiple tracks, transitions, etc.

krishnachitanyapidugu-sys:
I'd like to work on this feature.

I can start by implementing a basic EDL (CMX 3600) exporter first, since it's the simplest and most universal format, then expand to FCPXML / Premiere XML / OTIO if the approach is solid.

Happy to discuss the architecture before writing a lot of code.

JonnyBurger:
https://remotion.dev/docs/export-opentimeline

We made a skill!

PR #10474, opened 2026-08-15
Skills: Add `/remotion-opentimeline` skill

krishnachitanyapidugu-sys:
Related to #10235.

That issue asks for a way to hand a Remotion edit over to DaVinci Resolve or Premiere. This PR proposes the smallest useful step in that direction: an agent skill that teaches an agent to produce a valid OpenTimelineIO (`.otio`) file from a composition. No new package, no runtime code, no changes to shipped behaviour.

I went with a skill rather than an exporter because a Remotion composition is React code, so there is no deterministic mapping from a component tree to a timeline. Deciding which elements are clips, where they sit, and what has to be dropped is a judgement call, which is the kind of thing a skill handles better than an API. If you would rather see this as a real exporter in a package, say so and I will close this and work on that instead.

## What is in the PR

- `packages/skills/skills/remotion-export-otio/SKILL.md` — the skill
- `agents/openai.yaml` and `assets/remotion-icon.png`, matching what every other public skill ships
- The registrations a new public skill requires: `remotion-skill-names.ts`, the `npx skills update` list in `remotion-upgrade`, a router entry in `remotion-best-practices`, and the docs page (`packages/skills/README.md` is generated from it)
- The symlinks created by `bun run syncskills`

## What the skill does and does not claim

It maps `<Video>`, `<OffthreadVideo>` and `<Audio>` with a real `src` into OTIO clips, uses `<Sequence from>` for placement, `trimBefore` for the source in-point, and `Gap.1` for everything with no media behind it.

It explicitly does not try to map text, captions, shapes, animations, effects or transitions. Those are React and have no media file, so no timeline format can carry them. The skill is written to leave them out and to tell the user exactly what it skipped, rather than pretend the export is lossless.

## Testing

An exported timeline was imported into DaVinci Resolve and came in as expected.

The rest was validated against the reference implementation (OpenTimelineIO 0.18.1) rather than by eye:

- A composition with a title card, a trimmed video, captions and a music bed exports to a file that parses, round-trips through `otio_json`, and converts to FCP7 XML for the Premiere path
- `otiostat`: 2 tracks of equal length, 3 clips, 00:00:16:00, matching `durationInFrames={480}` at 30fps
- Every media reference resolves on disk, and every `source_range` fits inside the real file length as reported by `ffprobe`
- Conforming the timeline with ffmpeg reproduces the edit: the frame at timeline 5s matches the source at 6s at 52.7 dB PSNR, against 14.6 dB for a deliberately one-second-off control, which confirms `trimBefore` is honoured
- Four deliberately broken exports (ignored `trimBefore`, missing gap, missing `available_range`, captions emitted as clips) are each caught, so the passing result means something

Repo checks: `bun run checkskills` passes, the four skill-related test files in `packages/cli` pass (5 tests), the `codex-plugin`, `claude-code-plugin` and `kimi-code-plugin` test suites pass, and so do `lint` and `formatting` for the affected packages.

Two details the testing corrected in the skill text: the FCP XML adapter is no longer bundled with OpenTimelineIO and needs `otio-fcp-adapter` plus `-O fcp_xml`, and that converter fails on any media reference without an `available_range`, so the skill tells the agent to fill it in from `npx remotion ffprobe`.


pureluminaryglow-hue (approved):

JonnyBurger:
Thanks a lot for this!

Finally got around to test it, and with some caveats, it worked well!
To get it super accurate, I made the skill pre-render videos in Remotion if there is an incompatibility.

While this is not yet so battle-tested, I put it just in the Docs under Tooling for now.
If it works well, we will be moving it to the official Remotion Skills though!

PR #10478, opened 2026-08-16
`@remotion/timeline-utils`: Guard waveform canvas against negative width

Fukucchi:
## Summary

Guard the Studio audio waveform worker against zero or negative canvas dimensions.

## Problem

`getTimelineSequenceLayout` intentionally lays out very short sequences with a track width of 0 (or slightly negative, since the border width is subtracted from the spatial width). For audio sequences, that width is forwarded verbatim to the waveform worker, which assigns it to the `OffscreenCanvas`:

```
Failed to set the 'width' property on 'OffscreenCanvas': Value is outside the 'unsigned long' value range.
```

This surfaces as an error overlay in Studio whenever a composition contains many short audio sequences (e.g. a few-frame SFX cue) and the timeline is zoomed out or the window is narrow enough that a track rounds to ≤ 0 px. #7293 already guards `drawBars` against a zero-width canvas, but the `=== 0` check lets negative values through, and the assignment in the worker happens before `drawBars` is reached.

## Fix

- `audio-waveform-worker.ts`: clamp `width` / `height` to ≥ 0 before assigning them to the transferred canvas (and ≥ 1 for the temporary loop canvas), so a negative layout width becomes a harmless empty canvas.
- `draw-peaks.ts`: widen the existing early return from `=== 0` to `<= 0`.

No behavior change for normal-width tracks.

## Repro

Any composition with a handful of `<Audio>` sequences of ~4 frames on a long timeline (e.g. 170 s at 30 fps); shrink the Studio window until the timeline error appears.

### DIFF
--- a/packages/timeline-utils/src/audio-waveform-worker.ts
+++ b/packages/timeline-utils/src/audio-waveform-worker.ts
@@ -63,7 +63,7 @@ const drawPartialWaveform = (
 	});
 	const targetCanvas = new OffscreenCanvas(
 		Math.max(1, Math.ceil(loopWidth)),
-		message.height,
+		Math.max(1, Math.ceil(message.height)),
 	);
 	drawBars({
 		canvas: targetCanvas,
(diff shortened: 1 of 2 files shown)


Fukucchi:
Closing this — the premise is wrong for current `main`.

I originally hit this on 4.0.457, but `getWidthOfTrack` in `packages/studio/src/helpers/get-timeline-sequence-layout.ts` has since been clamped to `Math.max(0, base - SEQUENCE_BORDER_WIDTH + nonNegativeMarginLeft)` (in #7303, which also added regression tests), and `getTimelineMediaVisualizationLayout` clamps again. The only sender of the worker's `render` message passes `Math.ceil(visualizationWidth)` from that clamped value, so a negative width cannot reach the canvas on `main`. Combined with the `w === 0` guard from #7293, both crashes I saw are already fixed.

That makes this change unreachable defensive code with no way to write a failing test for it, so it is not worth the maintenance. Sorry for the noise, and thanks for the quick fixes back in May.

PR #10511, opened 2026-08-18
`remotion`: Fix InvalidStateError when the HtmlInCanvas layout effect re-runs

alec-watts:
## Problem

`transferControlToOffscreen()` may only be called once per canvas — a second call throws:

```
InvalidStateError: Cannot transfer control from a canvas for more than one time.
```

The layout effect in `<HtmlInCanvas>` that transfers the placeholder canvas can run more than once for the same canvas element:

- Its dependencies (`onPaintCb` and what it closes over) can change identity without the size-keyed canvas remounting.
- React StrictMode intentionally double-invokes effects in development, so any app rendering `<HtmlInCanvas>` with a custom `onPaint`/`onInit` under StrictMode hits the throw immediately.

We hit this in production in Kino's editor.

## Fix

Cache the transferred `OffscreenCanvas` per canvas element in a module-level `WeakMap` and reuse it when the effect re-runs.

## Tests

- The existing test mock for `transferControlToOffscreen` now mirrors real browser behavior by throwing on a second transfer of the same canvas, which makes the existing `transferControlToOffscreenCalls` assertions meaningful.
- Added a StrictMode regression test. It fails without the fix (`InvalidStateError` surfaces through `cancelRender`) and passes with it.

All 17 tests in `packages/core/src/test/html-in-canvas.test.tsx` pass.

### DIFF
--- a/packages/core/src/HtmlInCanvas.tsx
+++ b/packages/core/src/HtmlInCanvas.tsx
@@ -191,6 +191,31 @@ export type HtmlInCanvasOnPaintParams = {
 	readonly pixelDensity: number;
 };
 
+// `transferControlToOffscreen()` may only be called once per canvas — a second
+// call throws an `InvalidStateError`. The layout effect that needs the
+// `OffscreenCanvas` can run more than once for the same canvas element: its
+// dependencies (e.g. the paint callback) can change identity without the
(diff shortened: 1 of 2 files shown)


JonnyBurger:
Thanks!

PR #10512, opened 2026-08-18
`@remotion/web-renderer`: Fix missing HtmlInCanvas WebGL output in software composition

alec-watts:
## Problem

When the web renderer composes frames in software — the path used whenever native nested html-in-canvas is unavailable, which is every stable browser today (Chromium disabled nested html-in-canvas by default in 149) — `<HtmlInCanvas>` components with `onPaint`/`onInit` WebGL effects render incorrectly, for two independent reasons:

1. **Transparent readback.** The placeholder canvas's control is transferred to an `OffscreenCanvas` (required for `texElementImage2D` to work), and a transferred canvas yields transparent pixels to `drawImage()`. The compositor drew the placeholder, so all painted WebGL output (shader effects, etc.) was silently missing from stills and exports.
2. **Double-drawn children.** The compositor then re-rasterized the canvas's layout-only DOM children on top, painting crisp HTML over where the canvas output belongs. Browsers never display `layoutsubtree` children until they are drawn into the canvas, so software composition must skip them too.

The visible symptom: a still/export where the WebGL effect is absent and the un-effected HTML is shown instead, while the preview looks correct.

## Fix

- `remotion`: after each paint during client-side rendering, `<HtmlInCanvas>` blits the painted frame into a readable shadow `OffscreenCanvas` — in the same task, before the WebGL drawing buffer can be discarded — and exposes it on the placeholder canvas (`__remotionRenderReadback`).
- `@remotion/web-renderer`: `drawDomElement` prefers the readback copy when drawing a canvas element, and `processNode` returns `skip-children` for layout-subtree canvases (mirroring browser behavior).

The readback only runs during client-side rendering (`isRendering && isClientSideRendering`), so preview performance is unaffected.

## Tests

- All existing tests in `packages/core/src/test/html-in-canvas.test.tsx` pass; the readback is a no-op in environments without `OffscreenCanvas`.
- Verified end-to-end in Electron (Chromium 152) with a composition using `<HtmlInCanvas>` CRT shader effects: before this change, stills rendered without the effect; after it, still and video output match the preview pixel-for-pixel.


alec-watts:
Good call — I added an automated repro to this PR: `includes custom onPaint output when composing in software` in `packages/web-renderer/src/test/html-in-canvas.test.tsx`, with a minimal fixture.

The key to reproducing it is that the composition must use a **custom `onPaint`** (or `onInit`). Effects-only usage like the existing `nestedHtmlInCanvas` fixture takes the direct layout-canvas path (`usesDirectLayoutCanvas === true`), where the placeholder canvas holds readable pixels — that path was always fine, which is why the existing snapshot tests never caught this. A custom `onPaint` forces `transferControlToOffscreen()`, and a transferred canvas yields transparent pixels to `drawImage()`, so the software composer loses everything the handler painted.

The fixture makes the failure unambiguous without WebGL or screenshots: `onPaint` ignores the captured HTML (solid red children) and fills the canvas solid green, then the test probes one center pixel of the still:

- **green** → the painted output made it into the still (fixed)
- **red** → the layout-only children were re-rasterized over the canvas (the skip-children half of the bug)
- **blue/transparent** → the transferred canvas read back as empty (the readback half)

Verified both ways on this branch: on `main` sources the test fails with `expected 255 to be less than 64` (red wins — children painted over the lost output); with this PR it passes.

One note on running it locally: my stable Chrome 151 doesn't expose `drawElementImage` by default, so the test self-skips via `supportsNativeHtmlInCanvas()` like the other tests in the file; I verified it by launching Chromium with `--enable-blink-features=CanvasDrawElement`.

For a manual repro outside the test suite: any composition with `<HtmlInCanvas onPaint={...}>` rendered via `renderStillOnWeb` while the software composer is active (e.g. `setForceDisableHtmlInCanvasForTesting(true)`, or any environment without native nested support) — the still shows the raw HTML children instead of the painted canvas, while the preview looks correct.

JonnyBurger:
Closing in favor of a replacement that models custom painting as part of the same capture → paint → effects → publish pipeline and covers ElementImage consumption.

PR #10543, opened 2026-08-18
`@remotion/captions`: Add silence-based caption page breaks

### LINKED ISSUE #10542: createTikTokStyleCaptions(): opt-in page breaks on punctuation and silence gaps
Following up from #10476 

Two optional fields on `CreateTikTokStyleCaptionsInput`, both off by default so existing output stays the same

`breakOnPunctuation?: boolean` (default false). A caption ending in sentence-ending punctuation (. ! ? … plus closing quotes/brackets) finishes the page, so you stop getting pages like "the blank page. This". No abbreviation detection, "Dr." also breaks. Keeping it a plain punctuation match so the rule stays explicit, same reasoning as the whitespace convention.

`silenceGapMs?: number` (default off). A silence of at least this length between captions starts a new page. The previous page's durationMs still extends to the next page start so captions hold through the pause, same as today.

## Semantics

Both breaks only happen at word boundaries (space-prefixed captions), never inside a word.
Both reset the accumulated duration window that combineTokensWithinMilliseconds measures.
Both only add earlier breaks. combineTokensWithinMilliseconds stays the upper bound, pages can only get shorter, never longer.

Also considereda shouldStartNewPage callback that receives the accumulated page, the incoming caption, and the gap, and returns whether to break. It is the most unopinionated shape, and every future rule (char caps, word counts, abbreviation handling, etc) becomes user code instead of a new option.

Went with two flat options anyway, for three reasons..

1. The common cases should need no code. With a callback everyone re-writes the same punctuation regex and gap check, which is the same problem that #4555 was.

2. A callback is just generally a bigger API commitment. The context object's shape becomes public contract, and it hands users a new way to break determinism with an impure callback. This is the biggest reason not to in my mind.

3. Options don't rule out the callback. If its wanted, shouldStartNewPage can come later and these two become built-on presets for it. The reverse migration just seems messier IMO.

That being said.. I can reshape if you'd prefer the callback. 😃 

PR incoming with tests and docs
Cheers!

Fats403:
## Summary

- add opt-in `breakOnPunctuation` to finish a page after sentence-ending punctuation (plain match, no abbreviation heuristics)

- add opt-in `silenceGapMs` to start a new page after a silence between captions

- both default off, existing output is byte-identical (covered by an equivalence test)

- punctuation check runs on a fixed tail slice, O(1) per caption

- docs for both options (AvailableFrom markers set to 4.0.513, can bump if needed)

## Testing

- `bun run test` in `packages/captions` (11 pass)

- `bun run make`

- `bun run lint` (no new warnings)

Closes #10542

### DIFF
--- a/packages/captions/src/create-tiktok-style-captions.ts
+++ b/packages/captions/src/create-tiktok-style-captions.ts
@@ -16,7 +16,11 @@ export type TikTokPage = {
 export type CreateTikTokStyleCaptionsInput = {
 	captions: Caption[];
 	combineTokensWithinMilliseconds: number;
+	breakOnPunctuation?: boolean;
+	silenceGapMs?: number;
 };
+
+const SENTENCE_ENDING_PUNCTUATION = /[.!?…]["'”’)\]]*$/;
 
 export type CreateTikTokStyleCaptionsOutput = {
 	pages: TikTokPage[];
@@ -25,6 +29,8 @@ export type CreateTikTokStyleCaptionsOutput = {
 export const createTikTokStyleCaptions = ({
 	captions,
 	combineTokensWithinMilliseconds,
+	breakOnPunctuation,
+	silenceGapMs,
 }: CreateTikTokStyleCaptionsInput): CreateTikTokStyleCaptionsOutput => {
 	const tikTokStyleCaptions: TikTokPage[] = [];
 	let currentText = '';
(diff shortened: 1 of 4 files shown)


Fats403:
@pullfrog Both addressed in e8b48bf: markers bumped to 4.0.514, and silenceGapMs: 0 was intentional behaviour but good catch, it is now documented (word-by-word pages) with a test pinning it. Kept the >= semantics since "at least 0ms" reads consistently and word-per-page is a useful setting.

JonnyBurger:
The silence gap looks good to me.

The breakage on punctuation maybe helps in some cases, but also introduces problems like unintentional page break for things like "Dr. " or "Prof. ".

We should go back to the whiteboard with that one...

Issue #7458, opened 2026-05-19
HtmlInCanvas cannot draw with effects in latest Chrome Canary


JonnyBurger:
In latest Chrome canary: 

`experimental-controls-showcase` composition

```
InvalidStateError: Failed to execute 'drawElementImage' on 'OffscreenCanvasRenderingContext2D': The source was captured from a different canvas.
at defaultOnPaint (../core/dist/esm/index.mjs:3566:?)
at HTMLCanvasElement.<anonymous> (../core/dist/esm/index.mjs:3641:?)
```

PR #10555, opened 2026-08-19
`@remotion/web-renderer`: Preserve HtmlInCanvas WebGL output in software composition

JonnyBurger:
## Summary

- add a browser-level WebGL repro for the missing custom `<HtmlInCanvas onPaint>` output reported in #10512
- keep the client-side rendering placeholder readable by painting custom callbacks on a detached `OffscreenCanvas`, then publishing the finished bitmap to the layout canvas before the frame becomes ready
- make the DOM compositor treat `layoutSubtree` children as layout-only and enable `CanvasDrawElement` in Chromium browser tests so the regression cannot silently return early

## Architecture

The DOM canvas remains the canonical readable output surface. Preview and server rendering keep the transferred `OffscreenCanvas` path, while client-side rendering uses a detached paint target because transferred placeholders read back as transparent.

This avoids adding a private readback property to `HTMLCanvasElement` or teaching `@remotion/web-renderer` about a component-specific side channel.

## Verification

- Red/green verified: the final WebGL pixel regression fails when the detached-surface publication is disconnected and passes when restored
- `bunx vitest src/test/html-in-canvas.test.tsx --run`
- `bun run testwebrenderer` — 693 passed, 27 skipped
- `bun run build`
- `bun run stylecheck`

### DIFF
--- a/packages/core/src/HtmlInCanvas.tsx
+++ b/packages/core/src/HtmlInCanvas.tsx
@@ -182,8 +182,7 @@ declare global {
 
 export type HtmlInCanvasOnPaintParams = {
 	/**
-	 * The `OffscreenCanvas` from {@link HTMLCanvasElement.transferControlToOffscreen}
-	 * on the layout `<canvas>` (same logical canvas as the forwarded ref).
+	 * The `OffscreenCanvas` used to paint the layout `<canvas>`.
 	 */
 	readonly canvas: OffscreenCanvas;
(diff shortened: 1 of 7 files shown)


JonnyBurger:
Superseded. The detached paint target reintroduced the ElementImage context-linkage problem fixed in #7458; replacing this with a unified paint/effects/publication pipeline.

PR #10560, opened 2026-08-19
`@remotion/web-renderer`: Preserve HtmlInCanvas paint pipeline output

JonnyBurger:
## Summary

- model `<HtmlInCanvas>` rendering as one capture → paint → effects → publish pipeline, independent of whether the paint stage is built-in or supplied through `onPaint`
- publish every completed `<HtmlInCanvas>` frame into a readable surface while GPU drawing buffers are still valid
- have the DOM compositor consume the published frame and treat `layoutSubtree` children as paint inputs instead of drawing them again
- cover both an `ElementImage`-consuming custom painter and a WebGL painter initialized through `onInit`

The custom painter still uses the transferred `OffscreenCanvas` tied to the layout canvas, preserving the same-logical-canvas requirement from #7458. Publication is a property of the paint pipeline itself; the web renderer no longer opts into it or controls it.

Supersedes #10512 and #10555.

## Verification

- Red/green verified: on `main`, software composition returned the raw red layout child instead of the custom green output; the final test passes with the publication stage connected
- `bun test src/test/html-in-canvas.test.tsx` in `packages/core` — 17 passed
- `bunx vitest src/test/html-in-canvas.test.tsx --run` in `packages/web-renderer` — 21 passed across Chromium, Firefox, and WebKit
- `bun run testwebrenderer` — 691 passed, 27 skipped; two unrelated screenshot-stability timeouts passed immediately when rerun in isolation (6/6 browser cases)
- `bun run build`
- `bun run stylecheck`

### DIFF
--- a/packages/core/src/HtmlInCanvas.tsx
+++ b/packages/core/src/HtmlInCanvas.tsx
@@ -29,6 +29,10 @@ import {
 	transformSchema,
 	type InteractivitySchema,
 } from './interactivity-schema.js';
+import {
+	CanvasFrameOutputContext,
+	publishCanvasFrame,
+} from './published-canvas-frame.js';
 import type {AbsoluteFillLayout} from './Sequence.js';
 import {Sequence} from './Sequence.js';
 import {useCropStyle} from './use-crop-style.js';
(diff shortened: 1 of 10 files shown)


JonnyBurger:
Closing because client-side rendering of compositions containing <HtmlInCanvas> depends on unsupported nested HTML-in-canvas behavior. A replacement PR will roll back the software-composer support claim instead of adding a private publication protocol.

PR #10597, opened 2026-08-19
Elements: Track requests sent to Studio

samohovets:
## Summary

- Send a small event after `remotion.dev` sends an Element request to Studio.
- Send only the Element slug.
- Keep the current installation request and user interface unchanged.

The API and database change is in https://github.com/remotion-dev/pro/pull/823.

## Data collection

The browser sends the Element slug to `remotion.pro`. It sends the event only after Studio accepts the installation request.

The application stores only these values:

- a generated row ID;
- the event time;
- the Element slug.

The application does not store an IP address, a cookie, a user ID, project data, source code, dependency data, or account data.

The hosting service can process standard request data, such as an IP address. The application does not copy this data to the event table.

The event shows that the request reached Studio. It does not show that the user confirmed the installation. We do not use an external analytics service.

@MehmetAdemi Please review this section and tell us if we must change the privacy policy.

## Tests

- `bun run build`
- `bun run stylecheck`
- `bunx turbo run test --filter=docs`

### DIFF
--- a/packages/docs/src/components/Elements/ElementPage.tsx
+++ b/packages/docs/src/components/Elements/ElementPage.tsx
@@ -73,17 +73,17 @@ export const ElementPage: React.FC<ElementPageProps> = ({
 			return;
 		}
 
-		if (window.location.origin === 'https://www.remotion.dev') {
-			navigator.sendBeacon(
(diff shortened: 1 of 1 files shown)


MehmetAdemi:
Privacy Policy wise this looks good to me, so we don't need to change anything.

PR #10662, opened 2026-08-20
`@remotion/media`: paint completed frame of a superseded seek to fix paused-scrub freeze

### LINKED ISSUE #10661: @remotion/media: preview freezes during paused scrubbing — superseded seeks discard every completed frame
## Bug

When scrubbing (dragging the playhead) while paused in a Player using `<Video>` from `@remotion/media`, the preview freezes for the entire drag. The frame only updates once the drag stops.

## Environment

- remotion / @remotion/media / @remotion/player 4.0.499 (code path unchanged on current `main`, checked v4.0.514)
- Chrome, macOS + Windows
- `<Player>` embedding a composition with `<Video>` from `@remotion/media`

## Reproduction

1. Render a `<Player>` with a composition using `<Video>` from `@remotion/media`.
2. Pause.
3. Drag the seek position continuously for a second or more.

Expected: the preview repaints as the playhead moves, like scrubbing in any NLE.
Actual: the preview shows the same frame for the whole drag, then jumps when you release.

## Cause

In `packages/media/src/video-iterator-manager.ts`, `startVideoIterator` discards its result when the seek nonce has gone stale:

```ts
if (nonce.isStale()) {
	return;
}
```

During a continuous drag, every pointer move stales the previous seek's nonce *before* that seek's decode completes. So each iterator finishes a full decode and then throws the frame away undrawn — for every seek in the drag. 100% of decoded frames are discarded and nothing ever paints: a livelock where all the work happens and none of it is visible.

## Suggested fix

When the nonce is stale but the decode has already completed, paint the completed frame before returning. The newer seek is already queued and lands last, so painting is strictly monotonic progress toward the final target — nothing can go backwards:

```ts
if (nonce.isStale()) {
	if (!videoFrameIterator.isDestroyed() && iterator.initialFrame) {
		await drawFrame(iterator.initialFrame);
	}
	return;
}
```

We have been running this as a small patch on the dist build in production for a few weeks; paused scrubbing repaints continuously with no regressions observed in playback, seek-while-playing, or rapid direction changes. Happy to open a PR.

ElScelt:
Fixes #10661

During paused scrubbing, every seek's nonce goes stale before its decode completes, so `startVideoIterator` discarded 100% of decoded frames undrawn and the preview froze for the entire drag.

This paints the already-completed frame before returning when the nonce is stale. The superseding seek is already queued and always lands last, so the painted frame is monotonic progress toward the target, never a step backwards.

Guarded on `videoFrameIterator.isDestroyed()` and `iterator.initialFrame` (media ended), matching the guards on the non-stale path below it.

Verified in a production editor built on `@remotion/player` + `@remotion/media`: paused scrubbing repaints continuously instead of freezing, with no regressions observed in playback, seek-while-playing, or rapid seek-direction changes over several weeks of use.

### DIFF
--- a/packages/media/src/video-iterator-manager.ts
+++ b/packages/media/src/video-iterator-manager.ts
@@ -186,6 +186,15 @@ export const videoIteratorManager = async ({
 		}
 
 		if (nonce.isStale()) {
+			// A superseded seek has still paid for a completed decode. During a
+			// paused scrub, every seek is superseded by the next pointer move
+			// before its decode lands, so returning here undrawn would discard
+			// every frame and freeze the preview for the whole drag. Painting
+			// the completed frame is safe: the newer seek is already queued and
+			// lands last, so this is monotonic progress toward the target.
+			if (!videoFrameIterator.isDestroyed() && iterator.initialFrame) {
+				await drawFrame(iterator.initialFrame);
+			}
 			return;
 		}
 
--- a/packages/media/src/video-iterator-manager.ts
+++ b/packages/media/src/video-iterator-manager.ts
@@ -192,6 +192,7 @@ export const videoIteratorManager = async ({
 			if (!videoFrameIterator.isDestroyed() && iterator.initialFrame) {
 				await drawFrame(iterator.initialFrame);
 			}
+
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


ElScelt:
CI note: the two red checks across runs look like flakes rather than this change — run 1 failed on the sequence-shift-keyframes browser test (passes in run 2 with the same code), run 2 failed on the web-renderer onProgress throttle timing test, which renders a pure SVG composition and does not load @remotion/media at all. Everything media-related is green in both runs. Happy to rebase or adjust if a re-run shows otherwise.

JonnyBurger:
Good idea!

PR #10812, opened 2026-08-25
New package: `@remotion/whisper-webgpu`

JonnyBurger:
## Summary

- add `@remotion/whisper-webgpu` for timestamped Transformers.js Whisper transcription
- support selectable tiny, base, small, medium, and large-v3-turbo models with WebGPU and WASM backends
- convert word timestamps directly to `@remotion/captions` output
- migrate `remotion.dev/transcribe` from `@remotion/whisper-web` to the selectable WebGPU implementation
- add package documentation, generated preview cards, and an interactive example

Cross-origin isolation is not required. WebGPU uses a secure context, while the WASM fallback uses one thread when the page is not cross-origin isolated. The old `/transcribe` COOP/COEP headers have been removed.

## Test plan

- `bun run build`
- `bun run stylecheck`
- `bun test packages/whisper-webgpu/src/test/transcribe.test.ts`
- `bun run typecheck`, `bun run lint`, `bun run test`, `bun run build-page`, and `bun run build-spa` in `packages/convert`
- browser smoke test of `/transcribe` using `dialogue.wav`: tiny.en loaded through WebGPU and rendered timestamped captions
- bundle the Remotion example project
- type-check all documentation Twoslash snippets

## Preview

- [`remotion.dev/transcribe`](https://remotion-git-codex-whisper-webgpu-remotion.vercel.app/transcribe)
- [`@remotion/whisper-webgpu`](https://remotion-git-codex-whisper-webgpu-remotion.vercel.app/docs/whisper-webgpu)
- [`canUseWhisperWebGpu()`](https://remotion-git-codex-whisper-webgpu-remotion.vercel.app/docs/whisper-webgpu/can-use-whisper-webgpu)
- [`getAvailableModels()`](https://remotion-git-codex-whisper-webgpu-remotion.vercel.app/docs/whisper-webgpu/get-available-models)
- [`loadWhisperModel()`](https://remotion-git-codex-whisper-webgpu-remotion.vercel.app/docs/whisper-webgpu/load-whisper-model)
- [`transcribe()`](https://remotion-git-codex-whisper-webgpu-remotion.vercel.app/docs/whisper-webgpu/transcribe)
- [`toCaptions()`](https://remotion-git-codex-whisper-webgpu-remotion.vercel.app/docs/whisper-webgpu/to-captions)
- [`resampleTo16Khz()`](https://remotion-git-codex-whisper-webgpu-remotion.vercel.app/docs/whisper-webgpu/resample-to-16khz)
- [`disposeWhisperModel()`](https://remotion-git-codex-whisper-webgpu-remotion.vercel.app/docs/whisper-webgpu/dispose-whisper-model)


packages/whisper-webgpu/src/backend.ts
```diff
@@ -0,0 +1,27 @@
+export type WhisperWebGpuBackend = 'auto' | 'webgpu' | 'wasm';
+export type ResolvedWhisperWebGpuBackend = Exclude<
+	WhisperWebGpuBackend,
+	'auto'
+>;
+
+const hasWebGpu = () => {
+	return (
+		typeof navigator !== 'undefined' &&
+		'gpu' in navigator &&
```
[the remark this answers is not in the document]
JonnyBurger: Fixed in a590e6ee80. Backend resolution now awaits navigator.gpu.requestAdapter(); auto falls back to WASM when it returns null or rejects, while an explicit WebGPU request reports an unsupported result. The package test covers both the no-adapter fallback and the usable-adapter WebGPU path, with red/green verification.

packages/docs/docs/whisper-webgpu/can-use-whisper-webgpu.mdx
```diff
@@ -0,0 +1,40 @@
...
+---
+
+# canUseWhisperWebGpu()<AvailableFrom v="4.0.517" />
+
+Checks whether WebGPU or the WASM fallback can run in the current browser.
+
+```ts twoslash title="support.ts"
+import {canUseWhisperWebGpu} from '@remotion/whisper-webgpu';
+
+const result = await canUseWhisperWebGpu();
```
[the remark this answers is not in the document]
JonnyBurger: Fixed in a590e6ee80. canUseWhisperWebGpu() is now genuinely async because it probes requestAdapter(), so the documented await is now required and the example remains correct.

JonnyBurger:
Addressed all Pullfrog feedback in a590e6ee80:

- `auto` now awaits `navigator.gpu.requestAdapter()` and falls back to WASM when no usable adapter is available; explicit WebGPU reports unsupported.
- `canUseWhisperWebGpu()` is now genuinely async, matching the documented `await`.
- audio decoding now uses `OfflineAudioContext`, avoiding a live audio device context.
- the docs now position `@remotion/whisper-webgpu` as preferred for new browser transcription integrations while explicitly keeping `@remotion/whisper-web` supported and non-deprecated.
- regression coverage exercises unavailable- and available-adapter paths; red/green verified.

JonnyBurger:
Fixed the two issues found during manual testing on `/transcribe`:

- Model progress now uses only Transformers.js aggregate `progress_total` events and is clamped monotonically, so per-file events can no longer make the bar jump backwards.
- Backwards or missing word end timestamps now fall back to the next word start (or the audio duration for the final word) instead of failing the transcription.

Added regression coverage for the real mixed progress event sequence and for a `free` chunk with `[1.8, 1.7]`. Verified the uncached `tiny` model in the local `/transcribe` UI: sampled progress was `0, 0, 23, 23, 23, 34, 40, 46, 54, 61, 67, 71`, then transcription completed successfully.

JonnyBurger:
Fixed the premature 100% model-download state in `732aacb81b`.

Transformers.js can report 100% for the first large artifact before it discovers the remaining files; preserving monotonic aggregate progress therefore kept the UI at 100% while more than a gigabyte was still loading. Progress now sums monotonic per-file byte counts against the selected model/backend's known total size, caps active loading below 100%, and reports 100% only on `ready`.

Added a public `loadWhisperModel()` regression fixture where the encoder reports 100% before the decoder begins. Red/green verified. Package build/lint/format/tests and Convert typecheck, 52 tests, and production SPA build pass.

JonnyBurger:
Made `/transcribe` WebGPU-only in `26d9a9ebbc`:

- model loading and transcription now explicitly request `backend: 'webgpu'`, so the page cannot fall back to WASM
- removed backend state from the UI workflow
- model choices show only the WebGPU download size
- status copy now uses `WebGPU`, not `WEBGPU`
- added a rendered selector regression asserting the visible WebGPU size and absence of a WASM alternative (red/green verified)

Verified the actual local page renders `tiny.en 119.7 MB WebGPU`. Convert typecheck, 53 tests, production SPA build, full monorepo build, and 262 style tasks pass.

packages/whisper-webgpu/src/transcribe.ts
```diff
@@ -0,0 +1,120 @@
...
+	const transcriber = await getLoadedWhisperPipeline({
+		model,
+		onProgress: onModelLoadProgress,
+	});
+
+	const output = (await transcriber(channelWaveform, {
+		return_timestamps: 'word',
+		chunk_length_s: chunkLengthInSeconds,
+		stride_length_s: strideLengthInSeconds,
+		...(language === 'auto' ? {} : {language}),
```
nico-martin, line 72: This passes language to English-only .en models. Transformers.js rejects any language value when is_multilingual is false, so the documented {model: 'small.en', language: 'en'} call throws. Could we use getModelInfo(model).multilingual to omit language for .en models and reject non-English values with a package-level error?

packages/whisper-webgpu/src/transcribe.ts
```diff
@@ -0,0 +1,120 @@
...
+
+type TransformersJsTranscription = {
+	text: string;
+	chunks?: TransformersJsWord[];
+};
+
+export const transcribe = async ({
+	channelWaveform,
+	model,
+	language = 'auto',
```
nico-martin, line 41: Transformers.js 4.2 does not implement Whisper language detection. Omitting language logs "No language specified", defaulting to English (en) and transcribes as English. Could we require a language for multilingual models or explicitly default and document it as 'en'?

packages/whisper-webgpu/src/transcribe.ts
```diff
@@ -0,0 +1,120 @@
...
+	chunkLengthInSeconds = 30,
+	strideLengthInSeconds = 5,
+	onModelLoadProgress,
+}: TranscribeOptions): Promise<WhisperWebGpuTranscription> => {
+	if (channelWaveform.length === 0) {
+		throw new Error('The audio waveform is empty.');
+	}
+
+	if (chunkLengthInSeconds <= 0) {
```
nico-martin, line 50: This validation accepts NaN, as does the stride validation below. Transformers.js then calculates the chunk window, jump, and offset as NaN. Its offsetEnd >= audio.length condition never becomes true, so the chunk-building loop does not terminate. Please require Number.isFinite() for both values before applying the range checks.

packages/whisper-webgpu/src/load-whisper-model.ts
```diff
@@ -0,0 +1,162 @@
...
+			status: 'ready',
+			file: null,
+			progress: 1,
+			loadedBytes: totalBytes,
+			totalBytes,
+		});
+		return {alreadyLoaded: true};
+	}
+
+	const {pipeline} = await import('@huggingface/transformers');
```
nico-martin, line 54: The cache was checked above, but no entry is registered before this asynchronous import. Two simultaneous calls can both observe an empty cache and initialize separate pipelines before the map is updated at line 115. Browser testing confirmed that concurrent pipeline() calls return distinct live sessions. One session can therefore be overwritten in the map and become impossible to dispose through this API. Could we store a shared initialization promise before awaiting the import?

packages/whisper-webgpu/src/load-whisper-model.ts
```diff
@@ -0,0 +1,162 @@
...
+}: DisposeWhisperModelOptions = {}): Promise<void> => {
+	const matching = [...pipelines.entries()].filter(([loadedModel]) => {
+		return model === undefined || loadedModel === model;
+	});
+
+	await Promise.all(
+		matching.map(async ([key, loaded]) => {
+			pipelines.delete(key);
+			const loadedPipeline = await loaded;
+			await loadedPipeline.dispose();
```
nico-martin, line 159: Nothing tracks whether this pipeline is currently being used by transcribe(). Disposing during active inference was reproduced in the browser and rejected with "cannot run inference. invalid session id". Could we track active users and defer disposal until initialization and all active transcriptions have completed?

nico-martin (commented):
I found a few problems with the current implementation, other wise I'm really looking forward to see Transformers.js in remotion ❤️

PR #10837, opened 2026-08-26
Fix captions Elements not being movable or hideable in Studio

### LINKED ISSUE #10778: Moving Pill Captions cannot be moved in Studio
Tested in browser studio, there is no eye to toggle visibility.
Might be a problem with Elements, the Studio, the schema, or Browser Studio.

Needs investigation

https://www.remotion.dev/elements/captions/moving-pill-captions/

MadeNavaneeth:
Fixes #10778

## Problem

Opening any of the caption Elements (Moving Pill Captions, Popping Word Captions, Word Highlight Captions) in the (Browser) Studio and installing it showed two symptoms:

1. **No eye to toggle visibility** on the timeline item
2. **The Element could not be moved** — drag overrides did not behave / persist correctly

## Root cause

The three caption Elements exported a **custom wrapper component** around their `Interactive.withSchema()` layer:

```tsx
const MovingPillCaptionsLayer = Interactive.withSchema({...});

export const MovingPillCaptions = ({captions, style, ...props}) => {
  // renders <MovingPillCaptionsLayer> internally
};
```

Only the component returned by `Interactive.withSchema()` is registered for Studio stack attribution (`addSequenceStackTraces`). Because the registered layer was rendered *inside the Element file* rather than directly in the composition, the Studio resolved the timeline item's stack to the internal `<MovingPillCaptionsLayer>` JSX (e.g. line 513 of `moving-pill-captions.element.tsx`) instead of the `<MovingPillCaptions />` usage in `Composition.tsx`.

Consequences of the subscription targeting the wrong node:

- `hidden` arrives at the internal node via a props spread, so its status is `computed` — the eye toggle requires status `static`, so it was never rendered
- Drag overrides and saves were keyed to the Element file internals instead of the composition, so moving the Element did not persist to the composition

Elements that export the `withSchema()` result directly (e.g. `<AudioOscilloscope>`) are unaffected — verified as a control in Browser Studio.

## Fix

Restructure the three caption Elements to export the `Interactive.withSchema()` component directly, following the documented pattern in [Make a component interactive](https://remotion.dev/docs/studio/make-component-interactive):

- The demo captions fallback moved into the inner component (`captions === undefined`)
- The demo sizing (900×180 area with a centered 681×252 captions region) moved into the inner component
- The `callerStyle` split was removed — it only existed to shield the caller's style from the wrapper's injected `style={{translate: '0px 0px'}}`, which is unnecessary when the public component *is* the schema-managed component

## Verification

New Browser Studio e2e regression test (`e2e/captions-elements.test.ts`) for all three Elements asserts after installing via the URL-fragment flow:

- the sequence-props subscription targets `/project/src/Composition.tsx` (not the Element file)
- the timeline item has a visibility eye
- dragging the Element on the canvas moves it
- saving persists the `translate` into `Composition.tsx` and leaves the Element file untouched

All 3 pass (previously: no eye, subscription targeted the Element file, moves didn't persist).

Also verified:

- `bun test src/test/elements.test.ts` in `packages/docs` (224 pass)
- Still renders of all three demo compositions are pixel-identical in layout to before
- Full Browser Studio e2e suite (14 pass)

### DIFF
--- a/packages/browser-studio/e2e/captions-elements.test.ts
+++ b/packages/browser-studio/e2e/captions-elements.test.ts
@@ -1,0 +1,222 @@
+import {readFileSync} from 'node:fs';
+import path from 'node:path';
+import {fileURLToPath} from 'node:url';
+import {
+	expect,
+	test,
+	type Frame,
+	type FrameLocator,
+	type Page,
+} from '@playwright/test';
+import {
+	createElementPayload,
+	StudioProtocolInternals,
+} from '@remotion/studio-protocol';
+
+const instrumentSequencePropsSubscriptions = async (page: Page) => {
+	const studioFrame = page.frames().find((frame) => frame !== page.mainFrame());
+	if (!studioFrame) {
+		throw new Error('No studio iframe');
+	}
+
+	await studioFrame.waitForFunction(
+		() =>
+			Boolean(
+				(window as unknown as {remotion_browserStudio: unknown})
+					.remotion_browserStudio,
+			),
+		null,
+		{timeout: 60_000},
+	);
+	await studioFrame.evaluate(() => {
+		const ops = (
+			window as unknown as {
+				remotion_browserStudio: {
+					subscribeToSequenceProps: (request: unknown) => Promise<unknown>;
+				};
+			}
+		).remotion_browserStudio;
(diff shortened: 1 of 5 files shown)


samohovets:
Already fixed in https://github.com/remotion-dev/remotion/pull/10798

PR #10864, opened 2026-08-26
core: guard window.remotion_delayRenderHandles initialization

### LINKED ISSUE #10856: Unguarded `window.remotion_delayRenderHandles = []` module init: a second bundled copy of remotion orphans armed delayRender timers (continueRender silently no-ops; long renders die at timeout)
## Summary

`packages/core/src/delay-render.ts` runs this at module top level:

```ts
if (typeof window !== 'undefined') {
	window.remotion_renderReady = false;
	if (!window.remotion_delayRenderTimeouts) {
		window.remotion_delayRenderTimeouts = {};
	}

	window.remotion_delayRenderHandles = [];   // <-- unguarded
}
```

The timeouts map is guarded; the handles array is not. When a bundle contains **two copies** of this module — in our case `remotion` (ESM) plus `remotion/no-react` (statically imported by `@remotion/studio`'s `renderEntry`) — the second copy's evaluation **wipes the handles array** during the same synchronous bundle startup.

Any `delayRender()` armed *before* the second copy evaluates (module-scope work — in our app, `@remotion/fonts` `loadFont()` calls) is then orphaned: `continueRenderInternal` requires `handleExists && environment.isRendering && timeoutEntry` before it clears, `handleExists` is now false, so **every `continueRender` for those handles silently no-ops** — `clearTimeout` is skipped and the map entry is retained. The async work completes fine, frames render fine (the wiped array makes `remotion_renderReady` flip true), and the orphaned timers fire at `timeoutInMilliseconds − 2000`, killing the render with:

```
A delayRender() "Loading font …" was called but not cleared after 58000ms
```

## Why this is nasty to diagnose

- **Only renders whose duration exceeds the timeout die.** Short compositions finish before the fuse, so the failure looks intermittent and load-correlated even though the wipe is deterministic.
- Raising `--timeout` just moves the death (we observed 58000 → 118000 → 298000 ms tracking the flag exactly, always `timeout − 2000`).
- The error names the *first-armed* handle (for us, always the first font in the manifest), which reads exactly like a font/network hang. We chased the static server, respawn behavior, and request starvation before instrumenting page state.
- Because the wipe also empties the gate array, `remotion_renderReady` can go true while module-scope delayRender work is still pending, so early frames can rasterize before fonts finish loading — nondeterministic first-render output.

## Evidence (remotion 4.0.499, Windows, webpack bundle via @remotion/bundler)

We beaconed `window.remotion_delayRenderHandles.length` and `Object.keys(window.remotion_delayRenderTimeouts).length` from the page. On **every page of every render**:

- immediately after arming ~716 module-scope `loadFont()` calls: `handles=716, timeouts=716`
- by the time the first font promise resolves: `handles=0, timeouts=716`
- after all 716 fonts load (every promise resolved, all files served locally in 3–4 s): `handles=0, timeouts=716` — all 716 timers still armed
- render dies at `timeout − 2000` naming the first face; shorter comps never surface it.

The served bundle contains the init block twice (`grep -c "remotion_delayRenderHandles = \[\]" bundle.js` → 2): once from `remotion`, once from `remotion/no-react`.

Verified the unguarded assignment is still present on `main` (packages/core/src/delay-render.ts) as of 2026-08-26. #5843 / #5850 fixed what looks like the same class of bug in the bundler's `index-html`, but the core module init can still collide with itself when both entry points end up in one bundle.

## Suggested fix

Guard the handles init the same way the timeouts map is guarded:

```ts
if (!window.remotion_delayRenderHandles) {
	window.remotion_delayRenderHandles = [];
}
```

(in all shipped builds — `cjs`, `esm/index.mjs`, `esm/no-react.mjs` all carry a copy).

## Related observation

While debugging we noticed `set-props-and-env.ts` stamps `window.remotion_attempt = 1` unconditionally for every page, including pages created by the delayRender-retry path. `delayRenderInternal` computes `retriesLeft = (options.retries ?? 0) - (scope.remotion_attempt - 1)`, so the "Retries left: N" printed into the timeout message never decreases, and `getRetriesLeftFromError` re-parses the same N each attempt — a delayRender with `retries > 0` that fails persistently appears to retry without bound. Happy to file separately if you prefer.

## Workaround we shipped in-app

After arming our module-scope delayRenders, a `queueMicrotask` (runs after the entire synchronous bundle evaluation, i.e. after any second copy's wipe, and before any network task can resolve) re-pushes our still-armed handles into the current `window.remotion_delayRenderHandles`, which restores the native `continueRender` path; a low-frequency sweeper mirrors `continueRenderInternal` for anything left. With that in place the same compositions render reliably at default flags.

MadeNavaneeth:
## Summary

Fixes #10856

### The problem

When a bundle contains two copies of the remotion module (e.g. `remotion` ESM + `remotion/no-react` statically imported by `@remotion/studio`), the second copys module-level evaluation wipes `window.remotion_delayRenderHandles = []`, orphaning any `delayRender()` calls armed before that point. `continueRender()` then silently no-ops for those handles.

### The fix

Add the same guard already used for `remotion_delayRenderTimeouts`:

```diff
-  window.remotion_delayRenderHandles = [];
+  if (!window.remotion_delayRenderHandles) {
+    window.remotion_delayRenderHandles = [];
+  }
```

### Verification

- ✅ 678 tests pass, 0 fail
- ✅ `tsgo --noEmit` clean

### DIFF
--- a/packages/core/src/delay-render.ts
+++ b/packages/core/src/delay-render.ts
@@ -42,7 +42,9 @@ if (typeof window !== 'undefined') {
 		window.remotion_delayRenderTimeouts = {};
 	}
 
-	window.remotion_delayRenderHandles = [];
+	if (!window.remotion_delayRenderHandles) {
+		window.remotion_delayRenderHandles = [];
+	}
 }
 
 const defaultTimeout = 30000;


JonnyBurger:
Sorry, this was already fixed by #10558

PR #10865, opened 2026-08-27
fix: lock Social Safe Zones overlay to 9:16 aspect ratio

### LINKED ISSUE #10835: Social Safe Zones overlay should have fixed dimensions
https://www.remotion.dev/elements/overlays/social-safe-zones/

Currently can be added to a widescreen composition:

Image

But the aspect ratio should always be locked to 9:16.
I suggest making it 1080x1920.

About sizing it correctly for 720x1280, I will make a separate issue

MadeNavaneeth:
Fixes #10835

## What
The `<SocialSafeZones>` overlay previously scaled to fill the full composition dimensions, which distorted the safe zone guides on widescreen compositions (e.g. 1920×1080). The safe zones are designed for 9:16 vertical video.

## Fix
The overlay container now maintains a fixed 9:16 aspect ratio (1080×1920) and is centered in the composition, regardless of composition dimensions. This ensures the safe zone paths and interface overlays are always accurate representations of actual TikTok/Instagram layouts.

## Testing
- TypeScript compiles clean (no new errors from this change)
- All existing studio tests pass

### DIFF
--- a/packages/docs/elements/overlays/social-safe-zones/social-safe-zones.tsx
+++ b/packages/docs/elements/overlays/social-safe-zones/social-safe-zones.tsx
@@ -66,7 +66,11 @@ const SocialSafeZonesInner = forwardRef<
 		ref,
 	) => {
 		const outlineRef = useRef<HTMLDivElement>(null);
-		const {height, width} = useVideoConfig();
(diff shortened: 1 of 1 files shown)


JonnyBurger:
Nope this is not how we wanna do this

PR #10873, opened 2026-08-27
`@remotion/studio`: Fix checkerboard bleeding through video edges

JonnyBurger:
## Summary

- detect video canvases whose rendered bounds exactly match the composition preview
- overscan only those full-bleed video canvases by one preview pixel so the checkerboard cannot bleed through their separately composited edges
- preserve the geometry of composition overlays and non-full-bleed videos
- add a 1920×1080 visual regression composition matching the `/NewVideo` preview geometry
- compare its opaque video perimeter pixels with the checkerboard disabled and enabled, and assert that overlay and smaller-video geometry does not move

## Testing

- `bun run build`
- `bun run stylecheck`
- `bunx playwright test e2e/studio.test.mts --grep "should preview media assets|should not let the checkerboard"`

### DIFF
--- a/packages/example/e2e/studio.test.mts
+++ b/packages/example/e2e/studio.test.mts
@@ -295,15 +295,17 @@ test.describe('visual mode', () => {
 				);
 			})
 			.toBeLessThan(1);
+	});
 
-		await page.goto(`${STUDIO_URL}/assets/framer.webm`);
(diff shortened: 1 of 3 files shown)


JonnyBurger:
Biig hack, probably don't wanna do this.

PR #10900, opened 2026-08-27
`@remotion/canvas`: Prevent outline controller update loops

### LINKED ISSUE #10897: `@remotion/canvas`: Prevent outline controller update loops
Hovering the preview canvas can crash the Studio with React error #185 (maximum update depth exceeded).

This reproduces in a bundled read-only Studio and on an immutable deployment of current `main`, so it is unrelated to the source-location changes in #10875.

## Reproduction

1. Use a recent Chrome with Canvas Draw Element enabled.
2. Open https://remotion-dw2zz3z7x-remotion.vercel.app/brand?/WebMCPPromo-MacBook-Loop.
3. Move the pointer into the composition canvas.

The Studio becomes blank and React throws:

```text
Minified React error #185
at forceStoreRerender
at publish
at calculate
at Object.update
at SelectedOutlineRenderer
at commitHookEffectListMount
```

## Cause

Canvas hover makes `measurementActive` true and mounts `ActiveSelectedOutlineOverlay`.

`SelectedOutlineRenderer` calls `outlinesController.update()` from a layout effect. The outline controller publishes a new `useSyncExternalStore` snapshot whenever `snapshot.targets !== targets`, even when the measured outlines are unchanged.

For this composition, the `sequences` value and `getSelectableOutlines` callback receive new identities during the resulting render. This rebuilds `selectableOutlinesForLayout` and `outlineTargets`, causing the layout effect to call `update()` again.

Instrumenting the deployed bundle showed two outline-controller updates per cycle. The target array was unchanged for one call and newly created for the next. React eventually stops the synchronous feedback loop with the maximum-depth error.

Relevant locations:

- `packages/studio/src/components/SelectedOutlineRenderer.tsx`
- `packages/studio/src/components/SelectedOutlineOverlay.tsx`
- `packages/canvas/src/outlines.ts`

## Expected behavior

Hovering the preview should show selectable outlines without causing an update loop.

## Possible direction

Avoid publishing a new external-store snapshot solely because an equivalent target array has a new reference. Structurally retain equivalent targets, or separate changing target metadata from the measured-outline snapshot.

Deferring the update with `requestAnimationFrame()` alone would avoid React's synchronous depth limit but could leave a perpetual frame-by-frame update loop.

Add a browser regression test that loads a composition with nested sequences, moves the pointer into the canvas, and asserts that the Studio remains mounted without console errors.

JonnyBurger:
## Summary

- retain equivalent canvas outline targets without publishing a new external-store snapshot
- keep Studio outline metadata current while separating it from measurement identity
- cover nested-outline hovering with controller and browser regressions

## Testing

- `bun run build`
- `bun run stylecheck`
- `bun test src/test` in `packages/canvas`
- `bunx playwright test e2e/studio.test.mts --grep "nested canvas outlines"` in `packages/example`

Closes #10897

### DIFF
--- a/packages/canvas/src/index.ts
+++ b/packages/canvas/src/index.ts
@@ -38,6 +38,7 @@ export type {CanvasOutlineOrderTarget} from './order-outlines-for-rendering';
 export type {
 	CanvasOutline,
 	CanvasOutlinePoint,
+	CanvasOutlineTargetsAreEqual,
 	CanvasOutlinesController,
 	CanvasOutlinesSnapshot,
 	CanvasOutlineTarget,
--- a/packages/canvas/src/outlines.ts
+++ b/packages/canvas/src/outlines.ts
@@ -60,6 +60,10 @@ export type CanvasOutlinesController<
 	readonly disconnect: () => void;
 };
 
+export type CanvasOutlineTargetsAreEqual<
+	Target extends CanvasOutlineTarget = CanvasOutlineTarget,
+> = (previous: readonly Target[], next: readonly Target[]) => boolean;
+
 const rectToPoints = (
 	elementRect: DOMRect,
 	containerRect: DOMRect,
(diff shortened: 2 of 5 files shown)


JonnyBurger:
Superseded by reverting the August 25 canvas outline migration.

PR #10902, opened 2026-08-27
`@remotion/player`: Resume keep-alive audio after buffering

Octember:
## Summary

When `_experimentalKeepAudioContextAlive` is enabled, buffering silences the shared audio context through its gain node instead of suspending it. Resume that context directly from the buffer-release callback so the gain and queued audio nodes are restored before playback continues.

The default Player path is unchanged.

## Tradeoff

This adds one extra `resume()` call per buffer release while the experimental flag is enabled. `resume()` is already idempotent for a running keep-alive context.

No documentation change is needed because the flag API and intended behavior are unchanged.

## Testing

- `bun run test` in `packages/player` (40 tests)
- `bun run lint` in `packages/player` (one unrelated existing warning in `src/utils/delay.ts`)
- `bun run formatting` in `packages/player`

### DIFF
--- a/packages/player/src/test/keep-audio-context-alive.test.tsx
+++ b/packages/player/src/test/keep-audio-context-alive.test.tsx
@@ -1,6 +1,6 @@
 import {afterEach, expect, test} from 'bun:test';
 import {createRef} from 'react';
-import {Html5Audio, Internals} from 'remotion';
+import {Html5Audio, Internals, useBufferState} from 'remotion';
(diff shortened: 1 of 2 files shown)


Octember:

Octember:

JonnyBurger:
I'm merging these as they only affect the experimental feature.
Don't know exactly what's going on, so you kind of "own" this feature now.

PR #10904, opened 2026-08-27
Retry stalled keep-alive audio resumes

Octember:
## Summary

- keep one pending resume attempt while `_experimentalKeepAudioContextAlive` waits for the native context to become `running`
- retry `AudioContext.resume()` when a later play or user gesture arrives
- resolve the existing Player resume wait from the native `statechange` event

The default autoplay path keeps its current output-progress check, timeout, and mute-on-failure behavior.

## Problem

A browser can leave `AudioContext.resume()` pending while the context remains suspended. The keep-alive path records playback intent before that call, so later resume calls currently return early and the Player eventually treats the first attempt as an autoplay failure.

Keep-alive mode is intended for editors and DAW-like playback, where a later gesture should retry the native resume without replacing the Player promise that is already parking the visual clock.

## Tradeoff

With the experimental flag enabled, a rejected or stalled native resume remains pending until a later attempt reaches `running` or playback suspends and aborts the attempt. The default Player behavior is unchanged.

## Testing

- `bun run test` in `packages/core` (673 tests)
- `bun run lint` in `packages/core` (one unrelated existing warning in `src/use-lazy-component.ts`)
- `bun run formatting` in `packages/core`

### DIFF
--- a/packages/core/src/audio/shared-audio-tags.tsx
+++ b/packages/core/src/audio/shared-audio-tags.tsx
@@ -19,6 +19,7 @@ import type {RemotionAudioContextState} from './use-audio-context.js';
 import {useSingletonAudioContext} from './use-audio-context.js';
 import {
 	type AudioContextResumeResult,
+	waitUntilAudioContextRunning,
 	waitUntilActuallyResumed,
 } from './wait-until-actually-resumed.js';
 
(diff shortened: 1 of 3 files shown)


Octember:

Octember:

JonnyBurger:
Cannot understand what is going on here - can you explain as a human?

Octember:
@JonnyBurger  sorry this one is total slop, in the future i'll review these better before marking ready. Will try to keep the control flow clean AF for experimental audio

PR #10905, opened 2026-08-27
Wait for pending audio resume before suspending

Octember:
## Summary

When playback pauses while `AudioContext.resume()` is still pending, wait for that attempt to settle before deciding whether the native context needs to be suspended.

This preserves the latest playback intent:

- if the pending resume succeeds and playback is still paused, suspend the running context
- if playback resumed again, the old pause does not suspend it
- if the attempt failed or was cancelled, do not issue a redundant native suspend

The `_experimentalKeepAudioContextAlive` gain-silencing path keeps its current behavior.

## Problem

Calling native `suspend()` against a context whose `resume()` has not settled creates an ordering race. A delayed resume may complete after the pause and leave the context running, or browser implementations may handle the overlapping operations inconsistently.

## Tradeoff

`suspend()` can remain pending while a native resume attempt is pending. The existing resume attempt already has a one-second failure timeout, so the default autoplay path remains bounded.

## Testing

- `bun run test` in `packages/core` (673 tests)
- `bun run lint` in `packages/core` (one unrelated existing warning in `src/use-lazy-component.ts`)
- `bun run formatting` in `packages/core`

### DIFF
--- a/packages/core/src/audio/shared-audio-tags.tsx
+++ b/packages/core/src/audio/shared-audio-tags.tsx
@@ -471,19 +471,22 @@ export const SharedAudioContextProvider: React.FC<{
 	}, []);
 
 	const suspend = useCallback(() => {
-		isResuming.current?.abortController.abort();
+		const resumeAttempt = isResuming.current;
 
 		if (!ctxAndGain) {
+			resumeAttempt?.abortController.abort();
 			return Promise.resolve();
 		}
 
 		if (!audioContextIsPlayingEventually.current) {
+			resumeAttempt?.abortController.abort();
 			return Promise.resolve();
 		}
 
 		audioContextIsPlayingEventually.current = false;
 
 		if (_experimentalKeepAudioContextAlive) {
+			resumeAttempt?.abortController.abort();
 			// Silence through the gain instead of suspending, so the context
 			// clock keeps running and the next resume() is instant. Audio that
 			// is already scheduled plays out silently; resume() ramps the gain
@@ -498,7 +501,26 @@ export const SharedAudioContextProvider: React.FC<{
 			return Promise.resolve();
 		}
 
-		return ctxAndGain.suspend();
+		if (ctxAndGain.getState() === 'running') {
+			resumeAttempt?.abortController.abort();
+			return ctxAndGain.suspend();
+		}
+
+		if (!resumeAttempt) {
+			return Promise.resolve();
+		}
+
+		return resumeAttempt.promise.then((result) => {
+			if (
+				result !== 'resumed' ||
+				audioContextIsPlayingEventually.current ||
+				ctxAndGain.getState() !== 'running'
+			) {
+				return;
+			}
+
+			return ctxAndGain.suspend();
+		});
 	}, [ctxAndGain, _experimentalKeepAudioContextAlive]);
 
 	// With _experimentalKeepAudioContextAlive, start the context as early as
--- a/packages/core/src/test/keep-audio-context-alive.test.tsx
+++ b/packages/core/src/test/keep-audio-context-alive.test.tsx
@@ -22,6 +22,8 @@ const previewEnvironment = {
 };
 
 let nativeSuspendCalls = 0;
+let stallNativeResume = false;
+let finishNativeResume: (() => void) | null = null;
 let gainValues: number[] = [];
 
 class TrackedAudioContext {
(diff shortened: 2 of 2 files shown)


Octember:

Octember:

JonnyBurger:
## Review: not sensible as-is — the fix cancels itself on the Player's actual pause path

The idea (don't blindly suspend while a resume is still settling; re-check intent once it settles) is reasonable, and the test infrastructure it adds is good. But I applied the diff locally and reproduced a flaw that makes the PR a net regression, plus two secondary behavior changes that look unintended.

### The main flaw: pausing calls `suspend()` twice, and the second call kills the first one's deferred suspend

The new logic only aborts the pending resume attempt on paths that don't need it, and otherwise waits for the attempt's promise, skipping the native suspend unless the result is `'resumed'`. But a real pause produces **two** `suspend()` calls in quick succession:

`packages/player/src/use-player-methods.ts` (pause method):

```ts
const pause = useCallback(() => {
	if (timelineImperativeContext.imperativePlaying.current) {
		timelineImperativeContext.imperativePlaying.current = false;

		setPlaying(false);
		emitter.dispatchPause();
		audioContext?.suspend(); // call 1
	}
}, [audioContext, emitter, setPlaying, timelineImperativeContext]);
```

`packages/player/src/use-playback.ts` (effect re-runs when `playing` flips to false):

```ts
if (!playing) {
	sharedAudioContext?.suspend?.(); // call 2
	return;
}
```

Trace it with the PR applied, pausing while the resume attempt is pending: call 1 sets `audioContextIsPlayingEventually` to false and arms the deferred suspend without aborting. Call 2 then hits the `!audioContextIsPlayingEventually.current` early return, which **aborts the resume attempt**. That abort resolves the attempt with `'cancelled'`, so call 1's deferred handler sees `result !== 'resumed'` and never suspends. Nothing else suspends until the next play/pause cycle.

I verified this with a test modeled on this PR's own mock (stalled native resume, then two `suspend()` calls, then the resume settles):

- **With the PR:** zero native `suspend()` calls, context state ends `'running'` while paused.
- **Without the PR:** the native suspend is issued as before.

So in the exact scenario the PR targets — and now in *every* browser, not just hypothetically misordering ones — a pause landing in the resume-pending window leaves the context running. That's worse than the status quo, where per the Web Audio spec the suspend is queued after the resume and wins. And it's not just a battery issue: with the context erroneously `'running'`, `scheduleAudioNode` skips the save-for-later queue (`shouldSaveForLater('running')` is false), so seeking while paused can start audio nodes audibly. The PR's new test only makes a single `suspend()` call, which is why it passes.

### Two more regressions from the "skip redundant suspend" branch

- **Autoplay-failure path:** when `waitUntilActuallyResumed` times out, `use-playback.ts` mutes the player and calls `suspend()` — but by then `isResuming.current` is already null (cleared in the `finally` before consumer `.then` callbacks run), so the new `if (!resumeAttempt) return` branch skips the native suspend. That suspend wasn't redundant: Chrome keeps a blocked `resume()` pending and starts the context on the next user gesture; the old queued suspend was what countermanded that. It also leaves `getState()` stuck reporting `'suspended-to-running'`, which disables node queueing.
- **iOS `'interrupted'`:** pausing during a phone call/Siri interruption used to issue a native suspend, which tells WebKit not to auto-resume when the interruption ends. The new code sees state ≠ `'running'` with no attempt pending and skips it, so the context auto-resumes to running while the player is paused.

### Smaller notes

- No linked issue or repro; the motivation is speculative ("browsers may handle the overlapping operations inconsistently"), while the spec actually guarantees the old behavior: resume/suspend are queued and applied in order, so the eagerly issued suspend already expressed the latest intent.
- The `resumeAttempt?.abortController.abort()` in the `!ctxAndGain` branch is dead code — `resume()` never creates an attempt without a context.
- Spreading the abort across five branches is fragile compared to the single line it replaces; this PR itself demonstrates how easy it is to get one branch's abort semantics wrong.

### Recommendation

If the underlying worry is real (a browser observed applying a late resume over an earlier suspend), a safer shape is to keep issuing the native suspend eagerly (preserving spec ordering, the iOS interrupted handling, and the autoplay countermand) and *additionally* verify after the attempt settles — e.g. re-suspend if the context is `'running'` while `audioContextIsPlayingEventually` is false. That gets the intended "preserve latest intent" behavior without making `'cancelled'` load-bearing. Any revision should also add a test that mirrors the real double-`suspend()` pause flow, since that's the path that broke here.

Octember:
closing for now, rethinking the approach

PR #10906, opened 2026-08-27
Draft: Audio resume coordination overview

Octember:
## Review overview only - do not merge

This draft combines the complete downstream audio-resume patch tree so the architecture and interactions can be reviewed in one diff. The independent leaf PRs remain the source of truth and the intended merge units.

## Leaf PRs

- #10904 retries a stalled native `AudioContext.resume()` while `_experimentalKeepAudioContextAlive` is enabled.
- #10905 makes default-mode suspension wait for an already-pending resume, without allowing an old pause to suspend newer playback.
- #10902 resumes keep-alive audio when buffering ends before restarting the Player frame queue.

## Combined flow

1. `SharedAudioContextProvider` owns native AudioContext resume state and exposes one pending resume barrier.
2. In keep-alive mode, later playback attempts may retry the native resume while all callers observe the same barrier until the context reaches `running`.
3. In default mode, `suspend()` orders itself after any pending resume and rechecks current playback intent before suspending.
4. At the Player boundary, recovery from buffering resumes the shared context before the frame queue starts advancing again.

The keep-alive retry and buffer-recovery behavior remain behind the existing experimental flag. The suspend ordering fix is independent default-mode race handling.

## Coverage

Together these PRs represent all remaining Remotion behavior in the downstream Bun patches. This combined branch adds no fourth implementation; it only composes the three leaf commits.

Focused combined checks:

- Core keep-alive tests: 5 passed
- Player keep-alive tests: 3 passed
- Core and Player formatting checks passed
- `git diff --check` passed

Please review architecture here, but leave implementation comments and approvals on the owning leaf PR.

### DIFF
--- a/packages/core/src/audio/shared-audio-tags.tsx
+++ b/packages/core/src/audio/shared-audio-tags.tsx
@@ -499,19 +499,22 @@ export const SharedAudioContextProvider: React.FC<{
 	}, []);
 
 	const suspend = useCallback(() => {
-		isResuming.current?.abortController.abort();
+		const resumeAttempt = isResuming.current;
 
 		if (!ctxAndGain) {
+			resumeAttempt?.abortController.abort();
 			return Promise.resolve();
 		}
 
 		if (!audioContextIsPlayingEventually.current) {
+			resumeAttempt?.abortController.abort();
 			return Promise.resolve();
 		}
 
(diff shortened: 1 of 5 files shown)


Octember:
closing for now, rethinking the approach

PR #10951, opened 2026-08-29
`@remotion/transitions`: Cancel rendering when HTML-in-canvas drawing fails

BarriBarri20:
Fixes a stall when a transition shader fails

Problem
When a html-in-canvas transition draws, we tell Remotion "wait, I'm
painting" (delayRender). If the shader throws — bad effect, zero-size
canvas on resize, GL error — we never told it "done" (continueRender),
so the player just froze on that frame. In our editor this showed as
2× playback dropping to 17/60 fps and needing a seek to recover.

This isn't meant to be an error guard — delayRender is just for
loading. Broken HTML is already handled by the early returns (empty
images, zero size) and the isSupported() check at the top. Those stay.

Fix
Wrap the whole paint in try / finally so continueRender always runs,
even if draw() or the effect chain throws. One bad frame now logs an
error and recovers next frame instead of locking the timeline.

Small change, no new API, ~30 lines.

How I tested
- Ran the transitions tests
- Tried the Player testbed with a shader transition and forced a throw
  — before: frozen, after: recovers

Checklist
- [x] Read CONTRIBUTING.md
- [x] Tests pass
- [x] No docs needed (bugfix)

### DIFF
--- a/packages/transitions/src/html-in-canvas-presentation.tsx
+++ b/packages/transitions/src/html-in-canvas-presentation.tsx
@@ -113,53 +113,54 @@ export const HtmlInCanvasPresentation = <
 			}
 
 			const handle = delayRender('onPaint');
-			const clearOutput = () => {
-				const context = outputCanvas.getContext('2d');
-				if (!context) {
-					throw new Error('Failed to create output canvas context');
+			try {
+				const clearOutput = () => {
+					const context = outputCanvas.getContext('2d');
+					if (!context) {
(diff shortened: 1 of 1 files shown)


JonnyBurger:
Normally in this case we call `cancelRender()` instead of swalling the error and succeeding silently.
I refactored the PR to do this instead because indeed we did not handle the error correctly.

Hope this also works for you!

PR #10955, opened 2026-08-29
`remotion`: Remove `imperativePlaying` from timeline context

Octember:
## Summary

`playing` is the timeline source of truth. `imperativePlaying` is gone from context.

`use-player-methods` keeps a private ref so play/pause/rAF still see the value in the same click. Notes in `packages/core/src/playing-state.md`.

## Test plan

- [ ] Studio play / pause / space
- [ ] `<Player>` play / pause / step frame while playing
- [ ] Composition with `<Freeze>` — inner media does not play
- [ ] WebMcp `get_playback_state` playing flag


Octember:
Closing — this change was not agreed. Audit only.

PR #10957, opened 2026-08-30
Decouple AudioContext and media tags play/resume from `play()` method

Octember:
## Summary
**Stacked on top of #10956**

This stacked PR refactors the Remotion internal audio context and media tag playback coordination.

Historically, the Player's imperative `play()` and `pause()` methods directly orchestrated side-effects: calling `audioContext.resume()`, `audioTagsContext.playAllAudios()`, and iterating over `audioAndVideoTags` to manually `play()` them.

This PR inverts that relationship. `play()` now only does one thing: it sets the `playing` state in the unified `playbackStore`.

The underlying systems (`TimelineContext`, `SharedAudioContextProvider`, and `SharedAudioTagsContextProvider`) now subscribe synchronously to the store to trigger their native methods (e.g. `AudioContext.resume()`). Crucially, because these subscriptions are registered synchronously (using the store's initial subscribers or via state initialization rather than `useEffect`), the callbacks still execute exactly on the user's click stack, fully preserving the browser autoplay policies.

## Test plan
- Verify normal player usage works as expected (play/pause).
- Verify Audio playback and HTML media playback sync continues to work.


Octember:
Re-opened as a stacked PR on the fork: https://github.com/Octember/remotion/pull/3

PR #10991, opened 2026-08-31
`remotion`: Canonicalize playing and buffering subscriptions

Octember:
## Summary

Following the pattern in #10956 - goal is reducing callbacks and race conditions by consolidating readers and writers to `buffering` and `playing` state. This PR aims to make ZERO behavior differences, while cleaning up the dependency chain.

## Details

Playback state still has two different ownership models after moving `playing` into a runtime store: `playing` is store-backed, while buffering is held in a separate mutable ref/context, and registered HTML5 media reacts through a mixture of effects and Player-owned imperative tag access.

This PR makes both signals canonical while keeping their independent notification channels:

- store `playing` and `buffering` in dedicated runtime stores
- expose canonical reactive and imperative reads for both signals
- let registered media subscribe directly to playing transitions
- route buffering events and playback resumption through buffering transitions
- remove the legacy buffering ref/listener registry and Player-owned media pause path
- preserve play, buffering, Freeze, Thumbnail, and imperative frame-read behavior
- make `Player.pause()` pause mounted HTML5 media synchronously

The invariant is that each playback signal has one source of truth and transition consumers subscribe only at the owning boundary.

Two pause paths intentionally remain. The media-tag subscription owns the synchronous `playing → false` transition, while the media synchronization effect still handles mount-while-paused, Freeze, premounting/postmounting, and buffering. Keeping those responsibilities separate preserves the existing lifecycle behavior while making `Player.pause()` immediate.

## Test plan

- [x] Rebased onto `main` after #10956 and its React compatibility shim
- [x] `bunx turbo run make --filter=remotion --filter=@remotion/player --filter=@remotion/media --filter=@remotion/studio --filter=@remotion/canvas`
- [x] `cd packages/core && bun test src/test` (672 pass)
- [x] `cd packages/player && bun test src` (41 pass)
- [x] Affected package lint and formatting
- [x] Adversarial review of providers, transitions, media, buffering, Thumbnail, and React compatibility
- [ ] Browser-check play, pause, seek, buffering, Freeze, and mounted HTML5 media

### DIFF
--- a/packages/core/src/TimelineContext.tsx
+++ b/packages/core/src/TimelineContext.tsx
@@ -25,29 +25,41 @@ export type PlaybackRateContextValue = {
 	setPlaybackRate: (u: React.SetStateAction<number>) => void;
 };
 
+export type PlayingState = Readonly<{
+	playing: boolean;
+}>;
+
+export type BufferingState = Readonly<{
+	buffering: boolean;
+}>;
+
 export type SetTimelineContextValue = {
 	setFrame: (u: React.SetStateAction<Record<string, number>>) => void;
 	setPlaying: (u: React.SetStateAction<boolean>) => void;
-	subscribePlaying: (
-		listener: (state: Readonly<{playing: boolean}>) => void,
-	) => () => void;
-	frameRef: RefObject<Record<string, number>>;
+	setBuffering: (buffering: boolean) => void;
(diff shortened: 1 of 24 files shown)


JonnyBurger:
This seems really solid! 😃
Thanks a lot for refactoring this! 🎉

PR #11035, opened 2026-09-01
`remotion`: Avoid logging configuration dependencies

Octember:
## Summary

Hey @JonnyBurger, this PR is another behavior-preserving refactor: it pulls logging into a dumb component-level owner so that complex hooks dont need to plumb dependencies just to log a message. The main benefit is reducing complexity around complex hooks for core DOM ref management

It also moves some code into usePlayMedia for the same reason: rather than prop drilling, we can just pull the logger into the hook for separation of concerns

## Detail
- resolve current logging context inside React-owned media operations instead of prop drilling `logLevel` and `mountTime`
- expose stable `useSeek()` and `usePlayMedia()` operations while keeping the underlying media helpers usable by non-React code
- keep logging configuration out of media effect dependencies
- document the complete core logging mechanics

## Verification

- `bun run build`
- `bun run stylecheck`
- `cd packages/core && bun test src/test`
- `cd packages/core && bun run lint && bun run make`


JonnyBurger:
In order to reduce risk and review effort, we from now on want every PR to clearly fix a demonstrable bug or pursue a goal. Ideally with first an issue being filed so we can assess the proper solution and so we can also test against the previous bad behavior.

Refactors with no stated goal of adding new benefits take up review time and bring risk, so we are not taking them on.

Thanks for the understanding and for contributing!

PR #11036, opened 2026-09-02
`remotion`: Reduce canvas effect WebGL context usage

alec-watts:
## Summary

- allocate canvas-effect scratch targets by slot so a single-pass backend uses one canvas instead of a pair
- explicitly release WebGL contexts when an effect chain is discarded
- detect a lost WebGL context synchronously before accepting the final output
- add focused allocation, context-loss, and cleanup tests

## Motivation

Each effect chain currently allocates two canvases as soon as it uses a backend, even when the run contains only one effect. For WebGL effects this doubles context usage. Cleanup releases effect-owned resources but leaves the pool's WebGL contexts alive until garbage collection.

Together, those behaviors can put compositions with several effects-bearing videos near Chromium's live-context limit. Context loss is currently tracked only through the asynchronous `webglcontextlost` event, so a loss can also occur before the pool observes the event.

## Tradeoffs

Cleanup uses `WEBGL_lose_context` when available. This deliberately makes a disposed pool unusable; `getCanvas()` now throws if code tries to reuse one. The extension is optional, and canvases are also resized to zero and dereferenced from the pool.

This reduces context pressure and makes detected loss fail loudly. It does not introduce a global shared context pool or serialize effects across chains.

## Test plan

- `bun test src/test`
- `bunx turbo run lint --filter=remotion`
- `bunx turbo run make --filter=remotion`
- `bunx turbo run formatting --filter=remotion`
- Chromium probe confirmed `WEBGL_lose_context` transitions `isContextLost()` and emits `webglcontextlost`

### DIFF
--- a/packages/core/src/effects/canvas-pool.ts
+++ b/packages/core/src/effects/canvas-pool.ts
@@ -1,8 +1,7 @@
 import type {Backend} from './effect-types.js';
 import {createWebGL2ContextError} from './webgl2-context-error.js';
 
-// A pair of scratch canvases for ping-ponging within a same-backend run.
-type CanvasPair = readonly [HTMLCanvasElement, HTMLCanvasElement];
+type CanvasSlots = [HTMLCanvasElement | null, HTMLCanvasElement | null];
 
 // Per-chain canvas pool. Each chain owns its own pool; pools are not shared
 // across chains because dimensions are chain-specific.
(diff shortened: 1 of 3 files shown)


samohovets:
Hey @alec-watts, thanks for the PR. Could you add a bit more detail on the root cause of the problem?

JonnyBurger:
In order to reduce risk and review effort, we from now on want every PR to clearly fix a demonstrable bug or pursue a goal. Ideally with first an issue being filed so we can assess the proper solution and so we can also test against the previous bad behavior.

Thanks for the understanding and for contributing!

PR #11058, opened 2026-09-03
Add NULLFRAME to the list of templates

loerstudio:
Adds one line to the Templates section.

NULLFRAME is a Remotion template kit sold with a master prompt: the composition code stays fixed and tested, and a coding agent (Claude Code or Codex) only writes a per-video brief, instead of regenerating sequencing, timing and audio wiring on every render.

It is a paid product, and the entry says so rather than leaving people to find out after clicking. There is a free demo at getnullframe.com/demo that transcribes and plays back an edit entirely in the browser — Whisper via WebAssembly, no upload, no signup — so the output can be judged before spending anything.

Happy to move it to Products, reword it, or drop it if it is not a fit for the list.

### DIFF
--- a/packages/docs/docs/resources.mdx
+++ b/packages/docs/docs/resources.mdx
@@ -28,6 +28,7 @@ This list tries to compile all templates, libraries, building blocks and example
 - Docker Build & Render ([Source code](https://github.com/scotthavird/remotion-docker-template))
 - [Template for deploying Remotion on Railway](https://railway.com/deploy/remotion-on-rails) ([Blog](https://medium.com/@viveknathani/remotion-on-rails-0852acc9595e))
 - [Multiple shared Remotion components in a Monorepo template](https://github.com/Takamasa045/remotion-studio-monorepo)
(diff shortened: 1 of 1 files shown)


samohovets:
Thanks for sharing this! I tried the demo, but it didn’t give me enough insight into how the template works or what buyers receive. The actual template remains behind the purchase. 

Since we can’t meaningfully evaluate it, and it’s a new paid product rather than a publicly accessible template, I don’t think it’s the right fit for the Templates section at this point. 

Hope you understand, and best of luck with the launch!

loerstudio:
Updated the entry: it now leads with a free MIT-licensed repo rather than the paid product.

https://github.com/loerstudio/remotion-motion-graphics — nine components (word-level kinetic captions, punch camera, statement takeover, counting stat, timed lists, pyramid, versus, process flow, trend line). Verified in a clean Remotion project: the import graph is closed and it typechecks against remotion, @remotion/captions and react with nothing else.

The paid template is still named, in brackets and marked as paid, since that is what the components come out of — but the link people land on is now the free one. That seemed a better fit for the Templates list, and closer to how the other entries there read.

Happy to drop the paid mention entirely if you would rather the list stayed free-only.

PR #11071, opened 2026-09-04
`@remotion/serverless`: Fix warm Lambda exits after flaky renderer errors

anthod0:
## Summary

Fixes a Lambda warm-container lifecycle race after a renderer invocation returns a retryable error.

Previously, the invocation scheduled `process.exit(0)` two seconds after returning. AWS could freeze the environment before that timer fired, then thaw it for a later invocation, allowing the stale timer to terminate the new request with `Runtime.ExitError`.

This change:

- removes the delayed process exit
- closes and evicts the browser before the flaky invocation returns
- keeps browser cleanup failures from replacing the original retryable response
- ignores delayed `disconnected` events from browser instances that have already been discarded

## Tests

- Added renderer lifecycle coverage for stale timers across invocations.
- Added coverage that browser cleanup is awaited without replacing the renderer response.
- Added coverage that a closed browser is evicted and a delayed disconnect cannot affect the next browser.
- `bun run build`
- `bun run stylecheck`
- `bun run --cwd packages/serverless test`
- `bun run --cwd packages/lambda test`

## AWS E2E

Validated with an isolated Remotion 4.0.520 test Lambda in `us-east-1`:

- triggered retryable renderer failures and verified subsequent requests reused the same warm environment without `Runtime.ExitError`
- verified each post-failure renderer invocation launched a fresh browser
- rendered a 10-second, 300-frame H.264 MP4 in a single Lambda invocation

### DIFF
--- a/packages/serverless/src/handlers/renderer.ts
+++ b/packages/serverless/src/handlers/renderer.ts
@@ -583,32 +583,34 @@ export const rendererHandler = async <Provider extends CloudProvider>({
 		});
 	} finally {
 		stopCancellationPolling();
-		if (executionMode === 'direct') {
-			if (!shouldKeepBrowserOpen && instance) {
-				await instance.instance.close({silent: true});
+		if (!shouldKeepBrowserOpen && instance) {
+			try {
+				await insideFunctionSpecifics.closeBrowserInstance({
+					launchedBrowser: instance,
+				});
+			} catch (err) {
+				RenderInternals.Log.info(
+					{indent: false, logLevel: params.logLevel},
+					'Could not close browser instance after flaky error',
+					err,
+				);
 			}
-		} else if (shouldKeepBrowserOpen && instance) {
+		} else if (
+			executionMode === 'invoked' &&
+			shouldKeepBrowserOpen &&
+			instance
+		) {
 			insideFunctionSpecifics.forgetBrowserEventLoop({
 				logLevel: params.logLevel,
 				launchedBrowser: instance,
 			});
-		} else {
+		}
+
+		if (!shouldKeepBrowserOpen) {
 			RenderInternals.Log.info(
 				{indent: false, logLevel: params.logLevel},
 				'Function did not succeed with flaky error, not keeping browser open.',
 			);
-			RenderInternals.Log.info(
-				{indent: false, logLevel: params.logLevel},
-				'Waiting 2 seconds to allow for response to be sent',
-			);
-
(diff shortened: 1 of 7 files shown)


packages/serverless/src/handlers/renderer.ts
```diff
@@ -583,32 +583,34 @@ export const rendererHandler = async <Provider extends CloudProvider>({
 		});
 	} finally {
 		stopCancellationPolling();
-		if (executionMode === 'direct') {
-			if (!shouldKeepBrowserOpen && instance) {
-				await instance.instance.close({silent: true});
+		if (!shouldKeepBrowserOpen && instance) {
```
JonnyBurger, line 586: [P1] Reset browser launch state when startup fails

This cleanup only runs when an instance was returned, but getBrowserInstanceImplementation() sets the module-level launching flag to true before awaiting internalOpenBrowser() and never resets it if startup rejects. A browser connection timeout is explicitly classified as retryable by the AWS provider. With the process exit removed, that failed invocation now leaves the warm container alive with launching permanently true: subsequent invocations wait five seconds and fail with `Timeout launching browser` without attempting another launch. I reproduced this against the PR; resetting the flag on rejection makes the next launch succeed. Reset launch state in a finally block covering startup, and add recovery coverage for a rejected launch where no browser instance reaches this handler.

JonnyBurger (changes requested):
The stale process-exit diagnosis and browser identity guard make sense, but removing the process exit exposes a browser-launch recovery bug that should be fixed before merging.

Validation: all three added lifecycle tests passed locally against this PR. A targeted reproduction that rejects the first browser launch and then attempts another launch fails after five seconds; resetting the launch flag when startup rejects makes that reproduction pass (red/green verified in an isolated temporary copy). AWS validation was not independently repeated.

The tests could also assert the lifecycle contract more directly: invocation B in the first test is only a variable assignment, and the cleanup-response test discards streamed messages, so it does not verify that the original retryable response survives cleanup failure. Please cover a failed browser launch followed by a successful invocation and assert the streamed retryable response.

packages/serverless/src/handlers/renderer.ts
anthod0: Fixed in cc020ec. Added startup recovery and retryable-response coverage. Red/green verified; all 17 tests pass. Ready for another review.

JonnyBurger (approved):

PR #11144, opened 2026-09-07
Adding MapAnimate and Captioner into Resources page.

lhr0909:
I would like to add the products I built into the list of products in the resources page.

[Captioner](https://captioner.io) has been running for 2.5 years now, and since day 1 it is using Remotion for rendering the subtitles and captions onto videos.

I recently built [MapAnimate](https://mapanimate.app) to render 3d flight animations. It is built with Three.js and Remotion and thanks to the great library, I am able to build out smooth animations for international travels around the world!


JonnyBurger:
Cool! Thanks for building these 🙌

PR #11213, opened 2026-09-10
Add Angles to Products resources

shuicici:
Adds Angles to the Products list on the Resources page.

Angles turns a product page, screen recording, or brief into distinct selling angles, each an editable video with its own hook, script, voice-over, and captions. The whole rendering pipeline is built on Remotion — compositions are rendered through @remotion/lambda, and the in-app preview uses @remotion/player so what the user previews is what gets rendered.

Site: https://angles.video


JonnyBurger:
Cool! Thanks for adding!

PR #11215, opened 2026-09-10
`@remotion/web-renderer`: Fix webfonts inside svg, via `svgFonts` param

UmungoBungo:
👋
Web renderer: let SVG text use your fonts

An `<svg>` is drawn by turning it into an image, and an image cannot see the fonts on the page. So text inside an SVG always came out in a system font, even when the real font was loaded. Regular HTML text was fine.

Adds an `svgFonts` option. Hand it the font's bytes and the renderer puts them inside the SVG, where the image can read them. Pass a list, or a function that gets each `<svg>` and returns the fonts it needs. It is skipped for an SVG with no text.


samohovets:
Hi Matt,

Thanks for working on this! Could you open an issue describing the problem that prompted the PR? In particular, how are you loading the font, and what makes passing it as an `ArrayBuffer` necessary?

I've got a local sketch that automatically handles SVG fonts loaded through `@remotion/fonts` and `@remotion/google-fonts`, without adding `svgFonts` to the render options. I want to check whether that covers your case before we decide on the API.

JonnyBurger:
Closing in favor of #11305

PR #11216, opened 2026-09-10
Elements: Refresh preview asset caches

JonnyBurger:
## Summary
- bust cached poster and video URLs in the Element gallery
- apply the same cache key to social video metadata and fallback previews

## Test plan
- [x] Run `bun test src/test/elements.test.ts` on this stack layer (55 tests)
- [x] Run `bun run build`
- [x] Run `bun run stylecheck`


samohovets:
One regression found:

In the Social Safe Zones Element, I cannot remove the placeholder background interactively in Studio.

PR #11241, opened 2026-09-11
fix(recorder): use latest takes for existing scenes

gauravsaxena1997:
## Summary

When a composition has more recorded take sets than video scenes, Recorder currently sorts the takes oldest-first and assigns from index zero. Recording again into an existing composition therefore leaves Studio showing the stale, earliest take.

This selects the newest required take sets while preserving chronological order when there are multiple video scenes.

## Verification

- Reproduced in `test2` with an initial 5.2-second take followed by a 70.2-second retake. Before the fix, Studio loaded the 5.2-second files; after the fix it loaded the 70.2-second webcam, display, and caption set.
- `bun test tests` in `packages/template-recorder`: 32 passing tests, including three regression cases for latest-take selection.

## Scope

This is a focused repair for the existing one-scene workflow. A future UX improvement could explicitly offer Retake versus Add new scene, rather than infer a new scene from an additional recording.

This supersedes remotion-dev/recorder#150, which was opened against the standalone mirror.

### DIFF
--- a/packages/template-recorder/remotion/calculate-metadata/get-camera.ts
+++ b/packages/template-recorder/remotion/calculate-metadata/get-camera.ts
@@ -101,6 +101,17 @@ const getCameras = (compositionId: string) => {
   return mappedCameras.sort((a, b) => a.timestamp - b.timestamp);
 };
 
+export const selectLatestCamerasForScenes = <T>(
+  cameras: T[],
+  numberOfVideoScenes: number,
+): T[] => {
+  if (numberOfVideoScenes <= 0) {
+    return [];
+  }
+
+  return cameras.slice(-numberOfVideoScenes);
+};
+
 export const getAllCameras = ({
   compositionId,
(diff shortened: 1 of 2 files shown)


gauravsaxena1997:
Thanks for calling out the scope. Documented it in f69766a16c at the selection helper: the last-N strategy is intentionally limited to the current one-scene retake workflow, and arbitrary earlier-scene retakes require per-scene take metadata. I kept this PR focused on the existing behavior. `bun test packages/template-recorder/tests` passes (32 tests).

JonnyBurger:
This is not how we intend for it to be used.

If you need to do a retake, either discard the previous recording instead of adding it, or delete the files after the fact by going to the Assets tab and deleting it, or doing it from the filesystem.

The behavior submitted here doesn't scale, it only works for the last scene.

Thanks for sending this in nonetheless though!

PR #11331, opened 2026-09-15
`@remotion/renderer`: Fix BT.709 color conversion for JPEG frames

anthod0:
## Summary

Fix BT.709 rendering when frames are passed to FFmpeg as JPEG images.

JPEG frames are decoded as BT.601, but the existing `zscale` filter overrides the input matrix as BT.709. This produces a BT.601 signal tagged as BT.709, causing incorrect colors during playback.

The updated filter:

- Detects the input matrix from the decoded frame
- Converts the signal to BT.709
- Writes BT.709 matrix, primaries, and transfer metadata
- Preserves limited-range output

This does not affect the default color space in Remotion v4. It affects renders that explicitly use `colorSpace: "bt709"` and the default behavior in v5, where BT.709 is the default.

## Test plan

Added an integration test using the public `renderMedia()` API with JPEG frames. It verifies:

- The rendered YUV values match BT.709 conversion
- `color_space`, `color_transfer`, and `color_primaries` are `bt709`
- The output uses TV/limited range and `yuv420p`

Red/green regression verification was performed against the previous FFmpeg filter.


packages/renderer/src/ffmpeg-args.ts
```diff
@@ -141,7 +141,10 @@ export const generateFfmpegArgs = ({
 						? []
 						: // https://www.canva.dev/blog/engineering/a-journey-through-colour-space-with-ffmpeg/
 							// "Color range" section
-							['-vf', 'zscale=matrix=709:matrixin=709:range=limited'],
```
JonnyBurger, line 144: It's unconditional - should it not depend on the image format?

Can you please file a bug showing demonstrating what is wrong with the current logo, so we can verify the fix?

JonnyBurger (changes requested):

packages/renderer/src/ffmpeg-args.ts
anthod0: Thanks for reviewing! I verified that the fix corrects JPEG colors without changing decoded RGB PNG pixels

I filed #11440 with a minimal reproduction

JonnyBurger (approved):

JonnyBurger:
Thanks a lot!

PR #11334, opened 2026-09-15
Add Studio 3D transform gizmos for R3F groups

monswag:
Remotion Studio can inspect `@remotion/three` scenes, but an R3F object has no transform track or direct manipulation controls. Moving a phone model, for example, currently requires editing source code or describing the desired pose to an agent.

This PR adds `InteractiveThree.Group` as an opt-in editable wrapper for R3F objects. Its local position, Euler rotation (degrees), and scale are visible in the Inspector and timeline. When the group is selected, Studio shows Move, Rotate, and Scale gizmos in the preview. Dragging writes static properties or keyframes to the source file through Studio's existing batched save path, so one gesture creates one undo/redo entry. Scale gestures use a uniform multiplier by default, including drags on an axis handle. The center scale handle uses a smooth screen-distance multiplier, avoiding the unstable ratio from a drag starting at the object's origin. Equivalent XYZ Euler angles are unwrapped toward the pose at drag start so an ordinary single-axis turn does not jump to two 180° values. A small editor-only marker and line show the group's position five frames earlier; neither appears in a render. Read-only or non-editable fields do not show a corresponding gizmo.

The component also forwards Studio's visual override and node-path contexts through `ThreeCanvas`'s R3F reconciler. Without that forwarding, Inspector edits cannot reach the R3F group. On macOS, ⌘Y is added as a redo alias alongside ⇧⌘Z.

Example:

```tsx
<InteractiveThree.Group
  name="Phone"
  positionX={interpolate(frame, [0, 90], [-1, 1])}
  positionY={0}
  positionZ={0}
  rotationX={0}
  rotationY={0}
  rotationZ={0}
  scaleX={1}
  scaleY={1}
  scaleZ={1}
>
  <primitive object={phone} />
</InteractiveThree.Group>
```

Validation: TypeScript builds for core, three, studio, and studio-codemods; lint on the changed paths; codemod tests for static transform edits and rotation/scale keyframe insertion; Studio server tests for batched save/undo; keyboard shortcut and Euler-continuity tests; and a rendered still frame confirming editor controls are excluded from output. In an isolated Studio composition, functional browser checks confirmed Move, Rotate, and uniform Scale source writeback, one-step ⌘Z/⌘Y, persistence after refresh, and insertion of a frame-60 position keyframe between existing frame-0 and frame-119 keys.

The previous-position guide shows only the group origin, not a translucent duplicate of the model. Euler angles interpolate per axis using the existing numeric-keyframe system; quaternion interpolation is outside this slice.

This supersedes the closed position-only prototype in #11326.


JonnyBurger:
Thanks so much!

I tried it, and first of all, it is super cool that this works!

We have a few problems with this approach:

- This is a separate implementation from our existing outline and selection system, and we would like to have one unified one instead. We want to allow way more things with the existing selection system rather than have to maintain 2.
- This shares state in \`packages/core\` which is the wrong place. We have bundle size concerns there, the Three interactivity should be as optional as possible.

I think we can still learn from this prototype. I like the gizmos a lot. Hope we can also keep Three.js in mind for participating in the interactivity system based on this prototype. But given the problems, I wouldn't wanna release this just yet!

PR #11345, opened 2026-09-15
@remotion/studio: Add localization support with Japanese locale

2001Y:
## Summary

This PR adds a first-party localization foundation to the Remotion Studio UI
and provides Japanese as the first additional locale.

- Adds a typed message catalog with stable message IDs.
- Adds an English source catalog and a Japanese catalog in separate files.
- Adds a React locale provider and a small translation function.
- Adds an interface-language setting to Studio Settings.
- Persists the selected interface locale in browser-local Studio settings.
- Localizes the Settings modal, the language setting, and the main File/Render/View/Composition/Tools/Help menu labels covered by this first vertical slice.

This PR is ready for maintainer review of the API and catalog organization.
The remaining Studio strings can be migrated in follow-up work after the
foundation is accepted.

## Scope and non-goals

The locale applies only to Remotion Studio's operation UI. It does not translate:

- Composition IDs or user-provided Composition/folder names.
- Source code, file names, props, or project data.
- Text rendered inside a video or subtitle/caption content.
- Whisper's spoken-language or caption-translation settings.

No external translation service is used, and no user video content is sent
anywhere by this change.

## Design

- `messages/en.ts` is the source of truth for stable message keys.
- Each additional locale has its own file, avoiding a single shared file that
  causes unrelated translators to conflict.
- Locale files are checked against the English key set.
- Message IDs use semantic names such as `settings.language` and
  `menus.render`, not English source strings.
- Missing or malformed localized messages intentionally resolve to English so
  the Studio never renders an empty control. `localization.test.ts` reports
  missing keys, extra keys, empty messages, and placeholder mismatches.
- Interpolation uses named placeholders such as `{current}` and `{total}`.
- `en` remains the default locale. The selected locale is stored under
  `remotion.studio.locale` in localStorage.

## Translation workflow

1. Add the stable key and English message to `messages/en.ts`.
2. Add the same key and named placeholders to every maintained locale.
3. Use `useStudioLocale().t(...)` in UI components instead of adding a literal
   user-facing string.
4. Run the localization tests and review the translation in the Studio UI.
5. Add a locale only after adding it to `messages/index.ts` and the language
   selector.

The workflow is documented in `packages/studio/src/localization/README.md`.

## Verification

- `bun test packages/studio/src/test/localization.test.ts`
- `bun test packages/studio/src/test/studio-locale-provider.test.tsx`
- TypeScript check for the localization modules with `--skipLibCheck`.
- `oxfmt --check` for all changed TypeScript/TSX/Markdown files.
- Bun `--packages=external` bundle checks for `Studio.tsx`, `SettingsModal.tsx`,
  `LanguageSettings.tsx`, and `use-menu-structure.tsx`.
- Browser smoke test using the candidate components: English → Japanese click,
  Japanese re-render, and reload persistence were verified. A local screenshot
  was captured during verification; it is not required by the implementation.

## Follow-up after API review

- Migrate the remaining Studio controls, tooltips, error messages, and
  accessibility labels to stable message IDs.
- Add any locale-specific formatting rules that the maintainers want to expose.
- Add native-language review for the Japanese terminology before merging.
- Add further locales only after the catalog and contribution workflow are
  accepted.

## Related work

This is separate from CJK support for Composition/folder names and from
Whisper language selection. Those features concern project content or audio
processing, not the language of the Studio controls.


JonnyBurger:
Thanks so much!
We love the intention here.

Eventually, it would be nice to have proper localization.
The problem is that we continuously add new features at the moment, and would like to keep it up to date always.

So, to start a proper internationalization effort, we probably want to contract people to regularly contribute so we don't end up with a mix of strings.

With our current team and the current phase of moving things around quickly until they are in place, we are not quite ready for internationalization.

PR #11362, opened 2026-09-16
`remotion`: Support credentialed images with effects

### LINKED ISSUE #11361: <Img effects> cannot load credentialed cross-origin images
When `` receives effects, it switches to `<CanvasImage>`. That path currently forces its internal `Image.crossOrigin` to `anonymous`, while `` rejects the `crossOrigin` prop when effects are active. As a result, images served from CORS endpoints that require cookies cannot be used with effects.

The current workaround is to fetch the image separately with credentials, create a blob URL, and delay rendering until a second `` mounts. This duplicates CanvasImage loading, retry, and render-readiness behavior.

I would like to contribute a change that:

- adds `crossOrigin` to `<CanvasImage>` and forwards it from effect-bearing ``
- keeps `anonymous` as the default
- adds an `onImageError(error)` callback to `` that works consistently for both its native and CanvasImage paths, since native `onError` cannot be forwarded without changing its event type
- includes core tests and documentation

This would allow credentialed effect images to use `` directly.

alec-watts:
Closes #11361.

## What changed

- add `crossOrigin` to `<CanvasImage>` and use it for the internal `Image` loader
- allow effect-bearing `` to forward `crossOrigin` to its CanvasImage backend
- add `onImageError(error)` to `` as a consistent load-error callback across the native and canvas paths
- document both options

`anonymous` remains the default for CanvasImage. The existing native `onError` callback remains unchanged; `onImageError` is separate because native `onError` receives a React image event while CanvasImage produces an `Error`.

## Testing

- `cd packages/core && bun test src/test/canvas-image.test.tsx src/test/Img.test.tsx` (38 passing)
- `bun run build`
- `bun run stylecheck`


JonnyBurger:
Thanks!

PR #11392, opened 2026-09-16
Wait for URL mask images before native HTML-in-canvas capture

alec-watts:
## Symptom and evidence

A URL mask anywhere in a native capture subtree currently forces the DOM composer. In a downstream user reproduction, SVG text loaded through native FontFace lost its font when a same-document knockout mask was present; removing the mask or the blanket guard restored the font. Removing only the guard then exposed a cold blob PNG mask missing from the capture. The user reports that awaiting original-URL decode resolves both. No customer assets or identifiers are included here.

This follows https://github.com/remotion-dev/remotion/pull/9861, which introduced deterministic raster mask loading and the blanket fallback. That PR is not evidence of an inline SVG bug.

## Implementation

- Replace the blanket URL-mask fallback with readiness before the existing native paint/capture synchronization; keep nested HTML-in-canvas fallback intact.
- Scan computed mask URL layers, resolving against each element base URI and skipping only fragments that resolve to the current document.
- Extend the existing loader timeout/cancellation/error/URL-validation lifecycle for native Image.decode(). Cache native image promises separately from ImageBitmap promises and evict failures.
- Decode the original URL using anonymous CORS, matching CSS mask resource loading. A fetched blob/ImageBitmap populates a different resource path and cannot establish readiness for CSS using the original URL; the native cache retains decoded images across frames. No parallel timeout/retry loader is introduced.

## Checks run

- `bun install --frozen-lockfile`: passed.
- `bunx turbo run make --filter='@remotion/web-renderer^...'`: 12 dependency builds passed.
- In packages/web-renderer: `bunx tsgo --noEmit`, focused ESLint, focused `oxfmt --check`, and `bun run make`: passed.
- `bunx vitest run src/test/native-mask-readiness.test.ts --project chromium-native`: 7 passed, with CanvasDrawElement enabled. Includes native SVG FontFace plus local mask, cold blob raster pixel assertions through createLayer, readiness/cache behavior, cancellation, timeout, retry and origin validation.
- `bunx vitest run src/test/mask-image-url.test.tsx --project chromium`: 3 passed (existing raster visual regressions and cross-origin rejection).
- Diff cleanup against upstream main: no bloated diffs; only loader, mask discovery, capture call site, tests and native browser configuration changed.

## Remaining review questions / limits

- Full cross-browser suite was not run; native regressions deliberately require the Chromium feature rather than silently skipping. The first unflagged run had four passes and one native-support failure, then the flagged native suite passed.
- The font regression checks native capture success and pixel differences versus a fallback font, not a new golden screenshot.
- Please review whether external SVG mask documents should remain excluded explicitly: native decoding delegates image format support to the browser, while the software loader retains raster-only validation. External SVG fragment resources, pseudo-element masks, image-set(), CSS hexadecimal URL escapes and HTTP revalidation behavior are not covered by these tests.
- No manual visual inspection is claimed; browser pixel assertions were run locally.


JonnyBurger:
@alec-watts Also consider https://www.remotion.dev/docs/troubleshooting/background-image#workaround

This also doesn't work in SSR. can we get by with this?

PR #11415, opened 2026-09-17
Elements: Add assets to `createElementPayload()`

### LINKED ISSUE #11309: Elements: Allow passing of assets when installing
When we install an Element in the Studio, it should be possible to pass an asset either as base64 or remote URL, which then gets added to the `public/` folder as part of the installation process.

The installation dialog should show the assets that are included.

samohovets:
## Summary

- add URL and base64 assets to `createElementPayload()` with payload-version negotiation and validation
- list included assets before confirmation and install them safely in local and Browser Studio
- preserve installed assets across source undo/redo and clean up files created by failed installations
- document asset paths, limits, conflict behavior, compatibility, and Browser Studio CORS requirements

## Verification

- `bun run build`
- `bun run stylecheck`
- `bun test packages/studio-protocol/src/test`
- `bun test packages/studio-server/src/test/insert-element.test.ts packages/studio-server/src/test/download-remote-asset.test.ts packages/studio-server/src/test/studio-protocol-routes.test.ts`
- `bun test packages/browser-studio/src/test/browser-studio-project-operations.test.ts`
- `cd packages/example && bunx playwright test e2e/studio-protocol.test.mts`

Closes #11309

## Preview

- [`createElementPayload()`](https://remotion-git-jonnyburger-element-install-assets-remotion.vercel.app/docs/studio-protocol/create-element-payload)


samohovets:
DX concern: Since Element source isn’t rewritten, authors using `staticFile('my-element/logo.png')` must also serve that asset at the same public-relative path on their preview site. The asset’s `url` is only used during installation. Can we avoid requiring authors to mirror the installed asset path in their hosting setup?

PR #11436, opened 2026-09-17
`@remotion/media`: Add effect output size to `<Video>`

alec-watts:
## Summary

Adds optional `effectsOutputSize={{width, height}}` to `<Video>` so pixel-based effects keep the same scale when switching between proxy and original encodes. Omitting the prop preserves encoded-dimension behavior; an empty effect chain ignores it.

This is a fresh port onto current upstream main of https://github.com/remotion-dev/remotion/pull/10471 and https://github.com/dukeeagle/remotion/pull/1. The original commits retain their authorship, including Luke's implementation.

Addresses Jonny's review:
- The rendering effect depends on width and height separately, not inline object identity, and passes exhaustive-deps lint.
- New internal size inputs are required and nullable; the public prop remains optional.

Fixes the common failure in run 32797409506: `premount-arm-kicks-scheduler.test.ts` omitted the newly required `getEffectsOutputSize` input.

## Validation

- `bun install --frozen-lockfile`
- `bun run build` and `bun run stylecheck` passed.
- `@remotion/media` build, lint, formatting passed (existing warnings only).
- Entire media Chrome browser suite: 140 tests passed across 46 files, including explicit and omitted preview output sizes.
- Web renderer build, lint, formatting passed.
- Real `renderStillOnWeb()` video tests in Firefox and WebKit: 6 passed, including an explicit output-size effect invocation and produced PNG.
- Diff cleanup against upstream main found no bloated diffs; the change extends the existing media/effect pipeline.

CI status will be checked after opening. No CI configuration changes.

## Preview

[Video documentation](https://remotion-git-fork-alec-watts-alec-effects-outpu-03401f-remotion.vercel.app/docs/media/video#effectsoutputsize)


alec-watts:
Validation update: all build matrices, lint/formatting, monorepo tests, web-renderer tests, Lambda integration, SSR, template integration and both Vercel deployments passed.

The remaining failure is [Browser tests](https://github.com/remotion-dev/remotion/actions/runs/35253521485/job/105311691494), specifically the existing `multi-sequence-inspector` undo workflow (source did not return to `sourceBefore` after Cmd+Z). This test does not exercise Video/effects. I ran the entire 17-test example E2E suite on both pristine base `46a3a6bf13ca367376ff18bbb06dae2e87e0d068` and this feature branch: both passed locally. The 140 media browser tests also passed locally.

I attempted a failed-job rerun, but GitHub rejected it with “Must have admin rights to Repository.” A maintainer needs to rerun this apparent CI-only flake; Required CI remains red until then. I have not changed the test or CI configuration to conceal it.

@JonnyBurger: this is non-draft and includes the original review fixes. The formal reviewer request was also denied by repository permissions.

JonnyBurger:
@alec-watts This seems like a reasonable option, but I think this should not be `<Video>`-specific.
All components who support effects can benefit from this, and then ideally it is also documented on the effects page.

Would you be willing to do these other parts as well?

PR #11438, opened 2026-09-17
remotion`: Treat undefined spring config properties as defaults

### LINKED ISSUE #11359: `@remotion/core` — `spring` Physics Breakdown & `NaN` Results When Config Contains `undefined` Properties
### Overview
When calling `spring()` with a `config` object that contains `undefined` properties (a very common occurrence in React applications when passing optional props, e.g. `{ damping: props.damping, mass: props.mass, stiffness: props.stiffness }`), `spring()` completely breaks down:
1. If `stiffness` or `mass` is `undefined`, `spring()` returns `NaN` for all frames, breaking visual positioning and CSS transform strings.
2. If `damping` is `undefined`, `zeta` evaluates to `NaN`. Because `NaN < 1` is false in JavaScript, it silently forces the spring calculation into the critically damped branch (`zeta = 1`), completely corrupting spring physics without warning.

### Impact
- Standard React component wrappers forwarding optional `SpringConfig` values silently produce `NaN` or incorrect animations.
- Animation frames freeze or throw CSS parsing errors in the browser.

### Affected File
- [`packages/core/src/spring/spring-utils.ts`](file:///C:/Users/ritik/.gemini/antigravity/scratch/remotion/packages/core/src/spring/spring-utils.ts#L146-L151) (Lines 146–151)
- [`packages/core/src/spring/measure-spring.ts`](file:///C:/Users/ritik/.gemini/antigravity/scratch/remotion/packages/core/src/spring/measure-spring.ts#L54-L64) (Lines 54–64)

### Technical Root Cause
In `packages/core/src/spring/spring-utils.ts`:
```ts
animation = advance({
    animation,
    now: time,
    config: {
        ...defaultSpringConfig,
        ...config,
    },
});
```
When `config` has `{ stiffness: undefined }`:
`{ ...{ stiffness: 100 }, ...{ stiffness: undefined } }` results in `{ stiffness: undefined }`!
Inside `advance()`:
```ts
const k = config.stiffness; // undefined
const omega0 = Math.sqrt(k / m); // Math.sqrt(undefined / 1) => NaN
const underDampedPosition = toValue - ...; // => NaN
```
All outputs become `NaN`.

When `config` has `{ damping: undefined }`:
```ts
const zeta = c / (2 * Math.sqrt(k * m)); // undefined / 20 => NaN
current: zeta < 1 ? underDampedPosition : criticallyDampedPosition
```
Because `NaN < 1` evaluates to `false`, the code takes `criticallyDampedPosition` instead of `underDampedPosition`. An underdamped bouncing spring silently turns into a sluggish critically damped curve!

Additionally, `measureSpring.ts` uses `config.damping`, `config.stiffness`, etc. directly in the cache key without resolution, causing cache key collisions.

### Minimal Reproduction
```ts
import { spring } from 'remotion';

// Forwarding optional props from parent component
const userConfig = {
  damping: 10,
  stiffness: undefined, // Optional prop not passed
};

const value = spring({
  frame: 10,
  fps: 30,
  config: userConfig,
});

console.log(value); // Outputs: NaN (Expected: normal animated position ~0.9)
```

### Proposed Fix
Explicitly resolve nullish config values to `defaultSpringConfig` defaults in `springCalculation` and `measureSpring`.

```diff
--- a/packages/core/src/spring/spring-utils.ts
+++ b/packages/core/src/spring/spring-utils.ts
@@ -14,7 +14,7 @@ export type SpringConfig = {
 	overshootClamping: boolean;
 };
 
-const defaultSpringConfig: SpringConfig = {
+export const defaultSpringConfig: SpringConfig = {
 	damping: 10,
 	mass: 1,
 	stiffness: 100,
@@ -118,12 +118,18 @@ export function springCalculation({
 	const from = 0;
 	const to = 1;
+	const resolvedConfig: SpringConfig = {
+		damping: config.damping ?? defaultSpringConfig.damping,
+		mass: config.mass ?? defaultSpringConfig.mass,
+		stiffness: config.stiffness ?? defaultSpringConfig.stiffness,
+		overshootClamping:
+			config.overshootClamping ?? defaultSpringConfig.overshootClamping,
+	};
 	const cacheKey = [
 		frame,
 		fps,
-		config.damping,
-		config.mass,
-		config.overshootClamping,
-		config.stiffness,
+		resolvedConfig.damping,
+		resolvedConfig.mass,
+		resolvedConfig.overshootClamping,
+		resolvedConfig.stiffness,
 	].join('-');
 	if (calculationCache[cacheKey]) {
 		return calculationCache[cacheKey];
@@ -144,10 +150,7 @@ export function springCalculation({
 		animation = advance({
 			animation,
 			now: time,
-			config: {
-				...defaultSpringConfig,
-				...config,
-			},
+			config: resolvedConfig,
 		});
 	}
 
@@ -156,10 +159,7 @@ export function springCalculation({
 		animation = advance({
 			animation,
 			now: (frameClamped / fps) * 1000,
-			config: {
-				...defaultSpringConfig,
-				...config,
-			},
+			config: resolvedConfig,
 		});
 	}
```

---

costajohnt:
## Summary

- `{...defaultSpringConfig, ...config}` let an explicit `undefined` (e.g. a forwarded optional React prop) override the default, which made the spring physics return `NaN`; a missing `damping` did not throw but produced a `NaN` damping ratio that silently selected the critically damped branch
- The calculation cache key was built from the raw config, so `{stiffness: undefined}` and `{}` shared a key and the `NaN` result poisoned every later default-config call at the same frame and fps
- Resolve the config once with `??` and use it for the cache key and both `advance()` calls
- Add regression coverage for `damping`, `stiffness` and `mass` set to `undefined`

Tradeoff: an alternative is to throw a `TypeError` on non-finite config values like the existing `damping <= 0` check does. Falling back to the defaults matches how a missing property already behaves and is what forwarding optional props expects, so this PR goes with the fallback. `measureSpring()` still keys its own cache on the raw config, so `{}` and `{damping: undefined}` share a key; both now compute the same value through `springCalculation()`, so the collision is harmless and it is left as is.

## Validation

- `bunx turbo make --filter=remotion`
- `cd packages/core && bun test src/test/spring.test.ts`
- `bunx turbo run test --filter=remotion`
- `cd packages/core && bunx oxfmt src --check && bunx eslint src && bunx tsgo --noEmit`

Closes #11359


JonnyBurger:
Thanks!

PR #11475, opened 2026-09-19
Elements: Add snapcn library

SriSomanaath:
## Summary

Adds [snapcn](https://snapcn.dev) to the third-party Element libraries listing.

snapcn is a shadcn registry of Remotion components (titles, captions, device frames, product-demo scenes). It is integrated with the Studio Protocol: every standalone component page has an **Add to Remotion Studio** button that sends the component through `installInStudio()`. 36 components ship as Element payloads today, each with its preview's props as `initialProps`.

- `browseUrl` and `catalogUrl` are both `https://snapcn.dev/docs/components`, the gallery. Opening a component there shows the install action. The site sends no frame-blocking headers, so it loads in the Studio modal.
- `bannerUrl` is `null` for now. Happy to provide artwork sized like the existing banners if you host it on `remotion.media`.

## Verification

- End to end against Remotion Studio 4.0.526: clicking **Add to Remotion Studio** on a snapcn component page opened the Studio install dialog. Studio wrote `src/text-reveal.element.tsx`, installed `@remotion/google-fonts`, inserted the component into the current composition, and the frame rendered.
- All 36 Element sources type-check under `strict`.
- Not run locally: `bun run build` / `stylecheck` for the docs package. This is a data-only change following the existing entries' format.


samohovets:
Hi @SriSomanaath!

Thanks for submitting this. We like the components and would be happy to welcome snapcn to the Elements ecosystem.

Before we list it, could you review the collection against our [Element Guidelines](https://www.remotion.dev/docs/elements/contributing#element-guidelines), especially the guidance on composability?

For example, Word Gather and Punch Lines include a background, fill the entire composition, and do not let you edit the text in the Studio Inspector. The Studio integration works, but the installed Elements should be easy to edit and combine with other content.

SriSomanaath:
Thanks @samohovets, that was fair feedback. I reviewed the whole collection against the Element Guidelines and rebuilt the Element payloads. The changes are live on snapcn.dev.

All 41 free components are now Elements, and each one:

- installs as a single layer sized to its content instead of filling the composition. For example, Word Gather is 860×380 and Punch Lines is 1120×380.
- has a transparent background, so it sits on your own footage.
- has its text, colours, font and timing editable in the Inspector.
- has an entrance and an exit, and assumes nothing about the frame.
- carries no snapcn branding and pins its dependencies to exact versions.
- resolves local files (`/logo.png` or `staticFile()`) from the project's `public/` folder.

I installed all 41 into Studio 4.0.526 and checked that each is one timeline row, text edits apply, move and trim work, and there are no console errors. I also type-checked and linted them in a fresh `npx create-video` project.

I've updated the PR description to match. Happy to change anything else.

samohovets:
@SriSomanaath Thanks for all the updates! I also appreciate you adding "Add to Remotion Studio" below "Make a video with this". It's much easier to find now.

I tried Card Rail in Studio and had trouble editing a single card. Its titles, small print, tags, and image URLs are in four separate `|`-separated fields, so changing one card means finding the matching entry in each field. I couldn't select a card on the canvas either.

Treeport Dev 2026-09-23 12 52 45

Would you mind taking another look at composability across the collection? A short list in one field works well for something like Word Flip, but cards have more individual pieces to edit. In the [official Elements catalog](https://www.remotion.dev/elements), we try to expose outlines for individual objects where possible, so people can select and edit one thing at a time. That doesn't mean every card needs its own timeline row. It would also help to check that each outline fits the Element throughout its animation, including any movement after the first frame.

One small thing I noticed while testing: Screen Recording and Laptop Frame default to MP4s, but their source controls use `assetType: "image"`, so Studio opens an image-only picker when replacing the video.

